PR #3, opened 2017-02-18
introspection: Add function to execute introspection query

YosubShin:
One can generate a schema JSON object that can be used for generating
type information for a GraphQL client
(e.g. https://github.com/apollographql/apollo-codegen).

### DIFF
--- a/graphql/introspection/introspection.go
+++ b/graphql/introspection/introspection.go
@@ -273,7 +273,7 @@ func (s *introspection) registerMutation(schema *schemabuilder.Schema) {
 	schema.Mutation()
 }
 
-func (s *introspection) Schema() *graphql.Schema {
+func (s *introspection) schema() *graphql.Schema {
 	schema := schemabuilder.NewSchema()
 
 	s.registerDirective(schema)
@@ -298,7 +298,7 @@ func AddIntrospectionToSchema(schema *graphql.Schema) {
 		query:    schema.Query,
 		mutation: schema.Mutation,
 	}
-	isSchema := is.Schema()
+	isSchema := is.schema()
 
 	query := schema.Query.(*graphql.Object)
 
--- a/graphql/introspection/introspection.go
+++ b/graphql/introspection/introspection.go
@@ -1,6 +1,9 @@
 package introspection
 
 import (
+	"strings"
+
+	"github.com/bradfitz/slice"
 	"github.com/samsarahq/thunder/graphql"
 	"github.com/samsarahq/thunder/graphql/schemabuilder"
 )
@@ -140,6 +143,7 @@ func (s *introspection) registerType(schema *schemabuilder.Schema) {
 			}
 		}
 
+		slice.Sort(fields, func(i, j int) bool { return strings.Compare(fields[i].Name, fields[j].Name) < 0 })
 		return fields
 	})
 
@@ -158,6 +162,7 @@ func (s *introspection) registerType(schema *schemabuilder.Schema) {
 						Type: Type{Inner: a},
 					})
 				}
+				slice.Sort(args, func(i, j int) bool { return strings.Compare(args[i].Name, args[j].Name) < 0 })
 
 				fields = append(fields, field{
 					Name: name,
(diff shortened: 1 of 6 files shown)


graphql/introspection/introspection.go
```diff
@@ -299,3 +301,128 @@ func AddIntrospectionToSchema(schema *graphql.Schema) {
...
+//
+//	valueJSON, err := json.Marshal(value)
+//	if err != nil {
+//		panic(err)
+//	}
+//
+//	fmt.Print(string(valueJSON))
+// ```
+func ExecuteIntrospectionQueryOnSchema(schema graphql.Schema) (interface{}, error) {
+	AddIntrospectionToSchema(&schema)
```
jellevandenhooff: AddIntrospectionToSchema modifies the schema, which is unfortunate, and might be surprising to callers of ExecuteIntrospectionQueryOnSchema. Maybe just document that, or maybe change AddIntrospectionToSchema to not modify the schema?

graphql/introspection/introspection.go
```diff
@@ -299,3 +301,128 @@ func AddIntrospectionToSchema(schema *graphql.Schema) {
...
+//
+// Example:
+//
+// ```
+//	value, err := introspection.ExecuteIntrospectionQueryOnSchema(*schema)
+//	if err != nil {
+//		panic(err)
+//	}
+//
+//	valueJSON, err := json.Marshal(value)
```
jellevandenhooff: I think it would make ExecuteIntrospectionQueryOnSchema nicer to use if it returned the JSON bytes.

graphql/introspection/introspection.go
```diff
@@ -299,3 +301,128 @@ func AddIntrospectionToSchema(schema *graphql.Schema) {
...
+	}
+`
+
+// ExecuteIntrospectionQueryOnSchema returns the result of executing a GraphQL
+// introspection query.
+//
+// Example:
+//
+// ```
+//	value, err := introspection.ExecuteIntrospectionQueryOnSchema(*schema)
```
jellevandenhooff: Maybe call this function ComputeSchemaJSON or something similar?

graphql/introspection/introspection.go
```diff
@@ -299,3 +301,128 @@ func AddIntrospectionToSchema(schema *graphql.Schema) {
 
 	schema.Query = isQuery
 }
+
+// Copied from https://github.com/graphql/graphiql/blob/master/src/utility/introspectionQueries.js
```
jellevandenhooff: Maybe stick this in its own file?

graphql/introspection/introspection.go
YosubShin: Wouldn't it be easier for a user to manipulate the resulting object if it's in unmarshaled form? (iterating through object and replace some nodes?)

graphql/introspection/introspection.go
YosubShin: Easy thing to do is to take `schemabuilder.Schema` and create a new `graphql.Schema` inside this function. This way the behavior won't be surprising (we don't change the parameter we're given).

graphql/introspection/introspection.go
YosubShin: Done.

graphql/introspection/introspection.go
YosubShin: Done.

graphql/introspection/introspection.go
YosubShin: Hmm actually unless we have a concrete usecase for this function other than returning the JSON bytes, I think you're right.

graphql/introspection/introspection.go
```diff
@@ -299,3 +302,27 @@ func AddIntrospectionToSchema(schema *graphql.Schema) {
 
 	schema.Query = isQuery
 }
+
+// ComputeSchemaJSON returns the result of executing a GraphQL introspection
+// query.
+func ComputeSchemaJSON(schemaBuilderSchema schemabuilder.Schema) ([]byte, error) {
```
jellevandenhooff: It would be nice if this also worked on schema.Schema, but we can do that some other time.

jellevandenhooff: A test would also be nice...

YosubShin: Done.

PR #4, opened 2017-02-19
schemabuilder/reflect: Support GraphQL default scalar types

YosubShin:
Add support for "Int", "Float", "String", and "Boolean" GraphQL scalar
types as defined in: http://graphql.org/learn/schema/#scalar-types

### DIFF
--- a/graphql/schemabuilder/reflect.go
+++ b/graphql/schemabuilder/reflect.go
@@ -612,8 +612,8 @@ func (sb *schemaBuilder) buildStruct(typ reflect.Type) error {
 }
 
 var scalars = map[reflect.Type]string{
-	reflect.TypeOf(bool(false)): "bool",
-	reflect.TypeOf(int(0)):      "int",
+	reflect.TypeOf(bool(false)): "Boolean",
+	reflect.TypeOf(int(0)):      "Int",
 	reflect.TypeOf(int8(0)):     "int8",
 	reflect.TypeOf(int16(0)):    "int16",
 	reflect.TypeOf(int32(0)):    "int32",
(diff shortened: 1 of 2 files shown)


graphql/schemabuilder/reflect.go
```diff
@@ -612,8 +612,8 @@ func (sb *schemaBuilder) buildStruct(typ reflect.Type) error {
 }
 
 var scalars = map[reflect.Type]string{
-	reflect.TypeOf(bool(false)): "bool",
-	reflect.TypeOf(int(0)):      "int",
+	reflect.TypeOf(bool(false)): "Boolean",
+	reflect.TypeOf(int(0)):      "Int",
```
jellevandenhooff: The graphql spec defines an Int as an int32. Go's `int` has a platform-dependent size. So, to follow the spec to the letter, only `int32` could map to a built-in graphql scalar. However, it might be easier if we expose all int-like types as Int. (and similar for float.) What do you think?

YosubShin: I mistakenly interpreted golang's int type to be exactly 32 bits, when it was at least 32 bits:
```
int is a signed integer type that is at least 32 bits in size. It is a distinct type, however, and not an alias for, say, int32.
```
I can revert the change from `int` to `Int`.

As far as mapping all int like types to `Int` and float like types to `Float` goes, I personally prefer to stick with the spec, unless we have a good reason to coerce them into the same type. GraphQL spec also seems to suggest creating a new scalar type if larger than 32 bit integer is required. https://facebook.github.io/graphql/#sec-Int

If we decide to stick with the spec, we'll map golang types `int32`, `float64`, `string`, and `bool` into GraphQL scalar types, and leave other types as it was before.

jellevandenhooff: Hm. In, that case, I think I prefer sticking to our Go scalar types: If we expose only some types as GraphQL scalar types, and others as Go scalars, almost any user will still have to deal with Go scalar types. That means the upside of this change is pretty minimal (since users will still have to deal with Go types), while the downside is the inconsistency between the two different naming schemes.

YosubShin: I agree with you in that I don't like introducing an inconsistency in the type naming scheme.
But I also think there is a value to natively supporting types like `String`, `Boolean`.
One other thing to note is that when we return an `int64` type value from thunder to a client, we lose the ability to express all possible integers of `int64`, beyond +/-`Number.MAX_SAFE_INTEGER`. 

For both of what I mentioned above, I don't think we need to worry about it now, but this may come up in the future and could potentially introduce a weird behavior if someone's dealing with a large integer.

jellevandenhooff:
let's postpone this change for now

PR #5, opened 2017-02-19
graphiql: add a prebuilt graphiql package

jellevandenhooff:
Getting the example running requires a lot of steps. I want to make a simpler example, and I think an embedded graphiql would be neat.

I really wish we had a client package on NPM because this code copy-paste is a recipe for problems...


jellevandenhooff:
CC @YosubShin @stephen

stephen:
@jellevandenhooff can we add instructions in the readme for how to run this?

PR #7, opened 2017-02-21
Add opentracing support

stephen:
Add opentracing support on thunder subscriptions for individual field execution, and live/non-live sql query execution.

### DIFF
--- a/graphql/executor.go
+++ b/graphql/executor.go
@@ -145,15 +145,12 @@ type resolveAndExecuteCacheKey struct {
 }
 
 func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source interface{}, selection *Selection) (interface{}, error) {
-	var span opentracing.Span
-	if _, ok := field.Type.(*Scalar); !ok {
-		span, ctx = opentracing.StartSpanFromContext(ctx, fmt.Sprintf("thunder.graphql.resolve(%s)", selection.Alias))
-		span.LogFields(
-			log.Bool("expensive", field.Expensive),
-		)
-	}
-
 	if field.Expensive {
+		var span opentracing.Span
+		if _, ok := field.Type.(*Scalar); !ok {
+			span, ctx = opentracing.StartSpanFromContext(ctx, fmt.Sprintf("thunder.graphql.resolve(%s)", selection.Alias))
+		}
+
 		// TODO: Skip goroutine for cached value
 		return fork(func() (interface{}, error) {
 			value := reflect.ValueOf(source)
@@ -169,17 +166,7 @@ func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source i
 			}
 
 			// TODO: Consider cacheing resolve and execute independently
-			fromCache := true
-			defer func() {
-				if span != nil {
-					span.LogFields(
-						log.Bool("cached", fromCache),
-					)
-					span.Finish()
-				}
-			}()
-			return reactive.Cache(ctx, key, func(ctx context.Context) (interface{}, error) {
-				fromCache = false
+			resolvedValue, valueFromCache, err := reactive.CacheWithStatus(ctx, key, func(ctx context.Context) (interface{}, error) {
 				value, err := safeResolve(ctx, field, source, selection.Args, selection.SelectionSet)
 				if err != nil {
 					return nil, err
@@ -193,14 +180,16 @@ func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source i
 				}
 				return await(value)
 			})
+
+			if span != nil {
+				span.LogFields(log.Bool("fromCache", valueFromCache))
+				span.Finish()
+			}
+
+			return resolvedValue, err
 		}), nil
 	}
 
-	defer func() {
-		if span != nil {
-			span.Finish()
-		}
-	}()
 	value, err := safeResolve(ctx, field, source, selection.Args, selection.SelectionSet)
 	if err != nil {
 		return nil, err
--- a/graphql/executor.go
+++ b/graphql/executor.go
@@ -9,6 +9,9 @@ import (
 	"sync"
 
 	"github.com/samsarahq/thunder/reactive"
+
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
 )
 
 type pathError struct {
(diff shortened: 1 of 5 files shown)


graphql/executor.go
```diff
@@ -175,6 +197,11 @@ func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source i
 		}), nil
 	}
 
+	defer func() {
```
stephen: i think it might make sense to only trace the expensive resolvers here, since the non-expensive ones should (hopefully) take trivial amounts of time to compute.

graphql/server.go
```diff
@@ -157,9 +159,12 @@ func (c *conn) handleSubscribe(id string, subscribe *subscribeMessage) error {
 		ctx = c.makeCtx(ctx)
 
 		start := time.Now()
+		span, ctx := opentracing.StartSpanFromContext(ctx, "thunder.subscription")
+		ext.SamplingPriority.Set(span, 1)
```
stephen: This can be set to 0 to disable the trace. This is effectively a no-op for now, but we can follow up to set this for a "debug" flag. Considering either adding a `trace` flag on the `InMessage` struct, or adding an incoming http header for the entire connection.

graphql/executor.go
jellevandenhooff: I agree. Could you test that change and see if the traces still look good?

graphql/executor.go
```diff
@@ -142,6 +145,15 @@ type resolveAndExecuteCacheKey struct {
 }
 
 func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source interface{}, selection *Selection) (interface{}, error) {
+	var span opentracing.Span
+	if _, ok := field.Type.(*Scalar); !ok {
+		innerSpan, innerCtx := opentracing.StartSpanFromContext(ctx, fmt.Sprintf("thunder.graphql.resolve(%s)", selection.Alias))
```
jellevandenhooff: Code similar to this repeats a lot. Could we simplify / extract some logic into a helper, or is this the way it's going to be?

jellevandenhooff: I think you can directly assign `span, ctx = opentracing.StartSpanFromContext(...`

graphql/executor.go
```diff
@@ -158,7 +170,17 @@ func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source i
 			}
 
 			// TODO: Consider cacheing resolve and execute independently
+			fromCache := true
```
jellevandenhooff: This `fromCache` thing is clever, but it's a little opaque. Maybe we could add an overload to reactive.Cache that returns if it was cached or not?

graphql/executor.go
```diff
@@ -158,7 +170,17 @@ func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source i
...
+			defer func() {
```
jellevandenhooff: Defer only if span != nil

graphql/executor.go
```diff
@@ -158,7 +170,17 @@ func (e *Executor) resolveAndExecute(ctx context.Context, field *Field, source i
...
+				if span != nil {
+					span.LogFields(
+						log.Bool("cached", fromCache),
+					)
+					span.Finish()
```
jellevandenhooff: It would be good to comment why you call span.Finish here and not from a defer right after you add the span to the context.

graphql/server.go
jellevandenhooff: If it's a no-op, let's not include it.

Does the code run without problems if there is no global tracer setup?

livesql/live.go
```diff
@@ -111,17 +113,30 @@ type queryCacheKey struct {
 
 // query reactively performs a SelectQuery
 func (ldb *LiveDB) query(ctx context.Context, query *sqlgen.SelectQuery, filter sqlgen.Filter) ([]interface{}, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "thunder.livesql.query")
+	fromCache := true
+	defer func() {
+		span.LogFields(
```
jellevandenhooff: I think these (and the other calls to `LogFields`) would look nicer on just a single line.

sqlgen/db.go
```diff
@@ -33,9 +36,20 @@ func (db *DB) Query(ctx context.Context, result interface{}, filter Filter, opti
 	if err != nil {
 		return err
 	}
-	querySql, fields := query.ToSQL()
+	clause, fields := query.ToSQL()
 
-	res, err := db.QueryExecer(ctx).Query(querySql, fields...)
+	if span := opentracing.SpanFromContext(ctx); span != nil {
```
jellevandenhooff: These all look very copy-pasty. Could we de-duplicate the code?

sqlgen/db.go
```diff
@@ -33,9 +36,20 @@ func (db *DB) Query(ctx context.Context, result interface{}, filter Filter, opti
 	if err != nil {
 		return err
 	}
-	querySql, fields := query.ToSQL()
+	clause, fields := query.ToSQL()
```
jellevandenhooff: Thanks for this cleanup.

sqlgen/db.go
```diff
@@ -125,6 +174,18 @@ func (db *DB) UpdateRow(ctx context.Context, row interface{}) error {
 	}
 
 	clause, args := query.ToSQL()
+
+	if span := opentracing.SpanFromContext(ctx); span != nil {
+		span, innerCtx := opentracing.StartSpanFromContext(ctx, "thunder.sqlgen.UpsertRow")
```
jellevandenhooff: UpdateRow

sqlgen/db.go
```diff
@@ -143,6 +204,18 @@ func (db *DB) DeleteRow(ctx context.Context, row interface{}) error {
...
```
jellevandenhooff: DeleteRow

graphql/server.go
stephen: sgtm. Locally, it looks like this code will execute but the spans will not be reported anywhere when there is no tracer.

Do you know if we can get the query's name (if given) from here?

jellevandenhooff: I think that will require some refactoring in parser.go. Would be good to do anyway, because it will make for better logging.

jellevandenhooff:
Oh, and a meta-comment: could you split this in 2 commits? I think the commit message prefix already shows that this one touches a bunch. I'm also guilty of that...

sqlgen/db.go
stephen: I refactored these lines a little to a point that I am okay with. I can't think of good ways to make this more succinct, since we need to both replace the existing `ctx`, and also call `defer` in this function.

reactive/rerunner.go
```diff
@@ -158,9 +158,10 @@ func run(ctx context.Context, f ComputeFunc) (*computation, error) {
 	return c, nil
 }
 
-func Cache(ctx context.Context, key interface{}, f ComputeFunc) (interface{}, error) {
+func CacheWithStatus(ctx context.Context, key interface{}, f ComputeFunc) (interface{}, bool, error) {
```
stephen: @jellevandenhooff - lmk if you have a better name for this.

stephen:
@jellevandenhooff addressed comments, ptal again!

PR #23, opened 2017-03-28
graphql server: Limit number of concurrent rerunners

silverspace:
When a graphql server first starts up and has an influx of subscribers,
we've seen that it performs badly with a cold cache. This is because we
attempt to recompute the results of all 200 subscriptions concurrently,
which spreads the server's resources too thin.

In an effort to alleviate some of this pain, we are limiting the maximum
number of subscription rerunners that can execute in parallel to 64.
This is done by acquiring a semaphore in the NewRerunner computation
function and releasing the semaphore at the end of the function.

### DIFF
--- a/graphql/server.go
+++ b/graphql/server.go
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	MaxSubscriptions = 200
-	MinRerunInterval = 5 * time.Second
+	MaxSubscriptions       = 200
+	MaxConcurrentRerunners = 64
+	MinRerunInterval       = 5 * time.Second
 )
 
 type JSONSocket interface {
@@ -36,6 +37,24 @@ type GraphqlLogger interface {
 	Error(ctx context.Context, err error, tags map[string]string)
 }
 
+type empty struct{}
+type semaphore chan empty
+
+// acquire n resources
+func (s semaphore) acquire(n int) {
+	e := empty{}
+	for i := 0; i < n; i++ {
+		s <- e
+	}
+}
+
+// release n resources
+func (s semaphore) release(n int) {
+	for i := 0; i < n; i++ {
+		<-s
+	}
+}
+
 type conn struct {
 	writeMu sync.Mutex
 	socket  JSONSocket
@@ -46,7 +65,8 @@ type conn struct {
 
 	url string
 
-	mutateMu sync.Mutex
+	mutateMu   sync.Mutex
+	computeSem semaphore
 
 	mu            sync.Mutex
 	subscriptions map[string]*reactive.Rerunner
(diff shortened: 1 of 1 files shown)


silverspace:
As a follow-up to this PR, we should consider adding middleware support for server.go. This will allow us to:
 * Implement this semaphore locking as middleware, so different consumers can specify different values of MaxConcurrentRerunners.
 * Collect metrics on the number of concurrent re-runners (without adding a dependency on prometheus, etc in this package).

graphql/server.go
```diff
@@ -344,9 +367,10 @@ func ServeJSONSocket(socket JSONSocket, schema *Schema, makeCtx MakeCtxFunc, log
 	c := &conn{
 		socket: socket,
 
-		schema:  schema,
-		makeCtx: makeCtx,
-		logger:  logger,
+		schema:     schema,
+		makeCtx:    makeCtx,
+		logger:     logger,
+		computeSem: make(semaphore, MaxConcurrentRerunners),
```
jellevandenhooff: Could we make the semaphore part of the server struct so that it's shared across computations?

graphql/server.go
```diff
@@ -36,6 +37,24 @@ type GraphqlLogger interface {
 	Error(ctx context.Context, err error, tags map[string]string)
 }
 
+type empty struct{}
+type semaphore chan empty
```
jellevandenhooff: Instead of empty, just inline `struct{}`

graphql/server.go
```diff
@@ -36,6 +37,24 @@ type GraphqlLogger interface {
...
+
+// acquire n resources
```
jellevandenhooff: nit (and a little silly in a file that's as messy as it is): please end your comments with a period

graphql/server.go
```diff
@@ -36,6 +37,24 @@ type GraphqlLogger interface {
...
+func (s semaphore) acquire(n int) {
```
jellevandenhooff: Since you only ever acquire and release with n=1, let's remove the argument.

graphql/server.go
```diff
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	MaxSubscriptions = 200
-	MinRerunInterval = 5 * time.Second
+	MaxSubscriptions       = 200
+	MaxConcurrentRerunners = 64
```
jellevandenhooff: Could we make this configurable on the server? I regret these constants and would like to not add any more.

silverspace:
I am closing this PR in favor of this implementation:

https://github.com/samsarahq/thunder/pull/25
https://github.com/samsara-dev/backend/pull/1554

PR #25, opened 2017-04-07
Greg/graphql Add subscription adapter support to thunder

silverspace:
Adding a subscription adapter middleware, based loosely on the http.Handler middleware implementation and Adapter design pattern. 

This adapter is designed to support logging, stats and throttling support for Graphql subscription Executors. The implementation is source API compatible with previous releases.


sandeepmeenuga:
Any issue with subscriptions, you did delete?

PR #31, opened 2017-04-24
Add MaybeStartSpanFromContext helper

stephen:
- [ ] Consider writing a test

### DIFF
--- a/opentracingkit/opentracingkit.go
+++ b/opentracingkit/opentracingkit.go
@@ -1,0 +1,29 @@
+package opentracingkit
+
+import (
+	"context"
+
+	"github.com/opentracing/opentracing-go"
+)
+
+var noopTracer = &opentracing.NoopTracer{}
+
+func MaybeStartSpanFromContext(
+	ctx context.Context,
+	operationName string,
+	opts ...opentracing.StartSpanOption,
+) (opentracing.Span, context.Context) {
+	if span := opentracing.SpanFromContext(ctx); span != nil {
+		span, ctx := opentracing.StartSpanFromContext(ctx, operationName, opts...)
+		return span, ctx
+	} else {
+		// If there is no parent span, we pass back a working
+		// NoopSpan for the current context to call on, but not
+		// report back to the underlying tracer.
+		// We also do not contribute the new span to the
+		// context so that downstream spans do not declare
+		// themselves children of the NoopSpan.
+		span := noopTracer.StartSpan(operationName)
+		return span, ctx
+	}
+}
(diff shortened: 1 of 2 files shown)


helpers/opentracing.go
```diff
@@ -0,0 +1,34 @@
+package helpers
```
stephen: this might belong in a different package? (or eventually not in this repo altogether)

helpers/opentracing.go
```diff
@@ -0,0 +1,34 @@
...
+		span, ctx := opentracing.StartSpanFromContext(ctx, operationName, opts...)
+		return span, ctx
+	} else {
+		// If there is no parent span, we pass back a working
+		// MockSpan for the current context to call on, but not
+		// report back to the underlying tracer.
+		// We also do not contribute the new span to the
+		// context so that downstream spans do not declare
+		// themselves children of the MockSpan.
+		span, _ := opentracing.StartSpanFromContext(ctx, operationName, opts...)
```
jellevandenhooff: What are the consequences of calling StatsSpanFromContext here? Will it propagate outward, somehow?

helpers/opentracing.go
jellevandenhooff: I really don't like the name helpers.

helpers/opentracing.go
stephen: I've found that it's possible for these spans to still be propagated in the case where a MockSpan has descendent spans that are eventually collected by the tracer, which is why the block discards the `ctx` passed back by `StartSpanFromContext`. I'll investigate the no-op tracer approach.

stephen:
@jellevandenhooff - Changed this to call out to the noop tracer.

helpers/opentracing.go
```diff
@@ -0,0 +1,30 @@
...
+		span, ctx := opentracing.StartSpanFromContext(ctx, operationName, opts...)
+		return span, ctx
+	} else {
+		// If there is no parent span, we pass back a working
+		// NoopSpan for the current context to call on, but not
+		// report back to the underlying tracer.
+		// We also do not contribute the new span to the
+		// context so that downstream spans do not declare
+		// themselves children of the NoopSpan.
+		// span, _ := opentracing.StartSpanFromContext(ctx, operationName, opts...)
```
jellevandenhooff: get rid of this line

helpers/opentracing.go
jellevandenhooff: Could you make call this packing opentracinghelpers? Or opentracingutil? Or tracingutil?

stephen: from irl - this is now `opentracingkit`!

PR #36, opened 2017-05-08
Computation Middleware

stephen:
This PR adds support for middleware in computing queries/mutations.

An example use case for this might be contributing an id to all computations:
```golang
c.Use(func(input *graphql.ComputationInput, output *graphql.ComputationOutput, next graphql.NextFunc) {
  output.Metadata["computationId"] = uuid.New()
  next(input, output);
})
```
the `Metadata` map is appended to the websocket response alongside the `message`:
screen shot 2017-05-10 at 4 49 22 pm

Another use-case might be contributing to the `context`:
```golang
c.Use(func(input *graphql.ComputationInput, output *graphql.ComputationOutput, next graphql.NextFunc) {
  input.Ctx = opentracing.StartSpanFromContext(input.Ctx, "graphql.query");
  next(input, output);
})
```

#### Notes:
- note that this PR also removes starting a subscription span from the graphql handler.
- if this is a pattern we like, we might consider modeling existing features as middleware:
    - `makeCtx`
    - `GraphqlLogger`

#### Breaking changes:
- opentracing spans are no longer automatically started by thunder subscriptions

### DIFF
--- a/graphql/middleware.go
+++ b/graphql/middleware.go
@@ -21,17 +21,17 @@ type MiddlewareFunc func(input *ComputationInput, next MiddlewareNextFunc) *Comp
 type MiddlewareNextFunc func(input *ComputationInput) *ComputationOutput
 
 func runMiddlewares(middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput {
-	var _runMiddlewares func(index int, middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput
-	_runMiddlewares = func(index int, middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput {
+	var run func(index int, middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput
+	run = func(index int, middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput {
 		if index < len(middlewares) {
 			return &ComputationOutput{}
 		}
 
 		middleware := middlewares[index]
 		return middleware(input, func(input *ComputationInput) *ComputationOutput {
-			return _runMiddlewares(index+1, middlewares, input)
+			return run(index+1, middlewares, input)
 		})
 	}
 
-	return _runMiddlewares(0, middlewares, input)
+	return run(0, middlewares, input)
 }
--- a/graphql/middleware.go
+++ b/graphql/middleware.go
@@ -17,20 +17,21 @@ type ComputationOutput struct {
 	Error    error
 }
 
-type MiddlewareFunc func(input *ComputationInput, output *ComputationOutput, next NextFunc)
-type NextFunc func(input *ComputationInput, output *ComputationOutput)
+type MiddlewareFunc func(input *ComputationInput, next MiddlewareNextFunc) *ComputationOutput
+type MiddlewareNextFunc func(input *ComputationInput) *ComputationOutput
 
-func _runMiddlewares(index int, middlewares []MiddlewareFunc, input *ComputationInput, output *ComputationOutput) {
-	if index >= len(middlewares) {
-		return
+func runMiddlewares(middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput {
+	var _runMiddlewares func(index int, middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput
+	_runMiddlewares = func(index int, middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput {
+		if index < len(middlewares) {
+			return &ComputationOutput{}
+		}
+
+		middleware := middlewares[index]
+		return middleware(input, func(input *ComputationInput) *ComputationOutput {
+			return _runMiddlewares(index+1, middlewares, input)
+		})
 	}
 
-	middleware := middlewares[index]
-	middleware(input, output, func(input *ComputationInput, output *ComputationOutput) {
-		_runMiddlewares(index+1, middlewares, input, output)
-	})
-}
-
-func runMiddlewares(middlewares []MiddlewareFunc, input *ComputationInput, output *ComputationOutput) {
-	_runMiddlewares(0, middlewares, input, output)
+	return _runMiddlewares(0, middlewares, input)
 }
--- a/graphql/middleware.go
+++ b/graphql/middleware.go
@@ -1,0 +1,36 @@
+package graphql
+
+import "context"
+
+type ComputationInput struct {
+	Id          string
+	Query       string
+	ParsedQuery *Query
+	Variables   map[string]interface{}
+	Ctx         context.Context
+	Previous    interface{}
+}
+
+type ComputationOutput struct {
+	Metadata map[string]interface{}
+	Current  interface{}
+	Error    error
+}
+
+type MiddlewareFunc func(input *ComputationInput, output *ComputationOutput, next NextFunc)
+type NextFunc func(input *ComputationInput, output *ComputationOutput)
+
+func _runMiddlewares(index int, middlewares []MiddlewareFunc, input *ComputationInput, output *ComputationOutput) {
+	if index >= len(middlewares) {
+		return
+	}
+
+	middleware := middlewares[index]
+	middleware(input, output, func(input *ComputationInput, output *ComputationOutput) {
+		_runMiddlewares(index+1, middlewares, input, output)
+	})
+}
+
+func runMiddlewares(middlewares []MiddlewareFunc, input *ComputationInput, output *ComputationOutput) {
+	_runMiddlewares(0, middlewares, input, output)
+}
--- a/graphql/server.go
+++ b/graphql/server.go
@@ -248,7 +248,28 @@ func (c *conn) handleMutate(id string, mutate *mutateMessage) error {
 
 		start := time.Now()
 		c.logger.StartExecution(ctx, tags, true)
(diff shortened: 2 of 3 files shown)


graphql/middleware.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+type ComputationOutput struct {
+	Metadata map[string]interface{}
+	Current  interface{}
+	Error    error
+}
+
+type MiddlewareFunc func(input *ComputationInput, output *ComputationOutput, next NextFunc)
+type NextFunc func(input *ComputationInput, output *ComputationOutput)
```
jellevandenhooff: Maybe MiddlewareNextFunc?

graphql/middleware.go
```diff
@@ -0,0 +1,36 @@
...
+
+func _runMiddlewares(index int, middlewares []MiddlewareFunc, input *ComputationInput, output *ComputationOutput) {
```
jellevandenhooff: This would be cleaner as a lambda function below:
```go
func runMiddlewares(middlewares []MiddlewareFunc, input *ComputationInput, output *ComputationOutput) {
    var run func(index int, input *ComputationInput, output *ComputationOutput)
    run = func(index int, input *ComputationInput, output *ComputationOutput) {
        if index >= len(middlewares) {
            return
        }
        middleware := middlewares[index]
        middle(input, output, func(input *ComputationInput, output *ComputationOutput) {
              run(index+1, input, output)
        }
    }
    run(0, input, output)
}

File: graphql/middleware.go [hunk 3]
```diff
@@ -0,0 +1,36 @@
...
+	Previous    interface{}
+}
+
+type ComputationOutput struct {
+	Metadata map[string]interface{}
+	Current  interface{}
+	Error    error
+}
+
+type MiddlewareFunc func(input *ComputationInput, output *ComputationOutput, next NextFunc)
```
jellevandenhooff: What do you think of making MiddlewareFunc and NextFunc return *ComputationOutput instead?

graphql/server.go
```diff
@@ -109,14 +110,15 @@ func isCloseError(err error) bool {
 	return ok || err == websocket.ErrCloseSent
 }
 
-func (c *conn) writeOrClose(id string, typ string, message interface{}) {
+func (c *conn) writeOrClose(id string, typ string, message interface{}, metadata map[string]interface{}) {
```
jellevandenhooff: This thing should just take an outEnvelope as argument instead, I think.

graphql/server.go
```diff
@@ -165,14 +167,34 @@ func (c *conn) handleSubscribe(id string, subscribe *subscribeMessage) error {
...
+
 		c.logger.StartExecution(ctx, tags, initial)
-		current, err := e.Execute(ctx, c.schema.Query, nil, query)
+
+		output := &ComputationOutput{
+			Metadata: make(map[string]interface{}),
+		}
+
+		var middlewares []MiddlewareFunc
+		middlewares = append(middlewares, c.middlewares...)
```
jellevandenhooff: You could `middlewares := append(nil, c.middlewares...)` or maybe that confuses the type inferencer.

graphql/server.go
```diff
@@ -226,11 +248,32 @@ func (c *conn) handleMutate(id string, mutate *mutateMessage) error {
 
 		start := time.Now()
 		c.logger.StartExecution(ctx, tags, true)
-		current, err := e.Execute(ctx, c.schema.Mutation, c.schema.Mutation, query)
+
+		output := &ComputationOutput{
+			Metadata: make(map[string]interface{}),
+		}
+
+		var middlewares []MiddlewareFunc
```
jellevandenhooff: While you're here, could you maybe try and de-duplicate these two functions? They're starting to bother me a lot.

graphql/server.go
```diff
@@ -344,8 +387,17 @@ func Handler(schema *Schema) http.Handler {
 	})
 }
 
+func (c *conn) Use(fn MiddlewareFunc) {
+	c.middlewares = append(c.middlewares, fn)
+}
+
 func ServeJSONSocket(ctx context.Context, socket JSONSocket, schema *Schema, makeCtx MakeCtxFunc, logger GraphqlLogger) {
```
jellevandenhooff: I'd be fine breaking the ServeJSONSocket interface. I am not a big fan of the Use function.

graphql/server.go
```diff
@@ -165,14 +167,34 @@ func (c *conn) handleSubscribe(id string, subscribe *subscribeMessage) error {
...
+		middlewares = append(middlewares, func(input *ComputationInput, output *ComputationOutput, next NextFunc) {
+			output.Current, output.Error = e.Execute(input.Ctx, c.schema.Query, nil, input.ParsedQuery)
+			next(input, output)
+		})
+
+		runMiddlewares(middlewares, &ComputationInput{
+			Ctx:         ctx,
+			Id:          id,
+			ParsedQuery: query,
+			Previous:    previous,
```
jellevandenhooff: Is Previous relevant?

graphql/server.go
```diff
@@ -165,14 +167,34 @@ func (c *conn) handleSubscribe(id string, subscribe *subscribeMessage) error {
...
+			Query:       subscribe.Query,
```
jellevandenhooff: I would like to drop the non-parsed query.

jellevandenhooff: (Maybe)

graphql/middleware.go
stephen: Earlier version did this - I think it's reasonable, though it does impact when you can contribute to `Metadata`, i.e. you must wait until the base middleware (the executor) contributes `output.Metadata` before any middleware can write to it.

e.g.
```golang
c.Use(func(input *graphql.ComputationInput, output *graphql.ComputationOutput, next graphql.NextFunc) {
  requestId := middleware.RequestIdFromContext(input.Ctx)
  output.Metadata["requestId"] = requestId
  next(input, output)
})
```
would become:
```golang
c.Use(func(input *graphql.ComputationInput, next graphql.NextFunc) *graphql.ComputationOutput {
  requestId := middleware.RequestIdFromContext(input.Ctx)
  output := next(input, output)
  output.Metadata["requestId"] = requestId
})
```

graphql/middleware.go
stephen: Thanks - I had this originally but was a bit annoyed by the `run` type duplication.

graphql/server.go
stephen: I think it'd be useful for letting the middleware know whether or not we are doing a recomputation. In the tracing middleware, I have a line that looks like:
```golang
span.SetTag("computation.initial", input.Previous == nil)
```

If we were to also to model the graphql logger as middleware in the future, it'd also be useful to have this "is recomputation" bit.

graphql/server.go
stephen: If there's a method to print the parsed query with the raw one, I'm fine with that.

graphql/server.go
stephen: Can you clarify the second point? I find the create/use thing a little strange because we're creating/hooking up middleware once per every connection handler.

I left `ServeJSONSocket` for compatibility, but maybe it's fine to break.

graphql/middleware.go
stephen: Also, the base case would have to be:
```golang
middlewares = append(middlewares, func(input *ComputationInput, output *ComputationOutput, next NextFunc) {
  output := next(input, output)
  output.Current, output.Error = e.Execute(input.Ctx, c.schema.Query, nil, input.ParsedQuery)
})
```

graphql/server.go
stephen: Yes, the type inferencer gets unhappy. (`first argument to append must be typed slice; have untyped nil`)

graphql/server.go
stephen: I think this is not the case because we don't have a printer method and we discard the underlying document node here: https://github.com/samsarahq/thunder/blob/master/graphql/parser.go#L344

graphql/server.go
stephen: They're subtly different enough that I'd prefer to punt for now.

jellevandenhooff: Fair enough.

stephen:
@jellevandenhooff - ptal at the last two commits.

graphql/middleware.go
```diff
@@ -0,0 +1,37 @@
...
+	Metadata map[string]interface{}
+	Current  interface{}
+	Error    error
+}
+
+type MiddlewareFunc func(input *ComputationInput, next MiddlewareNextFunc) *ComputationOutput
+type MiddlewareNextFunc func(input *ComputationInput) *ComputationOutput
+
+func runMiddlewares(middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput {
+	var _runMiddlewares func(index int, middlewares []MiddlewareFunc, input *ComputationInput) *ComputationOutput
```
jellevandenhooff: can you drop the underscore?

graphql/server.go
jellevandenhooff: I find the "Use" function name to be to generic to be helpful. I think breaking ServeJSONSocket is fine. We'll break it again when we drop the makeCtx and logger endpoints (which I think is a great idea.)

graphql/middleware.go
stephen: What do you prefer here? How about `run`?

jellevandenhooff: that's great

PR #38, opened 2017-05-12
Update tracing in sqlgen

stephen:


sqlgen/db.go
```diff
@@ -32,10 +30,8 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
 		Many: func(ctx context.Context, items []interface{}) ([]interface{}, error) {
 			table := items[0].(*BaseSelectQuery).Table
 
-			if span := opentracing.SpanFromContext(ctx); span != nil {
-				span, ctx = opentracingkit.MaybeStartSpanFromContext(ctx, fmt.Sprintf("thunder.sqlgen.BatchQuery(%s)", table.Name))
-				defer span.Finish()
-			}
+			span, ctx := opentracingkit.MaybeStartSpanFromContext(ctx, fmt.Sprintf("thunder.sqlgen.BatchQuery(%s)", table.Name))
```
silverspace: Quick fly by comment:
- We should also tag the number of items being queried and the number of rows returned.

jellevandenhooff: +1

sqlgen/db.go
```diff
@@ -188,10 +190,16 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
 	clause, args := query.ToSQL()
 
 	span, ctx := opentracingkit.MaybeStartSpanFromContext(ctx, "thunder.sqlgen.InsertRow")
-	span.LogFields(log.String("query", clause))
 	defer span.Finish()
+	span.SetTag("query", clause)
+	span.SetTag("queryVariables", args)
 
-	return db.QueryExecer(ctx).Exec(clause, args...)
+	result, err := db.QueryExecer(ctx).Exec(clause, args...)
```
jellevandenhooff: Seems like a wrapper function execWithTrace (result, error) would reduce a bunch of code duplication.

stephen: not sure

stephen:
ptal at the last two commits, @jellevandenhooff @silverspace.
screen shot 2017-05-16 at 6 02 45 pm

sqlgen/db.go
```diff
@@ -55,17 +53,26 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
 				return nil, err
 			}
 			clause, args = selectQuery.ToSQL()
+			span.SetTag("query", clause)
+			span.SetTag("batchItemCount", len(items))
+			if marshaledItems, err := json.Marshal(filters); err == nil {
+				span.SetTag("batchItems", string(marshaledItems))
```
jellevandenhooff: Kind of expensive maybe? It's be nice if we could only do this if this is a non noop trace.

stephen: cpu intensive for marshaling, or expensive to store on the span?

jellevandenhooff: I am more worried about marshaling, I think. I don't know what the maximum amount of data we can associate with a span is.

silverspace: Agreed, I don't think we should include the actual SQL results, since they could be fairly large (e.g. trip proto blobs). Having the number of returned results should be sufficient for most purposes, and we can always add this in later if needed.

sqlgen/db.go
```diff
@@ -55,17 +53,26 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
...
+			}
+			span.SetTag("queryVariables", args)
 
 			// Then, run the SQL query.
 			res, err := db.Conn.Query(clause, args...)
 			if err != nil {
+				ext.Error.Set(span, true)
+				span.SetTag("errorMessage", err.Error())
```
silverspace: I think we want to use span.LogFields(log.Error(err)) rather than defining our own errorMessage tag.

sqlgen/db.go
```diff
@@ -55,17 +53,26 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
 				return nil, err
 			}
 			clause, args = selectQuery.ToSQL()
+			span.SetTag("query", clause)
+			span.SetTag("batchItemCount", len(items))
+			if marshaledItems, err := json.Marshal(filters); err == nil {
+				span.SetTag("batchItems", string(marshaledItems))
+			}
+			span.SetTag("queryVariables", args)
```
silverspace: According to the opentracing godocs: "Tag values can be numeric types, strings, or bools. "

Rather than relying on undefined behavior, we should log the string: fmt.Sprintf("%v", args)

sqlgen/mysql.go
```diff
@@ -28,6 +28,10 @@ func (w *SimpleWhere) ToSQL() (string, []interface{}) {
 	return buffer.String(), w.Values
 }
 
+type SQLQuery interface {
```
silverspace: Nit: Could we rename this SQLStatement since we're using this interface for DeleteRow and InsertRow as well? (in my mind SQL query is synonymous with SQL select, though that may be inaccurate. Feel free to ignore if you disagree).

sqlgen/db.go
stephen: This is the number of filters applied (i.e. the args for the queries that are batched, not the result items.)

silverspace: Ah sorry, I meant to update that comment before posting. I think it should be fine to log the filters, but I don't think it's necessary. We won't gain visibility into how long each filter takes on the SQL side.

sqlgen/mysql.go
jellevandenhooff: It could also be an unexported interface.

PR #39, opened 2017-05-30
graphql batch: Increase waiting period if batch fn is invoked rapidly

silverspace:
Adding support for a second timer in the batch function execution logic,
which is reset every time the batch fn is invoked. This is separate from
the MaxDuration timer, which is increased to 10ms as an absolute time
restriction. The interval timer will expire unless the batch fn is
invoked every 500us.

### DIFF
--- a/batch/batch.go
+++ b/batch/batch.go
@@ -23,8 +23,11 @@ import (
 	"time"
 )
 
+// DefaultWaitInterval is the default WaitInterval for Func.
+const DefaultWaitInterval = 200 * time.Microsecond
+
 // DefaultMaxDuration is the default MaxDuration for Func.
-var DefaultMaxDuration = 200 * time.Microsecond
+const DefaultMaxDuration = 20 * time.Millisecond
 
 // A Func transforms a function that takes a batch of inputs (Func.Many) into a
 // function that takes single inputs (Func.Invoke). Multiple concurrenct
@@ -42,6 +45,10 @@ type Func struct {
 	// invocations, Many will be invoked even if some goroutines are stil running.
 	// Zero, the default, means no limit.
 	MaxSize int
+	// WaitInterval is the duration of a timer that is reset every time the
+	// batch function is invoked. Many will be invoked when either the
+	// WaitInterval or MaxDuration expires.
+	WaitInterval time.Duration
 	// MaxDuration limits the duration of a batch. After waiting for
 	// MaxDuration, Many will be invoked even if some goroutines are still
 	// running. Defaults to DefaultMaxDuration.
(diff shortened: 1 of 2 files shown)


batch/batch.go
```diff
@@ -23,8 +23,11 @@ import (
 	"time"
 )
 
+// DefaultWaitInterval is the default WaitInterval for Func.
+var DefaultWaitInterval = 500 * time.Microsecond
+
 // DefaultMaxDuration is the default MaxDuration for Func.
-var DefaultMaxDuration = 200 * time.Microsecond
+var DefaultMaxDuration = 10 * time.Millisecond
```
stephen: It'd be nice to have numbers to back this default. If we could get some traces that roughly how long is reasonable, or if we can track the default "actual wait" times batches. As a lower bound, I've definitely seen failed-to-batch patterns with our current timing.

batch/batch.go
```diff
@@ -154,6 +171,15 @@ func (f *Func) Invoke(ctx context.Context, arg interface{}) (interface{}, error)
 
 		// Publish the batchGroup.
 		bctx.pendingBatchGroups[fs] = bg
+	} else {
+		// This is a subsequent invocation of the batch function, so
+		// reset the intervalTimer.
+		if bg.intervalTimer.Stop() {
+			// Only reset the timer if Stop successfully
+			// stopped the timer. Otherwise, the receive channel
+			// will be populated and we'll select on it below.
```
stephen: Could you describe the case where we fail to stop the timer? Is if we've already `select`'d below?

batch/batch.go
silverspace: Agreed. I am starting with DefaultWaitDuration = 200us, which was the old DefaultMaxDuration. There is no magic constant that will be perfect for all use cases; these values are environment and request specific. The main use case we should be trying to solve with this implementation is a single GraphQL request that fans out to multiple batch requests. From the traces I've analyzed, this does satisfy that use case.

batch/batch.go
silverspace: Yes, if we fail to stop the timer, that means we've hit a race condition where we requested to stop the timer but the receive channel has already been written to. In this case, we should allow the select loop to read the value and execute the batching function.

silverspace:
Added a new batching test to verify the WaitInterval behavior.

batch/batch.go
stephen: Makes sense. Maybe as a future change we could make the sqlgen batcher options configurable from outside of the package so that it's easier to tweak this without changing thunder.

batch/batch_test.go
```diff
@@ -43,6 +44,77 @@ func TestBasic(t *testing.T) {
...
+				defer wg.Done()
+				if result, err := f(ctx, i); err != nil || result != i {
+					t.Error(err, i)
+				}
+			}(i)
+			time.Sleep(testcase.sleepDuration)
+		}
+		wg.Wait()
+
+		// Expect 1, allow for 2 in case of races.
```
stephen: consider moving this comment up to where the structs are defined + updating to match n vs. n + 1.

silverspace: Thanks, updated the comments to describe the expected vs tolerated values.

PR #40, opened 2017-06-05
thunder sqlgen: Add support for nil values in CopySlice

silverspace:
Fixing a panic in CopySlice when the input slice contains nil values:
panic: reflect: call of reflect.Value.Set on zero Value

Also adding a test to verify the behavior of CopySlice with nil values.

### DIFF
--- a/sqlgen/reflect.go
+++ b/sqlgen/reflect.go
@@ -143,7 +143,9 @@ func CopySlice(result interface{}, rows []interface{}) error {
 	slice := ptr.Elem()
 	slice.Set(reflect.MakeSlice(slice.Type(), len(rows), len(rows)))
 	for i, row := range rows {
-		slice.Index(i).Set(reflect.ValueOf(row))
+		if row != nil {
+			slice.Index(i).Set(reflect.ValueOf(row))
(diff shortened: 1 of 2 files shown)


batch/batch_test.go
```diff
@@ -48,34 +48,35 @@ func TestBasic(t *testing.T) {
...
-		allowedCounts map[int]bool
+		expectedCount int
 	}{
 		{
 			description:   "Expect sleep less than WaitDuration to result in a single call being made.",
-			interval:      500 * time.Microsecond,
-			sleepDuration: 100 * time.Microsecond,
-			maxDuration:   10 * time.Millisecond,
-			allowedCounts: map[int]bool{1: true},
+			interval:      2 * sleepDuration,
```
silverspace: Sure, I'll file a follow-up PR to increase the default wait interval to 1ms. I'm reluctant to go too much higher since any request that isn't batched will pay this penalty for every batch invocation.

I'll also include metrics in the new PR for recording the cardinality of the batch request to DataDog, so we can see how this increases or decreases as we play with the DefaultWaitDuration.

PR #43, opened 2017-06-13
reactive rerunner: Fix potential rerunner memory leak

silverspace:
Updating Rerunner.run() to immediately return if the cancelCtx function
is invoked, since we may have a client that has specified a large
minRerunInterval. For example, if the client specifies minRerunInterval
= math.MaxInt64, then run() will be called once with delay=0, and then a
second time (due to computation.node.handleInvalidate(r.run)) with a
very large delay. The second sleep causes the rerunner object and its
associated cache to stay alive for the lifetime of the process.

The fix for this potential memory leak is to select on both ctx.Done and
the sleep timer, and return immediately if cancelCtx() is called.

### DIFF
--- a/reactive/rerunner.go
+++ b/reactive/rerunner.go
@@ -227,8 +227,15 @@ func NewRerunner(ctx context.Context, f ComputeFunc, minRerunInterval time.Durat
 
 // run performs an actual computation
 func (r *Rerunner) run() {
+	// Wait for the minimum rerun interval. Exit early if the computation is stopped.
 	delta := r.minRerunInterval - time.Now().Sub(r.lastRun)
-	time.Sleep(delta)
+	t := time.NewTimer(delta)
+	select {
+	case <-r.ctx.Done():
+		t.Stop()
+		return
+	case <-t.C:
+	}
 
 	r.mu.Lock()
 	defer r.mu.Unlock()


reactive/rerunner.go
```diff
@@ -228,7 +228,19 @@ func NewRerunner(ctx context.Context, f ComputeFunc, minRerunInterval time.Durat
...
+	select {
+	case <-r.ctx.Done():
+		// r.Stop was called while we were waiting for minRerunInterval.
+		// This can happen, for example, if a client specifies
+		// minRerunInterval = math.MaxInt64 in an attempt to simulate
+		// one-shot behavior.
+		t.Stop()
+		return
+	case <-t.C:
+		break
```
jellevandenhooff: No need to break.

reactive/rerunner.go
```diff
@@ -228,7 +228,19 @@ func NewRerunner(ctx context.Context, f ComputeFunc, minRerunInterval time.Durat
 // run performs an actual computation
 func (r *Rerunner) run() {
 	delta := r.minRerunInterval - time.Now().Sub(r.lastRun)
-	time.Sleep(delta)
+
+	t := time.NewTimer(delta)
+	select {
+	case <-r.ctx.Done():
+		// r.Stop was called while we were waiting for minRerunInterval.
```
jellevandenhooff: Remove this comment.

reactive/rerunner.go
```diff
@@ -228,7 +228,19 @@ func NewRerunner(ctx context.Context, f ComputeFunc, minRerunInterval time.Durat
 // run performs an actual computation
 func (r *Rerunner) run() {
 	delta := r.minRerunInterval - time.Now().Sub(r.lastRun)
-	time.Sleep(delta)
+
```
jellevandenhooff: Remove this newline.

reactive/rerunner.go
```diff
@@ -228,7 +228,19 @@ func NewRerunner(ctx context.Context, f ComputeFunc, minRerunInterval time.Durat
 // run performs an actual computation
 func (r *Rerunner) run() {
```
jellevandenhooff: Add a comment here `// Wait for the minimum rerun interval. Exit early if the computation is stopped.`

silverspace:
Thanks for the review, I've updated the comments and style nits.

PR #44, opened 2017-08-03
merge: Implement merging update diffs into an existing object.

kav-ya:
Test Plan:
Minimal tests added; more upcoming.

### DIFF
--- a/merge/merge.go
+++ b/merge/merge.go
@@ -1,0 +1,171 @@
+package diff
+
+import (
+	"log"
+	"strconv"
+)
+
+const indicesReorderedKey = "$"
+
+// Merge merges an update diff into the previous version of a JSON object,
+// and returns the updated version.
+//
+// It effectively reverses the diff algorithm implemented in package diff.
+func Merge(prev map[string]interface{}, diff map[string]interface{}) interface{} {
+	new := make(map[string]interface{})
+
+	// Update exisiting fields.
+	for k, v := range prev {
+		d, ok := diff[k]
+		// No change.
+		if !ok {
+			new[k] = v
+			continue
+		}
+		// Removed.
+		if isRemoved(d) {
+			continue
+		}
+		// Updated.
+		var newV interface{}
+		switch v := v.(type) {
+		case map[string]interface{}:
+			newV = MergeMap(v, d)
+		case []interface{}:
+			newV = MergeArray(v, d)
+		default:
+			newV = MergeReplaced(d)
+		}
+		new[k] = newV
+	}
+
+	// Merge in new fields.
+	for k, d := range diff {
+		if _, ok := prev[k]; !ok {
+			new[k] = MergeReplaced(d)
+		}
+	}
+
+	return new
+}
+
+// MergeMap applies a delta to a map field.
+func MergeMap(prev map[string]interface{}, delta interface{}) interface{} {
+	d, ok := delta.(map[string]interface{})
+	if !ok {
+		return MergeReplaced(delta)
+	}
+	return Merge(prev, d)
+}
+
+// MergeArray applies a delta to an array field.
+func MergeArray(prev []interface{}, delta interface{}) interface{} {
+	d, ok := delta.(map[string]interface{})
+	if !ok {
+		return MergeReplaced(delta)
+	}
+
+	// Reorder elements if needed.
+	var new []interface{}
+	if compressedIndices, ok := d[indicesReorderedKey]; ok {
+		reorderedIndices := uncompressIndices(compressedIndices)
+		new = make([]interface{}, len(reorderedIndices))
+		for i, index := range reorderedIndices {
+			new[i] = prev[index]
+		}
+	} else {
+		new = make([]interface{}, len(prev))
+		for i := range prev {
+			new[i] = prev[i]
+		}
+	}
+
+	// Update complex elements.
+	for k := range d {
+		if k == indicesReorderedKey {
+			continue
+		}
+
+		diff, ok := d[k].(map[string]interface{})
+		if !ok {
+			log.Printf("MergeArray: diff of key %s is not a map[string]interface{}: %v",
+				k, d[k])
+			return nil
+		}
+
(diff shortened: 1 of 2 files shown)


merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+	"strconv"
+)
+
+const indicesReorderedKey = "$"
+
+// Merge merges an update diff into the previous version of a JSON object,
+// and returns the updated version.
+//
+// It effectively reverses the diff algorithm implemented in package diff.
+func Merge(prev map[string]interface{}, diff map[string]interface{}) interface{} {
```
jellevandenhooff: Should these arguments be map[string]interface{} or interface{}?

merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+			new[k] = v
+			continue
+		}
+		// Removed.
+		if isRemoved(d) {
+			continue
+		}
+		// Updated.
+		var newV interface{}
+		switch v := v.(type) {
```
jellevandenhooff: I think this switch would fit more cleanly at the top of the function (if you make Merge take interface{} as arguments).

merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+// and returns the updated version.
+//
+// It effectively reverses the diff algorithm implemented in package diff.
+func Merge(prev map[string]interface{}, diff map[string]interface{}) interface{} {
+	new := make(map[string]interface{})
+
+	// Update exisiting fields.
+	for k, v := range prev {
+		d, ok := diff[k]
+		// No change.
```
jellevandenhooff: Could you put these comments inside the if statement?

merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+		if !ok {
+			new[k] = v
+			continue
+		}
```
jellevandenhooff: I think this would read more nicely as an else-if chain without continues

merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+		if _, ok := prev[k]; !ok {
+			new[k] = MergeReplaced(d)
+		}
+	}
+
+	return new
+}
+
+// MergeMap applies a delta to a map field.
+func MergeMap(prev map[string]interface{}, delta interface{}) interface{} {
```
jellevandenhooff: Could you make all but the Merge function private?

merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		// Updated.
+		var newV interface{}
+		switch v := v.(type) {
+		case map[string]interface{}:
+			newV = MergeMap(v, d)
+		case []interface{}:
+			newV = MergeArray(v, d)
+		default:
+			newV = MergeReplaced(d)
```
jellevandenhooff: There's a lot of calls to MergeReplaced. I wonder if we could simplify the overall logic and replace those individual checks with single check before the switch statement that is "is this a slice, if so, this is a replacement".

merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Update complex elements.
+	for k := range d {
+		if k == indicesReorderedKey {
+			continue
+		}
+
+		diff, ok := d[k].(map[string]interface{})
+		if !ok {
+			log.Printf("MergeArray: diff of key %s is not a map[string]interface{}: %v",
```
jellevandenhooff: I worry that these log statements will cause silent failures. I am fine with either panicing or making the function also return an error.

merge/merge_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestMergeReplaced(t *testing.T) {
+	cases := []struct {
+		Case        string
+		Prev        map[string]interface{}
```
jellevandenhooff: Some of these tests might be easier to read by writing the test cases as raw JSON instead of Go types.

merge/merge_test.go
```diff
@@ -0,0 +1,127 @@
...
+				},
+			},
+			ExpectedNew: map[string]interface{}{
+				"friends": []interface{}{"eli"},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		new := Merge(c.Prev, c.Diff)
```
jellevandenhooff: I would try and reuse the machinery between the different tests.

merge/merge.go
```diff
@@ -0,0 +1,171 @@
...
+	compressedIndices, ok := indices.([]interface{})
+	if !ok {
+		log.Printf("uncompressIndices: indices is not an array: %v", indices)
+		return nil
+	}
+
+	var uncompressedIndices []int
+	for _, index := range compressedIndices {
+		switch index := index.(type) {
+		case []int:
```
jellevandenhooff: I think unmarshal will give you a []interface{} instead of an []int.

merge/merge.go
kav-ya: done.

merge/merge.go
kav-ya: done.

merge/merge.go
kav-ya: I want them to be testable, but our test live in a different package.

merge/merge.go
kav-ya: Updated to panics.

merge/merge.go
kav-ya: See below.

merge/merge.go
kav-ya: +1, done.

merge/merge.go
kav-ya: They don't need to be public anymore (after the restructuring); updated to be private.

merge/merge_test.go
kav-ya: done.

merge/merge.go
kav-ya: Yes, and `float64` for numbers too. Updating the tests to use JSON caught both of these.

PR #49, opened 2017-08-18
Non-nullable resolver functions

stephen:
This PR allows resolvers to be marked non-nullable. By default, pointer types will continue to be nullable, but can be marked otherwise.

```golang
object.FieldFunc("optional", function(ctx context.Context) (*User, object) {
  // ...
})

object.FieldFunc("required", function(ctx context.Context) (*User, object) {
  // ...
}, schemabuilder.NonNullable)
```

### DIFF
--- a/graphql/schemabuilder/reflect.go
+++ b/graphql/schemabuilder/reflect.go
@@ -369,7 +369,8 @@ func init() {
 	selectionSetType = reflect.TypeOf(selectionSet)
 }
 
-func (sb *schemaBuilder) buildFunction(typ reflect.Type, fun reflect.Value) (*graphql.Field, error) {
+func (sb *schemaBuilder) buildFunction(typ reflect.Type, m *method) (*graphql.Field, error) {
+	fun := reflect.ValueOf(m.Fn)
 	ptr := reflect.PtrTo(typ)
 
 	if fun.Kind() != reflect.Func {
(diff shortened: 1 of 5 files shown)


graphql/schemabuilder/reflect.go
```diff
@@ -507,6 +518,9 @@ func (sb *schemaBuilder) buildFunction(typ reflect.Type, fun reflect.Value) (*gr
 			var result interface{}
 			if hasRet {
 				result = out[0].Interface()
+				if out[0].Kind() == reflect.Ptr && out[0].IsNil() && !m.MarkedNullable {
```
stephen: this should probably check that `retType` is `NonNull` instead of checking `MarkedNullable`?

stephen: think more about this, i really wish this was enforced by the type system, instead of at runtime.

YosubShin: ^ I was thinking about the same thing and I wonder if we can just return pointer type for nullable type and a non-pointer type for non-nullable. (this is what we do today)
Are there other benefits of this PR's approach that I may have overlooked that can't be achieved by what we're doing today?

graphql/schemabuilder/types.go
```diff
@@ -29,16 +35,27 @@ type Object struct {
 //        userID, err := db.AddUser(ctx, args.FirstName, args.LastName)
 //        return userID, err
 //    })
-func (s *Object) FieldFunc(name string, f interface{}) {
+func (s *Object) FieldFunc(name string, f interface{}, options ...func(*method)) {
```
jellevandenhooff: I wish this was an interface FieldFuncOption.

graphql/schemabuilder/reflect.go
jellevandenhooff: I worry having non-pointer types will lead to people writing `return *foo` and causing panics.

stephen: I was previously in the mindset that "structs are safe" to pass around given their lack of `nil` (akin to null checking in ts/flow), but hadn't considered that dereferencing a pointer type (would be common with sqlgen code) is still not a statically checked operation in golang.

PR #52, opened 2017-10-10
Parse Selection Args Only Once

bojdell:
Fixes a bug when the same fragment is used multiple times
in a query and the args of its selections are parsed
multiple times, causing query preparation errors
(I observed [this one](https://github.com/samsarahq/thunder/blob/master/graphql/schemabuilder/reflect.go#L304) specifically).

Tested manually. Also ran all unit tests locally and they passed.

### DIFF
--- a/graphql/executor.go
+++ b/graphql/executor.go
@@ -90,11 +90,15 @@ func PrepareQuery(typ Type, selectionSet *SelectionSet) error {
 				return NewSafeError(`unknown field "%s"`, selection.Name)
 			}
 
-			parsed, err := field.ParseArguments(selection.Args)
-			if err != nil {
-				return NewSafeError(`error parsing args for "%s": %s`, selection.Name, err)
+			// Only parse args once for a given selection.
+			if !selection.parsed {
+				parsed, err := field.ParseArguments(selection.Args)
+				if err != nil {
+					return NewSafeError(`error parsing args for "%s": %s`, selection.Name, err)
+				}
+				selection.Args = parsed
+				selection.parsed = true
 			}
-			selection.Args = parsed
 
 			if err := PrepareQuery(field.Type, selection.SelectionSet); err != nil {
 				return err
--- a/graphql/types.go
+++ b/graphql/types.go
@@ -140,6 +140,10 @@ type Selection struct {
 	Alias        string
 	Args         interface{}
 	SelectionSet *SelectionSet
+
+	// The parsed flag is used to make sure the args for this Selection are only
+	// parsed once.
+	parsed bool
 }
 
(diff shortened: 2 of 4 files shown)


graphql/types.go
```diff
@@ -135,10 +135,13 @@ type SelectionSet struct {
 // (representing the name to be used in the output), args id: 166 (representing
 // arguments passed to the source field to be queried), and subselection name
 // representing the information to be queried from the resulting object.
+// The Parsed flag is used to make sure the args for this Selection are only
+// parsed once.
 type Selection struct {
 	Name         string
 	Alias        string
 	Args         interface{}
+	Parsed       bool
```
stephen: should we restrict access to this variable?

also, could you co-locate the comment above closer to the field?

livesql/binlog.go
```diff
@@ -88,7 +88,7 @@ func NewBinlog(ldb *LiveDB, host string, port uint16, username, password, databa
 		return nil, err
 	}
 
-	syncer := replication.NewBinlogSyncer(&replication.BinlogSyncerConfig{
+	syncer := replication.NewBinlogSyncer(replication.BinlogSyncerConfig{
```
stephen: why is this change necessary?

graphql/executor.go
```diff
@@ -90,11 +90,15 @@ func PrepareQuery(typ Type, selectionSet *SelectionSet) error {
...
-			if err != nil {
-				return NewSafeError(`error parsing args for "%s": %s`, selection.Name, err)
+			// Only parse args once for a given selection.
+			if !selection.Parsed {
+				parsed, err := field.ParseArguments(selection.Args)
+				if err != nil {
+					return NewSafeError(`error parsing args for "%s": %s`, selection.Name, err)
+				}
+				selection.Args = parsed
+				selection.Parsed = true
```
stephen: can you add a test for this behavior?

livesql/binlog.go
bojdell: I was just trying to fix the [build](https://travis-ci.org/samsarahq/thunder/builds/285862909?utm_source=github_status&utm_medium=notification) on my first commit. Seems like this API was changed [~22 days ago](https://github.com/siddontang/go-mysql/pull/163). Alternative would be to perhaps vendor this dependency

graphql/types.go
bojdell: Ah, package-private i.e. `parsed`? Seems reasonable to me

Yes, will move comment!

graphql/executor.go
bojdell: Sure!

bojdell:
@stephen I think I've addressed all your comments - PTAL when you have a second!

PR #53, opened 2017-11-02
Ryanr/scalar: Unwrap pointers to scalar values

ryanr23:
Updates the executor to unwrap pointers to scalar values to their raw values if the pointer is non-nil. This fixes an issue where pointers were being compared when calculating update diffs instead of underlying values leading to unnecessary updates.

Executor and diff tests were updated to verify the correct behavior.

### DIFF
--- a/diff/diff_test.go
+++ b/diff/diff_test.go
@@ -50,6 +50,28 @@ func TestStripKey(t *testing.T) {
 	}
 }
 
+func TestDiffIntNullable(t *testing.T) {
+	var testcases = []struct {
+		desc         string
+		left         interface{}
+		right        interface{}
+		expectedDiff interface{}
+	}{
+		{"both nil", nil, nil, nil},
+		{"nil, one", nil, 1, 1},
+		{"one, nil", 1, nil, []interface{}{nil}},
+		{"both one", 1, 1, nil},
+	}
+
+	for _, tc := range testcases {
+		d := diff.Diff(tc.left, tc.right)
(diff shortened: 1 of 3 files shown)


graphql/executor_test.go
```diff
@@ -119,6 +131,14 @@ func TestBasic(t *testing.T) {
 		t.Error(err)
 	}
 
+	// assert that result["as][1]["valuePtr"] == 1 (and not a pointer to 1)
```
stephen: nit:
```diff
- result["as][1]["valuePtr"]
+ result["as"][1]["valuePtr"]
```

diff/diff_test.go
```diff
@@ -50,6 +50,28 @@ func TestStripKey(t *testing.T) {
...
+	var testcases = []struct {
+		desc         string
+		left         interface{}
+		right        interface{}
+		expectedDiff interface{}
+	}{
+		{"both nil", nil, nil, nil},
+		{"nil, one", nil, 1, 1},
+		{"one, nil", 1, nil, []interface{}{nil}},
+		{"both one", 1, 1, nil},
```
stephen: should we also test diffing int pointer to int?

ryanr23: With this change, don't we eliminate that as valid input to the diff layer?

graphql/executor_test.go
ryanr23: oops, good catch

diff/diff_test.go
stephen: oh yes, good point. I had the wrong mindset reading this test.

PR #55, opened 2017-11-08
Distinguish ClientError from SafeError

stephen:
This PR distinguishes parsing errors (client input errors) from safe errors used otherwise.

### DIFF
--- a/graphql/server.go
+++ b/graphql/server.go
@@ -86,12 +86,26 @@ type SafeError struct {
 	message string
 }
 
+type ClientError SafeError
+
+func (e ClientError) Error() string {
+	return e.message
+}
+
+func (e ClientError) SanitizedError() string {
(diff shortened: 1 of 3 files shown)


bojdell:
Also we should figure out how we can get Travis passing

graphql/server.go
```diff
@@ -86,6 +86,16 @@ type SafeError struct {
 	message string
 }
 
+type ClientError SafeError
+
+func (e ClientError) Error() string {
```
changpingc: do we need to redeclare Error() and SanitizedError()?

stephen: strangely yes
```
graphql/server.go:100:20: cannot use ClientError literal (type ClientError) as type error in return argument:
	ClientError does not implement error (missing Error method)
```

PR #59, opened 2017-11-27
graphql: return name/kind on parsing failures

stephen:
This can be helpful for error reporting which query is failing to be parsed.

### DIFF
--- a/graphql/parser.go
+++ b/graphql/parser.go
@@ -374,6 +374,12 @@ func Parse(source string, vars map[string]interface{}) (*Query, error) {
 		name = queryDefinition.Name.Value
 	}
 
+	rv := &Query{
+		Name:         name,
+		Kind:         kind,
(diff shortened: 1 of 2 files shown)


graphql/parser.go
```diff
@@ -384,31 +390,29 @@ func Parse(source string, vars map[string]interface{}) (*Query, error) {
 	for name, fragment := range fragmentDefinitions {
 		selectionSet, err := parseSelectionSet(fragment.SelectionSet, globalFragments, vars)
 		if err != nil {
-			return nil, err
+			return rv, err
```
changpingc: Not a big fan of returning semi valid data and error at the same time, but I can't think of a simple alternative.

stephen: somethingsomethingoptions. I think this fine, since callers should be always checking for `err` anyways (vs. `query == nil` or something).

PR #60, opened 2017-11-30
graphql: let all subscriptions immediately after a mutation

jellevandenhooff:
Removes the lag between making a change and seeing the change appear.
Tested locally with the example application. It's much nicer.

### DIFF
--- a/reactive/rerunner_test.go
+++ b/reactive/rerunner_test.go
@@ -351,3 +351,35 @@ func TestMinRerunInterval(t *testing.T) {
 	r.Invalidate()
 	run.Expect(t, "expected rerun")
 }
+
+// TestRerunImmediately tests that RerunImmediately bypasses the
+// rerun delay.
+func TestRerunImmediately(t *testing.T) {
+	run := NewExpect()
+
+	r := NewResource()
+	var ran time.Time
+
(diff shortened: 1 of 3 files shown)


reactive/rerunner_test.go
```diff
@@ -351,3 +351,35 @@ func TestMinRerunInterval(t *testing.T) {
...
+	runner := NewRerunner(context.Background(), func(ctx context.Context) (interface{}, error) {
+		AddDependency(ctx, r)
+		run.Trigger()
+
+		if ran.IsZero() {
+			ran = time.Now()
+		} else {
+			delta := time.Now().Sub(ran)
+			if delta > 1*time.Second {
+				t.Error("expected imediate rerun")
```
stephen: `s/imediate/immediate/`

bojdell:
> It'd be nice to only flush queries that are impacted by the mutation

Yeah I was trying to think of a way to do this too, but couldn't come up with anything other than what you suggested

jellevandenhooff:
For mutations that aren't impacted, they won't rerun anyway until they get invalidated by the DB or some other mechanism. Likely their rerun delay will have passed by then anyway, so there should be no performance impact.

PR #64, opened 2017-12-12
diff: mark root map/array deletions correctly

stephen:
If a map/slice is removed, then we should mark the node as replaced with nil, instead of replaced with an empty map/slice.


stephen:
@jenniferjzhang can you check if this fixes your issue locally?

jenniferjzhang:
Yeah sure thing. I'll do this later today. @stephen

YosubShin:
👍

PR #65, opened 2017-12-12
diff: always use append to start diff indices

stephen:
Follow up from #62.

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -204,17 +204,13 @@ func computeReorderIndices(old, new []interface{}) []int {
 	oldIndices := make(map[interface{}][]int)
 	for i, item := range old {
 		key := reorderKey(item)
-		if oldIndices[key] == nil {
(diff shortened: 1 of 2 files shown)


diff/diff.go
```diff
@@ -204,17 +204,13 @@ func computeReorderIndices(old, new []interface{}) []int {
...
-			oldIndices[key] = append(oldIndices[key], i)
-		}
+		oldIndices[key] = append(oldIndices[key], i)
 	}
 
 	indices := make([]int, len(new))
 	for i, item := range new {
 		key := reorderKey(item)
-		if index, ok := oldIndices[key]; ok {
+		if index, ok := oldIndices[key]; ok && len(index) > 0 {
```
jellevandenhooff: do index := oldIndices[key]; len(index) > 0. The default value is nil which has len(0)

stephen: fixed, thanks!

PR #68, opened 2017-12-14
graphql: move query parsing into middleware

stephen:
This might help with issues where downstream middlewares are not called if the query parsing step fails.

### DIFF
--- a/graphql/server.go
+++ b/graphql/server.go
@@ -159,20 +159,6 @@ func (c *conn) handleSubscribe(id string, subscribe *subscribeMessage) error {
 
 	tags := map[string]string{"url": c.url, "query": subscribe.Query, "queryVariables": mustMarshalJson(subscribe.Variables), "id": id}
 
-	query, err := Parse(subscribe.Query, subscribe.Variables)
-	if query != nil {
-		tags["queryType"] = query.Kind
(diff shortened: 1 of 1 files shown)


graphql/server.go
```diff
@@ -189,18 +175,36 @@ func (c *conn) handleSubscribe(id string, subscribe *subscribeMessage) error {
 		var middlewares []MiddlewareFunc
 		middlewares = append(middlewares, c.middlewares...)
 		middlewares = append(middlewares, func(input *ComputationInput, next MiddlewareNextFunc) *ComputationOutput {
+			if input.ParsedQuery == nil {
+				query, err := Parse(subscribe.Query, subscribe.Variables)
+				if query != nil {
+					tags["queryType"] = query.Kind
+					tags["queryName"] = query.Name
```
stephen: maybe we should make tags available to the middleware inputs as well.

graphql/server.go
```diff
@@ -189,18 +175,36 @@ func (c *conn) handleSubscribe(id string, subscribe *subscribeMessage) error {
 		var middlewares []MiddlewareFunc
```
stephen: the logger (right above this) is not going to have the right tags.

PR #70, opened 2018-01-12
Allow undefined variable in query

oliversong:
Before:
screen shot 2018-01-12 at 4 10 28 pm

After:
screen shot 2018-01-12 at 4 10 38 pm

### DIFF
--- a/graphql/end_to_end_test.go
+++ b/graphql/end_to_end_test.go
@@ -201,3 +201,74 @@ func TestEndToEndAwaitAndCache(t *testing.T) {
 		t.Errorf("expected 4 total calls to slow, got %d", calls)
 	}
 }
+
+func verifyArgumentOption(t *testing.T, query graphql.Type, queryString string, variables map[string]interface{}, expectedResult string) {
+	q := graphql.MustParse(queryString, variables)
+
+	if err := graphql.PrepareQuery(query, q.SelectionSet); err != nil {
+		t.Error(err)
+	}
+
+	e := graphql.Executor{}
+	result, err := e.Execute(context.Background(), query, nil, q)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if !reflect.DeepEqual(internal.AsJSON(result), internal.ParseJSON(expectedResult)) {
+		t.Error(internal.AsJSON(result))
+	}
+}
+
+// TestArgumentOptionality tests that optional arguments can be omitted from
+// query variables and that mandatory arguments must be included.
+func TestArgumentOptionality(t *testing.T) {
+	schema := schemabuilder.NewSchema()
+	query := schema.Query()
+
+	query.FieldFunc("optional", func(args struct{ X *int64 }) int64 {
+		if args.X != nil {
+			return *args.X
+		}
+		return -1
+	})
+
+	query.FieldFunc("mandatory", func(args struct{ X int64 }) int64 {
+		return args.X
+	})
+
+	_ = schema.Mutation()
+	builtSchema := schema.MustBuild()
+	emptyVariables := map[string]interface{}{}
+	filledVariables := map[string]interface{}{
+		"testArg": float64(5),
+	}
+
+	// An optional argument that is passed in returns successfully.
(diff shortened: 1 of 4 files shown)


graphql/introspection/introspection.go
```diff
@@ -43,7 +43,7 @@ type InputValue struct {
 	Name         string
 	Description  string
 	Type         Type
-	DefaultValue string
+	DefaultValue *string
```
stephen: can you reference the spec doc in the commit message?

http://facebook.github.io/graphql/October2016/#sec-The-__InputValue-Type

graphql/parser.go
```diff
@@ -36,7 +36,7 @@ func valueToJson(value ast.Value, vars map[string]interface{}) (interface{}, err
 	case *ast.Variable:
 		actual, ok := vars[value.Name.Value]
 		if !ok {
-			return nil, NewClientError("unknown var: %s", value.Name.Value)
+			return nil, nil
```
stephen: let's add tests for the 2x2:
- argument is optional, and argument is not specified in variables
- argument is optional, and argument is specified in variables
- argument is not optional, and argument is not specified in variables
- argument is not optional, and argument is specified in variables

graphql/end_to_end_test.go
```diff
@@ -201,3 +201,104 @@ func TestEndToEndAwaitAndCache(t *testing.T) {
...
+		t.Error(err)
+	}
+
+	result, err = e.Execute(context.Background(), builtSchema.Query, nil, q)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if !reflect.DeepEqual(internal.AsJSON(result), internal.ParseJSON(`
+		{"optional": -1}`)) {
```
stephen: I think I expect an error here, since we passed in nothing, but the variable is required

graphql/end_to_end_test.go
```diff
@@ -201,3 +201,104 @@ func TestEndToEndAwaitAndCache(t *testing.T) {
...
+		}
+		return -1
+	})
+
+	_ = schema.Mutation()
+
+	builtSchema := schema.MustBuild()
+
+	// argument is optional and not passed in
+	q := graphql.MustParse(`
```
stephen: can we clean up the copy/paste in an inner function to make this easier to read?

graphql/end_to_end_test.go
oliversong: Oh you're right. I think I need to make a "mandatory" field that has a matching go function definition

stephen: (i.e. even though none of the users of the argument require it, the query itself should be requiring it.)

oliversong:
Smoke tested cloud. Nothing blew up.

graphql/end_to_end_test.go
```diff
@@ -201,3 +201,72 @@ func TestEndToEndAwaitAndCache(t *testing.T) {
...
+	})
+
+	_ = schema.Mutation()
+	builtSchema := schema.MustBuild()
+	emptyVariables := map[string]interface{}{}
+	filledVariables := map[string]interface{}{
+		"testArg": float64(5),
+	}
+
+	// arg is optional and passed in
```
stephen: nit: prefer full sentences (capital/period) in go comments

graphql/parser.go
```diff
@@ -36,7 +36,7 @@ func valueToJson(value ast.Value, vars map[string]interface{}) (interface{}, err
 	case *ast.Variable:
 		actual, ok := vars[value.Name.Value]
 		if !ok {
```
stephen: As we discussed irl, there is a new bug introduced in this PR that we think we are okay with ignoring for now.

That is, if a variable is marked optional in the operation definition, we do not check for whether or not it is passed in, which goes against the spec. We think this is okay because if the variable was actually required (i.e. used at some location), then the executor will get unhappy about it downstream.

stephen: filed an issue here: https://github.com/samsarahq/thunder/issues/73

oliversong: Awesome. Thanks for writing the issue.

PR #71, opened 2018-01-12
[WIP] graphql: support default variables in queries

oliversong:
This doesn't actually work, because variables is used elsewhere outside the call to Parser. And we still need to cast the default value to the proper type.


stephen:
here's where the reference implementation does this: https://github.com/graphql/graphql-js/blob/master/src/execution/values.js#L78

(before execution and after parsing)

PR #74, opened 2018-01-20
Check variable definition in parse step, support default arguments

stephen:
This PR adds support for checking that non-null operation variables are correctly passed in, fixing #70.

It also adds base support for default values.

### DIFF
--- a/graphql/parser.go
+++ b/graphql/parser.go
@@ -368,6 +368,14 @@ func Parse(source string, vars map[string]interface{}) (*Query, error) {
 		return nil, NewClientError("must have a single query")
 	}
 
+	for _, variableDefinition := range queryDefinition.VariableDefinitions {
+		if _, ok := variableDefinition.Type.(*ast.NonNull); ok {
+			if vars[variableDefinition.Variable.Name.Value] == nil {
+				return nil, NewClientError("required variable not provided: $%s", variableDefinition.Variable.Name.Value)
+			}
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


graphql/parser.go
```diff
@@ -368,6 +368,52 @@ func Parse(source string, vars map[string]interface{}) (*Query, error) {
...
+			// Lazily initialize defaultedVars if needed.
+			if defaultedVars == nil {
+				defaultedVars = make(map[string]interface{})
+				for k, v := range vars {
+					defaultedVars[k] = v
+				}
+			}
+
+			// TODO: properly implement coerceValue.
+			// See: https://github.com/graphql/graphql-js/blob/17a0bfd5292f39cafe4eec5b3bd0e22514243b68/src/execution/values.js#L84
```
stephen: note that without this, we are not properly checking that the default value can be validly used for the value.

graphql/parser_test.go
```diff
@@ -248,3 +248,49 @@ fragment foo on Foo {
...
+	field(x: $x)
+}	`, map[string]interface{}{})
+
+	if err == nil || err.Error() != "required variable not provided: $x" {
+		t.Error("expected unfulfilled required argument to fail, but got", err)
+	}
+}
+
+func TestParseRequiredVariableDefinitionWithDefaultValue(t *testing.T) {
+	// Expect required variables to be provided.
```
oliversong: should this comment say something like "expect required variables not to have default values"?

PR #75, opened 2018-01-22
add support for @skip/@include

stephen:


stephen:
This is not quite there yet. There is a case that looks like:
```
query Test($something: bool) {
  something @skip(if: $something) {
    ...Frag
  }
}

fragment Frag on Something { somethingElse }
```

If `$something` ends up being `true`, then we'll return an error about `unused fragment`. This happens since the selectionSet gets culled before the fragment check happens.

graphql/parser.go
```diff
@@ -138,13 +144,22 @@ func parseSelectionSet(input *ast.SelectionSet, globalFragments map[string]*Frag
 				return nil, NewClientError("unknown fragment")
 			}
 
-			fragments = append(fragments, fragment)
+			directives, err := parseDirectives(selection.Directives, vars)
+			if err != nil {
+				return nil, err
+			}
+
+			fragmentCopy := *fragment
```
stephen: alternatively, we could create a `FragmentInstance` type, but it seems ~equivalent.

shahryart:
what do you think about exposing a single endpoint for adding custom directives?

graphql/end_to_end_test.go
```diff
@@ -830,3 +832,121 @@ func TestConcurrencyLimiterDeadlock(t *testing.T) {
...
+		e := graphql.Executor{}
+		return e.Execute(context.Background(), builtSchema.Query, nil, q)
+	}
+
+	// Variable skip
+	result, err := execute(`
+		query x {
+			value @skip(if: $var)
+		}`, map[string]interface{}{"var": true})
+	if err != nil {
```
seanbae: out of curiosity - is there a reason for preferring the use of `if err != nil` + `t.Errorf` over `assert.NoError` here?

stephen: not really. this code was originally written in jan when we didn't use testify yet.

berfarah:
Is this still something we could/should add?

PR #76, opened 2018-01-26
diff: support diffing byte slices []uint8

jellevandenhooff:
They show up in JSON sometimes to encode binary as base64 data. Allow
that.

### DIFF
--- a/diff/diff_test.go
+++ b/diff/diff_test.go
@@ -29,6 +29,22 @@ func TestDiffListString(t *testing.T) {
 	}
 }
 
+func TestDiffBytes(t *testing.T) {
+	d := diff.Diff([]interface{}{
+		[]byte("123"),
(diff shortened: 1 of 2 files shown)


diff/diff_test.go
```diff
@@ -29,6 +30,23 @@ func TestDiffListString(t *testing.T) {
...
+		[]byte("456"),
+	}, []interface{}{
+		[]byte("123"),
+		[]byte("789"),
+	})
+
+	if !reflect.DeepEqual(internal.AsJSON(d), internal.ParseJSON(`
+		{"1": ["Nzg5"]}
+	`)) {
+		log.Println(d)
```
oliversong: nit: spurious log call?

jellevandenhooff: fixed

PR #83, opened 2018-02-23
Provide a non-live http interface

stephen:
This PR provides an http interface to use thunder for those who do not want to use websockets/live queries but still like all of our other sweet features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,33 @@ func main() {
 }
 ```
 
+## Using Thunder without Websockets (POST requests)
+
+For use with non-live clients (e.g. [Relay](https://facebook.github.io/relay/), [Apollo](https://www.apollographql.com/client/)) thunder provides an HTTP handler that can serve
+POST requests, instead of having the client connect over a websocket. In this mode, thunder
+does not provide live query updates.
+
+In the above example, the `main` function would be changed to look like:
+
+```go
+func main() {
+  // Instantiate a server, build a server, and serve the schema on port 3030.
+  server := &server{
+    posts: []post{
+      {Title: "first post!", Body: "I was here first!", CreatedAt: time.Now()},
+      {Title: "graphql", Body: "did you hear about Thunder?", CreatedAt: time.Now()},
+    },
+  }
+
+  schema := server.schema()
+  introspection.AddIntrospectionToSchema(schema)
+
+  // Expose GraphQL POST endpoint.
+  http.Handle("/graphql", graphql.HTTPHandler(schema))
+  http.ListenAndServe(":3030", nil)
+}
+```
+
 ## Code organization
 
 The source code in this repository is organized as follows:
--- a/graphql/http.go
+++ b/graphql/http.go
@@ -1,0 +1,121 @@
+package graphql
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"sync"
+
+	"github.com/samsarahq/thunder/batch"
+	"github.com/samsarahq/thunder/reactive"
+)
+
+func HTTPHandler(schema *Schema, middlewares ...MiddlewareFunc) http.Handler {
+	return &httpHandler{
+		schema:      schema,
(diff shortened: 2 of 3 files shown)


graphql/http.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/samsarahq/thunder/batch"
+	"github.com/samsarahq/thunder/reactive"
+)
+
+func CreateHTTPHandler(schema *Schema, middlewares ...MiddlewareFunc) http.Handler {
```
jellevandenhooff: is it time to start adding tests for the servers interfaces yet

graphql/http.go
```diff
@@ -0,0 +1,97 @@
...
+		return
+	}
+
+	if err := PrepareQuery(h.schema.Query, query.SelectionSet); err != nil {
+		writeResponse(nil, err)
+		return
+	}
+
+	e := Executor{}
+	runner := reactive.NewRerunner(r.Context(), func(ctx context.Context) (interface{}, error) {
```
jellevandenhooff: wish this stuff wasn't necessary

stephen: I think we might not need to use the re-runner at all. If there is no cache in the context, `reactive.Cache` and `reactive.AddDependency` ignores it.

graphql/http.go
```diff
@@ -0,0 +1,97 @@
...
+
+			writeResponse(nil, err)
+			return nil, err
+		}
+
+		writeResponse(current, nil)
+		return nil, nil
+	}, MinRerunInterval)
+
+	runner.Stop()
```
jellevandenhooff: i forget if it's safe to call this immediately after creating the runner

graphql/http.go
stephen: I think we might not need to use the re-runner at all. If there is no cache in the context, `reactive.Cache` and `reactive.AddDependency` ignores it.

I need to add a test to verify this claim.

graphql/http.go
stephen: yeah let's do it. (for this one)

graphql/http.go
jellevandenhooff: I guess even non-reactive code could benefit from the cache for eg. sql query deduping

stephen: Oh yes, I realized that before but forgot now. It might be nice to separate the cache from the re-runner?

graphql/http.go
stephen: I added some tests. Let's open source the snapshotter tool, because I want to use it.

graphql/http.go
stephen: I added a wait group since the re-runner starts a new goroutine.

stephen:
@jellevandenhooff ptal, i also updated the readme

README.md
```diff
@@ -295,6 +295,33 @@ func main() {
 }
 ```
 
+## Using POST requests
+
+For use with non-live clients (relay, apollo, lokka), thunder provides an HTTP handler
```
jellevandenhooff: just do relay, apollo, maybe capitalize, maybe have links.

README.md
```diff
@@ -295,6 +295,33 @@ func main() {
 }
 ```
 
+## Using POST requests
```
jellevandenhooff: Using Thunder without reactivity? Using Thunder without websockets?

graphql/http_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"strings"
+	"testing"
+
+	"github.com/kylelemons/godebug/pretty"
+
+	"github.com/samsarahq/thunder/graphql"
+	"github.com/samsarahq/thunder/graphql/schemabuilder"
+)
+
+func runRequest(req *http.Request) *httptest.ResponseRecorder {
```
jellevandenhooff: testHTTPRequest?

PR #85, opened 2018-03-07
graphql: parse name/kind before variables

stephen:
This PR changes the parser to parser the query kind/name earlier so that it will show up in case of client parsing errors.

### DIFF
--- a/graphql/parser.go
+++ b/graphql/parser.go
@@ -368,6 +368,18 @@ func Parse(source string, vars map[string]interface{}) (*Query, error) {
 		return nil, NewClientError("must have a single query")
 	}
 
+	kind := queryDefinition.Operation
+	var name string
+	if queryDefinition.Name != nil {
+		name = queryDefinition.Name.Value
+	}
+
+	rv := &Query{
+		Name:         name,
+		Kind:         kind,
+		SelectionSet: nil,
+	}
+
 	// Parse variable definitions, default values, etc.
 	var defaultedVars map[string]interface{}
 	for _, variableDefinition := range queryDefinition.VariableDefinitions {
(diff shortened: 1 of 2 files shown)


graphql/parser.go
```diff
@@ -368,18 +368,30 @@ func Parse(source string, vars map[string]interface{}) (*Query, error) {
...
+
 	// Parse variable definitions, default values, etc.
 	var defaultedVars map[string]interface{}
 	for _, variableDefinition := range queryDefinition.VariableDefinitions {
 		name := variableDefinition.Variable.Name.Value
 
 		if _, ok := variableDefinition.Type.(*ast.NonNull); ok {
 			if variableDefinition.DefaultValue != nil {
-				return nil, NewClientError("required varaible cannot provide a default value: $%s", name)
+				return rv, NewClientError("required variable cannot provide a default value: $%s", name)
```
jenniferjzhang: 🎉

jenniferjzhang: We are sure that the function that calls Parse looks for non-nil error instead of nil result?

graphql/parser_test.go
```diff
@@ -268,7 +268,7 @@ query Operation($x: int64! = 2) {
 	field(x: $x)
 }	`, map[string]interface{}{})
 
-	if err == nil || err.Error() != "required varaible cannot provide a default value: $x" {
+	if err == nil || err.Error() != "required variable cannot provide a default value: $x" {
```
jenniferjzhang: wow there are a lot of misspellings of variable

graphql/parser.go
stephen: yeah, we do this in other places downstream as wel

graphql/parser_test.go
stephen: yeah i am embarrassed

Issue #87, opened 2018-03-25
unable to compile schema_generator.go example


JonLundy:
Hi, I am trying to compile the provided example for printing out the generated schema but it seems that `introspection.ComputeSchemaJSON()` does not accept `*graphql.Schema`

```
	server := &server{}
	schema := server.schema()
	introspection.AddIntrospectionToSchema(schema)
	valueJSON, err := introspection.ComputeSchemaJSON(*schema)
	if err != nil {
		panic(err)
	}

	fmt.Sprint(string(valueJSON))
```
results in 

```
cannot use *schema (variable of type github.com/samsarahq/thunder/graphql.Schema) as github.com/samsarahq/thunder/graphql/schemabuilder.Schema value in argument to introspection.ComputeSchemaJSON
```

JonLundy:
Digging in a bit more.. it seems i should be passing `*builder` that is created in `server.schema()` instead of `*schema`

stephen:
Good point. This part is a bit unclear, but it should be taking the `*schemabuilder.Schema` and pulling a built `graphql.Schema` out of it. Adding the introspection schema (`__typename`, etc) should be part of "building" the schema, whereas the introspection json (schema.json) is computed off of the built schema.

stephen:
Going to close this, but feel free to comment if #88 isn't enough.

PR #88, opened 2018-03-27
readme: attempt to make schema dumper example more clear

stephen:
Attempt to address #87. Perhaps we should consider an API change here so that `ComputeSchemaJSON` does not also call `AddIntrospectionToSchema`..

@jonlundy does this help?


stephen:
cc/ @jellevandenhooff

PR #89, opened 2018-03-30
concurrencylimiter: bound number of concurrently running resolvers

jellevandenhooff:
The new concurrencylimiter package lets the graphql executor limit the number
of resolvers that are running at the same time. It integrates with the batch
package to allow resolvers that are joining a batched call to unblock other
resolvers.

All of this is opt-in using the concurrencylimiter.With call. Otherwise
everything behaves like before.

### DIFF
--- a/batch/batch.go
+++ b/batch/batch.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"sync"
 	"time"
+
+	"github.com/samsarahq/thunder/concurrencylimiter"
 )
 
 // DefaultWaitInterval is the default WaitInterval for Func.
@@ -225,8 +227,10 @@ func (f *Func) Invoke(ctx context.Context, arg interface{}) (interface{}, error)
 		close(bg.doneCh)
 
 	} else {
-		// Wait for the result.
-		<-bg.doneCh
+		concurrencylimiter.Block(ctx, func() {
+			// Wait for the result.
+			<-bg.doneCh
+		})
 	}
 
 	// Return the local result.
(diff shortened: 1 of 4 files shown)


concurrencylimiter/concurrencylimiter.go
```diff
@@ -0,0 +1,121 @@
...
+		status: acquired,
+	}
+	ctx = context.WithValue(ctx, holderKey{}, h)
+
+	return ctx, h.release
+}
+
+// Block temporarily releases a concurrency limiter token (if any) while calling
+// a long-running but no-resource-using function f.
+func Block(ctx context.Context, f func()) {
```
stephen: this name is a little strange to me. it's not quite blocking - more like opting that section of computation out.

concurrencylimiter/concurrencylimiter.go
```diff
@@ -0,0 +1,121 @@
...
+
+	select {
+	case l.ch <- struct{}{}:
+	case <-ctx.Done():
+		return ctx, func() {}
+	}
+
+	h := &holder{
+		l: l,
+
```
stephen: nit: no newline?

batch/batch.go
```diff
@@ -225,8 +227,10 @@ func (f *Func) Invoke(ctx context.Context, arg interface{}) (interface{}, error)
 		close(bg.doneCh)
 
 	} else {
-		// Wait for the result.
-		<-bg.doneCh
+		concurrencylimiter.Block(ctx, func() {
```
stephen: this should be `TemporarilyRelease`

PR #90, opened 2018-03-30
graphql: add support for metadata on incoming envelopes

stephen:
Useful for passing arbitrary metadata down. For instance, we can use this to send request ids or trace ids down to the server.

Other ideas for this kind of mechanism are welcome.


stephen:
One interesting note - reading the [graphql spec](http://facebook.github.io/graphql/October2016/#sec-Response-Format), we should have called the output metadata `extensions` instead of `metadata` to be spec compliant. It's a little too late now without causing a breaking change.

stephen:
hm, not very for us.

PR #94, opened 2018-04-10
graphql schemabuilder: gracefully fail if wrong argument type is passed

stephen:
cc/ @jbicket

It's a little tricky to provide a better message in case the caller is actually intending the parent "source" instead of an arguments struct since the function signature is overloaded. This at least helps with identifying what field is wrong.

Before, this would just panic and not identify the bad field.


stephen:
here's a second attempt that is a bit more verbose, but is hopefully more transparent that it's parsing the argument as field arguments instead of as a source:
```
```

PR #95, opened 2018-04-11
go livesql: add support for replica db

changpingc:


livesql/binlog.go
```diff
@@ -302,9 +312,26 @@ func (b *Binlog) parseBinlogRowsEvent(event *replication.BinlogEvent) (*update,
...
+func (b *Binlog) SetUpdateDelay(d time.Duration) {
+	b.delayMu.Lock()
+	defer b.delayMu.Unlock()
+	b.delay = d
+}
+
 // RunPollLoop is the core binlog function that fetches and distributes updates
 // from MySQL
 func (b *Binlog) RunPollLoop() error {
+	updateCh := make(chan delayedUpdate, 1024)
```
changpingc: alternative magic number 1024 is creating unbound number of goroutines here.

jellevandenhooff: alternatively, the poll loop itself could sleep?

livesql/binlog.go
```diff
@@ -71,18 +71,18 @@ func getPosition(conn *sql.DB) (mysql.Position, error) {
 //
 // NewBinlog verifies that the given DB has been correctly configured for
 // streaming changes.
-func NewBinlog(ldb *LiveDB, host string, port uint16, username, password, database string) (*Binlog, error) {
+func NewBinlog(ldb *LiveDB, binlogDB *sql.DB, host string, port uint16, username, password, database string) (*Binlog, error) {
```
stephen: should we add a new `NewBinlogWithSource` or similar so that this isn't a breaking change?

livesql/binlog.go
changpingc: poll loop shouldn't sleep because we don't want to wait replica lag between every binlog update

livesql/binlog.go
changpingc: sg

changpingc: done

PR #96, opened 2018-04-20
subscription logger + CreateConnection api

stephen:


graphql/server.go
```diff
@@ -500,6 +518,73 @@ func CreateJSONSocketWithMutationSchema(ctx context.Context, socket JSONSocket,
...
+func (l *noopSubscriptionLogger) Subscribe(ctx context.Context, id string, tags map[string]string) {}
+func (l *noopSubscriptionLogger) Unsubscribe(ctx context.Context, id string)                       {}
+
+type noopGraphqlLogger struct{}
+
+func (l *noopGraphqlLogger) StartExecution(ctx context.Context, tags map[string]string, initial bool) {
+}
+func (l *noopGraphqlLogger) FinishExecution(ctx context.Context, tags map[string]string, delay time.Duration) {
+}
+func (l *noopGraphqlLogger) Error(ctx context.Context, err error, tags map[string]string) {}
```
changpingc: nso: `nop` instead of `noop` following how `io.NopCloser` is named?

graphql/server.go
```diff
@@ -500,6 +518,73 @@ func CreateJSONSocketWithMutationSchema(ctx context.Context, socket JSONSocket,
...
+		socket:        socket,
+		ctx:           ctx,
+		schema:        schema,
+		subscriptions: make(map[string]*reactive.Rerunner),
+	}
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	if c.subscriptionLogger == nil {
```
changpingc: nso: save some lines by doing this above in struct literal and they will be simply override by functional options

graphql/server.go
```diff
@@ -486,6 +503,7 @@ func CreateJSONSocket(ctx context.Context, socket JSONSocket, schema *Schema, ma
 	}
 }
 
+// CreateJSONSocketWithMutationSchema is deprecated. Consider using CreateConnection instead.
 func CreateJSONSocketWithMutationSchema(ctx context.Context, socket JSONSocket, schema, mutationSchema *Schema, makeCtx MakeCtxFunc, logger GraphqlLogger) *conn {
 	return &conn{
```
changpingc: can we make these call CreateConnection instead?

stephen: done

graphql/server.go
stephen: don

graphql/server.go
stephen: done

PR #99, opened 2018-05-31
sqlgen: introduce TypeConverter, support proto fields

jellevandenhooff:


berfarah:
Nice! I see you added a check to see if the protobuf interface is implemented. I wonder if we could do something similar for JSON and SQL interfaces?
### JSON
https://golang.org/pkg/encoding/json/#Marshaler
https://golang.org/pkg/encoding/json/#Unmarshaler

### JSON fallback
https://golang.org/pkg/encoding/#TextMarshaler
https://golang.org/pkg/encoding/#TextUnmarshaler

### SQL
https://golang.org/pkg/database/sql/#Scanner
https://golang.org/pkg/database/sql/driver/#Valuer

berfarah:
This breaks `livesql/binlog` - I'm investigating alternatives here.

stephen:
could you roughly describe how?

berfarah:
Yep! We're getting rid of scannables here, but `livesql/binlog.go` depends on them. I don't think this abstraction quite works. I think what we need is a type that can give us a SQL type and convert to/from it.

berfarah:
Closing as this has been addressed in https://github.com/samsarahq/thunder/pull/142

PR #106, opened 2018-06-11
client: export GraphQLData, GraphQLError, GraphQLResult

stephen:
Export things that are missing for this to be useful.


stephen:
yes! i had intended to ship another small fix alongside this before bumping the version.

PR #108, opened 2018-06-12
QuerySpec object support

stephen:
Ah, it works.

### DIFF
--- a/client/src/tests/hoc.tsx
+++ b/client/src/tests/hoc.tsx
@@ -1,0 +1,40 @@
+import React from "react";
+import { GraphQLData } from "../connection";
+import { graphql } from "../hoc";
+import { QuerySpec } from "../query";
+
+interface Result {
+  output: string;
+}
+
+interface Variables {
+  y: string;
+}
+
+const exampleQuery: QuerySpec<Result, Variables> = {
+  query: "query test {}",
+};
+
+interface Props extends GraphQLData<Result> {
+  input: string;
+}
+
+const PresentationalComponent = (props: Props) => {
+  return <div>{props.data.value.output}</div>;
+};
+
+const ConnectedComponentWithQueryAndVariables = graphql<
+  Props,
+  Result,
+  Variables
+>("query test {}", props => ({ y: props.input }))(PresentationalComponent);
+
+const instanceWithQueryAndVariables = (
+  <ConnectedComponentWithQueryAndVariables input="test" />
+);
+
+const ConnectedComponentWithSpec = graphql(exampleQuery, props => ({
+  y: "string",
+}))(PresentationalComponent);
+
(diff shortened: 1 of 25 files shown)


client/src/tests/query.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const instanceWithQueryAndVariables = <ComponentWithQueryAndVariables />;
+
+const ComponentWithSpec: React.SFC<{}> = () => {
+  return (
+    <Query query={exampleQuery} variables={{ y: "yes" }}>
+      {data => (
+        <div>
+          {data.state === "subscribed" ? data.value.output : "loading..."}
```
stephen: i'm not actually sure how it's able to pull the type here... the query does not reference the output or input types at all, except implicitly via `exampleQuery`.

client/src/tests/hoc.tsx
```diff
@@ -0,0 +1,40 @@
...
+  Props,
+  Result,
+  Variables
+>("query test {}", props => ({ y: props.input }))(PresentationalComponent);
+
+const instanceWithQueryAndVariables = (
+  <ConnectedComponentWithQueryAndVariables input="test" />
+);
+
+const ConnectedComponentWithSpec = graphql(exampleQuery, props => ({
```
jellevandenhooff: if you inline PresentationalComponent here, does it infer the Props type?

client/src/tests/mutation.tsx
```diff
@@ -0,0 +1,44 @@
...
+      )}
+    </Mutation>
+  );
+};
+
+const instanceWithQueryAndVariables = <ComponentWithQueryAndVariables />;
+
+const ComponentWithSpec: React.SFC<{}> = () => {
+  return (
+    <Mutation query={exampleQuery}>
```
jellevandenhooff: this is awesome

client/src/tests/query.tsx
jellevandenhooff: ah it does do the inline! that's great

jellevandenhooff: this also sure is a lot of indentation

client/src/tests/hoc.tsx
stephen: sadly, i realized that it doesn't infer `props` here correctly, even not inlined.

stephen: i think this would be more possible if the signature looked closer to what it used to, e.g. `graphql(Component, document, props => ...)`.

i tried to be too cute and it was a bad idea. see https://github.com/Microsoft/TypeScript/issues/15016#issuecomment-292108004

jellevandenhooff:
Alternatively, we could codegen the entire graphql function...

PR #114, opened 2018-06-27
Steven/graphiql mutation

yuedongze:
graphiql now supports mutation


example/client/src/graphiql.js
```diff
@@ -1,11 +1,17 @@
 import React from 'react';
 import GraphiQL from 'graphiql';
 
-import { connection } from 'thunder-react';
+import { connection,mutate } from 'thunder-react';
```
stephen: nit: space after comma

stephen: we should add prettier checks here..

stephen:
thanks!

PR #118, opened 2018-06-29
Shahryart/typechecking enummap

shahryart:
typecheck enum map

### DIFF
--- a/graphql/schemabuilder/reflect.go
+++ b/graphql/schemabuilder/reflect.go
@@ -818,25 +818,30 @@ func (s *Schema) Enum(val interface{}, enumMap interface{}) {
 		s.enumTypes = make(map[reflect.Type]*EnumMapping)
 	}
 
-	eMap, rMap := getEnumMap(enumMap)
+	eMap, rMap := getEnumMap(enumMap, typ)
 	s.enumTypes[typ] = &EnumMapping{Map: eMap, ReverseMap: rMap}
 }
 
-func getEnumMap(enumMap interface{}) (map[string]interface{}, map[interface{}]string) {
+func getEnumMap(enumMap interface{}, typ reflect.Type) (map[string]interface{}, map[interface{}]string) {
 	rMap := make(map[interface{}]string)
 	eMap := make(map[string]interface{})
 	v := reflect.ValueOf(enumMap)
 	if v.Kind() == reflect.Map {
 		for _, key := range v.MapKeys() {
 			val := v.MapIndex(key)
+			valInterface := val.Interface()
+			if reflect.TypeOf(valInterface).Kind() != typ.Kind() {
+				panic("types are not equal")
+			}
 			if key.Kind() == reflect.String {
-				eMap[key.String()] = val.Interface()
+				mapVal := reflect.ValueOf(valInterface).Convert(typ)
+				eMap[key.String()] = mapVal.Interface()
 			} else {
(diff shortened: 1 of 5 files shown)


graphql/schemabuilder/reflect.go
```diff
@@ -184,6 +185,8 @@ func (sb *schemaBuilder) getEnumArgParser(typ reflect.Type) (*argParser, graphql
 		if !ok {
 			return fmt.Errorf("unknown enum value %v", asString)
 		}
+		log.Println(dest.Type())
+		log.Println(reflect.TypeOf(val))
```
stephen: remove these

graphql/schemabuilder/reflect.go
```diff
@@ -818,19 +821,23 @@ func (s *Schema) Enum(val interface{}, enumMap interface{}) {
...
+			if val.Type().Kind() != typ.Kind() {
+				panic("types are not equal")
+			}
 			if key.Kind() == reflect.String {
-				eMap[key.String()] = val.Interface()
+				mapVal := val.Convert(typ)
+				eMap[key.String()] = mapVal.Interface()
 			} else {
 				panic("keys are not strings")
 			}
```
stephen: can we change the panic below to be lowercased

graphql/schemabuilder/reflect_test.go
```diff
@@ -210,20 +226,65 @@ func TestEnumMapWrongArg(t *testing.T) {
 
 }
 
-func TestEnumMapKeys(t *testing.T) {
+func TestEnumMapWrongArg2(t *testing.T) {
```
stephen: Let's make the test names more descriptive to the case, i.e.`TestEnumMapMismatchedValues`

graphql/schemabuilder/reflect_test.go
```diff
@@ -210,20 +226,65 @@ func TestEnumMapWrongArg(t *testing.T) {
...
 	}()
-	type enumType int32
-	schema.Enum(enumType(1), map[int32]int32{
-		1: 1,
+
+	schema.Enum(enumType3(1), map[string]interface{}{
+		"firstField":  interface{}(1),
+		"secondField": interface{}(2),
+		"thirdField":  interface{}(3),
 	})
```
stephen: i think there are ways we could support this case: https://play.golang.org/p/9bjfb8w_8h2

graphql/schemabuilder/reflect_test.go
```diff
@@ -196,6 +196,22 @@ func TestExecuteGood(t *testing.T) {
 	}
 }
 
+func TestEnumMapKeys(t *testing.T) {
+	schema := NewSchema()
+	defer func() {
+		r := recover()
+		if r == nil {
```
stephen: let's assert on specific error messages here

graphql/schemabuilder/reflect_test.go
shahryart: It doesn't differentiate between int32 and int

stephen: could you paste a code example? https://play.golang.org/p/ItVllolNiF2

shahryart: This particular test case will still fail because the integers aren't of int32 type

stephen: ah, that makes sense.

PR #120, opened 2018-07-11
Fixes for schemabuilder and graphiql

vavrusa:
This fixes some small issues with schemabuilder, and fixes missing support for variables in graphiql.

### DIFF
--- a/graphql/schemabuilder/reflect.go
+++ b/graphql/schemabuilder/reflect.go
@@ -336,7 +336,12 @@ func wrapPtrParser(inner *argParser) *argParser {
 	}
 }
 
-func (sb *schemaBuilder) makeStructParser(typ reflect.Type) (*argParser, graphql.Type, error) {
+func (sb *schemaBuilder) getStructObjectFields(typ reflect.Type) (*graphql.InputObject, map[string]argField, error) {
+	// Check if the struct type is already cached
+	if cached, ok := sb.typeCache[typ]; ok {
+		return cached.argType, cached.fields, nil
+	}
+
 	fields := make(map[string]argField)
 	argType := &graphql.InputObject{
 		Name:        typ.Name(),
@@ -349,6 +354,9 @@ func (sb *schemaBuilder) makeStructParser(typ reflect.Type) (*argParser, graphql
 	if typ.Kind() != reflect.Struct {
 		return nil, nil, fmt.Errorf("expected struct but received type %s", typ.Name())
 	}
+
+	// Cache type information ahead of time to catch self-reference
+	sb.typeCache[typ] = cachedType{argType, fields}
 
 	for i := 0; i < typ.NumField(); i++ {
 		field := typ.Field(i)
@@ -394,6 +402,15 @@ func (sb *schemaBuilder) makeStructParser(typ reflect.Type) (*argParser, graphql
 			parser: parser,
 		}
 		argType.InputFields[name] = fieldArgTyp
+	}
+
+	return argType, fields, nil
+}
+
+func (sb *schemaBuilder) makeStructParser(typ reflect.Type) (*argParser, graphql.Type, error) {
+	argType, fields, err := sb.getStructObjectFields(typ)
+	if err != nil {
+		return nil, nil, err
 	}
 
 	return &argParser{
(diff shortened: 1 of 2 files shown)


graphql/schemabuilder/reflect.go
```diff
@@ -315,63 +325,74 @@ func wrapPtrParser(inner *argParser) *argParser {
...
-		if name == "-" {
-			continue
+
+		if typ.Kind() != reflect.Struct {
+			return nil, nil, fmt.Errorf("expected struct but received type %s", typ.Name())
 		}
 
-		var key bool
+		// Cache type information ahead of time to catch self-reference
+		typeCache[typ] = cachedType{argType, fields}
```
stephen: could you add a test for this this case?

graphql/schemabuilder/reflect.go
```diff
@@ -315,63 +325,74 @@ func wrapPtrParser(inner *argParser) *argParser {
...
-		}
-		if field.Anonymous {
-			return nil, nil, fmt.Errorf("bad arg type %s: anonymous fields not supported", typ)
+	var fields map[string]argField
+	var argType *graphql.InputObject
+	cached, ok := typeCache[typ]
+	if ok {
+		fields = cached.fields
+		argType = cached.argType
+	} else {
```
stephen: i wish this didn't need to introduce a new level indent. could you split this code out into a helper function and use an early return?

graphql/schemabuilder/reflect.go
```diff
@@ -240,7 +240,17 @@ func (sb *schemaBuilder) getEnumArgParser(typ reflect.Type) (*argParser, graphql
...
 }
 
+// cachedType is a container for GraphQL datatype and the list of its fields
+type cachedType struct {
+	argType *graphql.InputObject
+	fields  map[string]argField
+}
+
+// typeCache maps Go types to GraphQL datatypes
+var typeCache map[reflect.Type]cachedType
```
stephen: can we keep this cache on the schemabuilder instead of sharing one globally?

graphql/schemabuilder/reflect.go
stephen: could you also document in the commit message what the previous behavior used to be?

graphql/schemabuilder/reflect.go
```diff
@@ -107,6 +107,56 @@ var scalarArgParsers = map[reflect.Type]*argParser{
 			return nil
 		},
 	},
+	reflect.TypeOf(int8(0)): {
```
stephen:
```
--- FAIL: TestArgParser (0.00s)
	reflect_test.go:571: expected unsupported fields to fail
```

it looks like we had a test to verify that these were not supported, so that'll need a fixup. 😅

graphql/schemabuilder/reflect.go
vavrusa: Hey 👋, sure, I'll work on it some more next week-ish.

graphql/schemabuilder/reflect.go
vavrusa: updated

graphql/schemabuilder/reflect.go
vavrusa: refactored

vavrusa:
@stephen could you take a look again?

graphql/schemabuilder/reflect.go
```diff
@@ -350,6 +355,12 @@ func (sb *schemaBuilder) makeStructParser(typ reflect.Type) (*argParser, graphql
 		return nil, nil, fmt.Errorf("expected struct but received type %s", typ.Name())
 	}
 
+	// Cache type information ahead of time to catch self-reference
+	if sb.typeCache == nil {
+		sb.typeCache = make(map[reflect.Type]cachedType, 0)
+	}
```
stephen: why do we need to check this here if it's already initialized by https://github.com/samsarahq/thunder/pull/120/files#diff-14d89fd44582e138b4a541885877e741R1155 ?

stephen:
@vavrusa - thanks. this looks good but i had one last question.

Dorokhov:
hi @stephen can it be merged now? Thanks.

PR #122, opened 2018-07-12
sqlgen: support types that satisfy Scannable natively

berfarah:
Don't require registration for scalars that support the Scannable interface.

I'll add tests here but wanted to get your opinion on it first

### DIFF
--- a/sqlgen/reflect.go
+++ b/sqlgen/reflect.go
@@ -199,6 +199,21 @@ type Table struct {
 	Scannables *sync.Pool
 }
 
+var scanIface = reflect.TypeOf((*Scannable)(nil)).Elem()
+
+func coerceToScannable(scalarType reflect.Type) (func() Scannable, bool) {
+	if reflect.PtrTo(scalarType).Implements(scanIface) {
+		return func() Scannable {
+			return reflect.New(scalarType).Interface().(Scannable)
+		}, true
+	} else if scalarType.Implements(scanIface) {
+		return func() Scannable {
+			return reflect.New(scalarType).Elem().Interface().(Scannable)
+		}, true
+	}
+	return nil, false
+}
+
 func (s *Schema) buildDescriptor(table string, primaryKeyType PrimaryKeyType, typ reflect.Type) (*Table, error) {
 	if typ.Kind() != reflect.Struct {
 		return nil, fmt.Errorf("bad type %s: not a struct", typ)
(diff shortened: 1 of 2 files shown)


sqlgen/reflect.go
```diff
@@ -243,15 +243,14 @@ func (s *Schema) buildDescriptor(table string, primaryKeyType PrimaryKeyType, ty
...
-			scannable, _ = s.scalarTypes[field.Type.Elem()]
-		} else {
-			scannable, _ = s.scalarTypes[field.Type]
+			scalarType = field.Type.Elem()
 		}
 
-		if scannable == nil {
+		scannable, ok := s.scalarTypes[scalarType]
+		scaninterface := reflect.TypeOf((*Scannable)(nil)).Elem()
+		if !ok && !scalarType.Implements(scaninterface) && !reflect.PtrTo(scaninterface).Implements(scaninterface) {
```
berfarah: We have to check both the element and its pointer in case one of those satisfies the interface. This is because `(t *T) Scan()` on non-pointer `T` won't register as implementing the interface (despite being able to call that method on it)

berfarah:
@jellevandenhooff I'm reasonably certain that this doesn't fix existing behavior. When the scalar exists, behavior doesn't change. If it doesn't, we try to coerce it into the scannable interfac.e

sqlgen/reflect.go
```diff
@@ -243,15 +258,17 @@ func (s *Schema) buildDescriptor(table string, primaryKeyType PrimaryKeyType, ty
...
-		} else {
-			scannable, _ = s.scalarTypes[field.Type]
+			scalarType = field.Type.Elem()
 		}
 
-		if scannable == nil {
+		scannable, ok := s.scalarTypes[scalarType]
+		if !ok {
+			scannable, ok = coerceToScannable(scalarType)
+			s.scalarTypes[scalarType] = scannable
```
jellevandenhooff: let's not write to scalarTypes

sqlgen/reflect.go
```diff
@@ -199,6 +199,21 @@ type Table struct {
 	Scannables *sync.Pool
 }
 
+var scaninterface = reflect.TypeOf((*Scannable)(nil)).Elem()
```
jellevandenhooff: scanIface?

sqlgen/mock_test.go
```diff
@@ -0,0 +1,27 @@
+package sqlgen
+
+import (
+	"database/sql/driver"
+	"fmt"
+)
+
+type mockType []byte
```
jellevandenhooff: put this next to the test

sqlgen/reflect.go
berfarah: 👍

sqlgen/mock_test.go
berfarah: 👍

sqlgen/reflect.go
berfarah: 👍

PR #126, opened 2018-07-20
Shahryart/additional pagination key error check

shahryart:
add error checks for key field

### DIFF
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -303,6 +303,27 @@ func (funcCtx *funcContext) consumePaginatedArgs(sb *schemaBuilder, in []reflect
 
 }
 
+func (sb *schemaBuilder) getKeyFieldOnStruct(nodeType reflect.Type) (string, error) {
+
+	nodeObj := sb.objects[nodeType]
+	if nodeObj == nil && nodeType.Kind() == reflect.Ptr {
+		nodeObj = sb.objects[nodeType.Elem()]
+	}
+	nodeKey := reverseGraphqlFieldName(nodeObj.key)
+	if nodeKey == "" {
+		return nodeKey, fmt.Errorf("a key field must be registered for paginated objects")
+	}
+	if nodeType.Kind() == reflect.Ptr {
+		nodeType = nodeType.Elem()
+	}
+	if _, ok := nodeType.FieldByName(nodeKey); !ok {
+		return nodeKey, fmt.Errorf("field doesn't exist on struct")
+	}
+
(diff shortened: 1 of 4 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -303,6 +303,27 @@ func (funcCtx *funcContext) consumePaginatedArgs(sb *schemaBuilder, in []reflect
...
+	}
+	nodeKey := reverseGraphql(nodeObj.key)
+	if nodeKey == "" {
+		return nodeKey, fmt.Errorf("a key field must be registered for paginated objects")
+	}
+	if nodeType.Kind() == reflect.Ptr {
+		nodeType = nodeType.Elem()
+	}
+	if _, ok := nodeType.FieldByName(nodeKey); !ok {
+		return nodeKey, fmt.Errorf("field doesn't exist on struct")
```
shahryart: this check seems redundant? reflect will fail before this

stephen: is there a test verifying that behavior?

graphql/schemabuilder/reflect.go
```diff
@@ -56,6 +56,18 @@ func makeGraphql(s string) string {
 	return b.String()
 }
 
+func reverseGraphql(s string) string {
```
stephen: this feels too vague as-is (as is `makeGraphql`..)

`reverseGraphqlFieldName`? `graphqlToGoName`?

graphql/schemabuilder/types.go
```diff
@@ -61,6 +61,16 @@ func (s *Object) FieldFunc(name string, f interface{}, options ...FieldFuncOptio
 	s.Methods[name] = m
 }
 
+// Key registers the key field on an object. The field should be specified with the initial
```
stephen: how about "should be specified by the name of the graphql field, i.e. camelCased for struct fields"

graphql/schemabuilder/types.go
```diff
@@ -61,6 +61,16 @@ func (s *Object) FieldFunc(name string, f interface{}, options ...FieldFuncOptio
...
+// lowercased.
+// For example, for an object User:
+// type struct User {
+//	 Id int64
+// }
+// The object is registered as:
+// object := schema.Object("user", User{})
```
stephen: maybe remove lines 70/71

graphql/schemabuilder/pagination.go
shahryart: can't be tested with the current reflect check but nice to leave it in

PR #129, opened 2018-07-21
graphql: remove unused complexity code

stephen:
I don't see places where this is used in the package.


oscartbeaumont:
Could this code be reintroduced and exposed so it can be used as a security measure?

PR #131, opened 2018-07-23
Shahryart/pagination type ptr correction

shahryart:
add ptr type name conversion

### DIFF
--- a/graphql/introspection/test-schema.json
+++ b/graphql/introspection/test-schema.json
@@ -44,6 +44,116 @@
             "deprecationReason": "",
             "description": "",
             "isDeprecated": false,
+            "name": "edges",
+            "type": {
+              "kind": "NON_NULL",
+              "name": "",
+              "ofType": {
+                "kind": "LIST",
+                "name": "",
+                "ofType": {
+                  "kind": "NON_NULL",
+                  "name": "",
+                  "ofType": {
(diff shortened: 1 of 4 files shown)


graphql/introspection/test-schema.json
```diff
@@ -239,6 +239,63 @@
...
+            "deprecationReason": "",
+            "description": "",
+            "isDeprecated": false,
+            "name": "usersConnectionPtr",
+            "type": {
+              "kind": "NON_NULL",
+              "name": "",
+              "ofType": {
+                "kind": "OBJECT",
+                "name": "UserConnection",
```
stephen: does this work as expected? I think the ptr type would have to be different from the non-pointer one, since the type of `node` should be different. (one might have nil elements).

graphql/schemabuilder/pagination.go
```diff
@@ -43,10 +42,16 @@ type ConnectionArgs struct {
 	Args   interface{}
 }
 
+func getTypeName(typ reflect.Type) string {
+	if typ.Kind() == reflect.Ptr {
+		return typ.Elem().Name()
+	}
+	return fmt.Sprintf("nonNull%s", typ.Name())
```
stephen: how about `NonNull` instead of `nonNull`

PR #136, opened 2018-07-26
sqlgen: support custom types

berfarah:
Our past support of custom types was suspect at best. I've added a test for a not otherwise supported type ([16]byte).

**Not attempted**

- Testing binlog (we shouldn't have trouble getting anything out, though, given that it uses BuildStruct under the hood)
- GraphQL (these changes are local to SQL only)

**Caveats**

It seems like we need to implement `Value()` on the type without indirect as we can pass non-pointers into SQL statements and the Valuer interface would otherwise not be satisfied.

### DIFF
--- a/sqlgen/integration_test.go
+++ b/sqlgen/integration_test.go
@@ -79,21 +79,21 @@ func TestContextDeadlineEnforced(t *testing.T) {
 	}
 }
 
-type mockType []byte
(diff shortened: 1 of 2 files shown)


berfarah:
Closed in favor of https://github.com/samsarahq/thunder/pull/142

PR #137, opened 2018-08-02
Shahryart/change pages to end cursors

shahryart:
Use end cursors instead of start cursors for pagination.

This change comes with a reconsideration of the way we are supposed to use cursor based pagination with page numbers. If we use start cursors for each page then it becomes difficult to specify a query for a certain page (e.g for page i  last: itemsCount, before: pages[i+1]) especially if we are requesting the last page since we don't know the number of items in that case (might be lesser than itemsCount). With this change, we can simply request first:itemsCount, after:pages[i] and it would work for the last page.


graphql/schemabuilder/pagination.go
```diff
@@ -256,7 +259,7 @@ func getConnection(key string, nodes []interface{}, args ConnectionArgs) (Connec
 		}
 		keyString := []byte(fmt.Sprintf("%v", keyValue.FieldByName(key).Interface()))
 		cursorVal := base64.StdEncoding.EncodeToString(keyString)
-		if lim != 0 && (int64(i)%lim) == 0 {
+		if lim != 0 && i != len(nodes)-1 && (int64(i+1)%lim) == 0 {
```
stephen: could you add a comment for what this logic change is?

graphql/schemabuilder/pagination.go
```diff
@@ -256,7 +259,9 @@ func getConnection(key string, nodes []interface{}, args ConnectionArgs) (Connec
 		}
 		keyString := []byte(fmt.Sprintf("%v", keyValue.FieldByName(key).Interface()))
 		cursorVal := base64.StdEncoding.EncodeToString(keyString)
-		if lim != 0 && (int64(i)%lim) == 0 {
+		// If the next cursor is the start cursor of a page then push the current cursor to the
+		// list. If an end cursor is the last cursor, then it cannot be followed by a page
```
stephen: nit: period at the end of this sentence

PR #139, opened 2018-08-09
Change pagination args

shahryart:
This pr adds optional embedded pagination args if you want to pass them down into the resolver. This will be useful if we want to paginate all the way down to the db layer. The way this would work is:
`
object.PaginateFieldFunc("paginatedField", func (args struct {graphql.PaginationArgs; OtherArg string}) {
....
})
`


graphql/schemabuilder/pagination.go
```diff
@@ -42,6 +42,14 @@ type ConnectionArgs struct {
 	Args   interface{}
 }
 
+// PaginationArgs are embedded in a struct
```
stephen: why can't we expose `ConnectionArgs` directly?

graphql/schemabuilder/pagination.go
```diff
@@ -298,23 +306,41 @@ func (o *Object) PaginateFieldFunc(name string, f interface{}) {
...
+			argParser, argType, err = sb.buildPaginatedArgParser(in[0])
 		}
 		in = in[1:]
 	} else {
-		if argParser, argType, err = sb.buildPaginatedArgParser(nil); err != nil {
-			return nil, nil, in, err
-		}
+		argParser, argType, err = sb.buildPaginatedArgParser(nil)
+	}
+	if err != nil {
```
stephen: this is kind of weird. can we keep the err handling per branch, or is there a reason to change this structuring?

graphql/schemabuilder/pagination.go
```diff
@@ -298,23 +306,41 @@ func (o *Object) PaginateFieldFunc(name string, f interface{}) {
...
 	var err error
+	embedsArgs := false
+
+	// If the args passed into paginated field func embed the PaginationArgs then the arg parser
+	// needs to be constructed differently from the default case.
 	if len(in) > 0 && in[0] != selectionSetType {
-		if argParser, argType, err = sb.buildPaginatedArgParser(in[0]); err != nil {
-			return nil, nil, nil, err
+		if embeddedPaginationArgs(in[0]) {
+			argParser, argType, err = sb.buildEmbeddedArgParser(in[0])
```
stephen: how about `buildEmbeddedPaginatedArgParser`

graphql/schemabuilder/pagination.go
```diff
@@ -420,11 +448,36 @@ func (sb *schemaBuilder) buildPaginatedField(typ reflect.Type, f interface{}) (*
...
-func (funcCtx *funcContext) extractPaginatedRetAndErr(nodeKey string, out []reflect.Value, args interface{}, retType graphql.Type) (interface{}, error) {
+func getEmbedIndex(typ reflect.Type) int {
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
+		if field.Type == reflect.TypeOf(PaginationArgs{}) {
+			return i
+		}
+	}
+	return 0
+}
```
stephen: consider inlining this into the function?

graphql/schemabuilder/pagination.go
```diff
@@ -420,11 +448,36 @@ func (sb *schemaBuilder) buildPaginatedField(typ reflect.Type, f interface{}) (*
...
 
-func (funcCtx *funcContext) extractPaginatedRetAndErr(nodeKey string, out []reflect.Value, args interface{}, retType graphql.Type) (interface{}, error) {
+func getEmbedIndex(typ reflect.Type) int {
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
+		if field.Type == reflect.TypeOf(PaginationArgs{}) {
+			return i
+		}
+	}
+	return 0
```
stephen: this should be -1 or some error case i think

graphql/schemabuilder/pagination.go
```diff
@@ -420,11 +448,36 @@ func (sb *schemaBuilder) buildPaginatedField(typ reflect.Type, f interface{}) (*
...
+			After:  connectionArgs.After,
+			Before: connectionArgs.Before,
+		}
+	} else {
+		fieldInd := getEmbedIndex(reflect.TypeOf(args))
+		paginationArgs, _ = reflect.ValueOf(args).Field(fieldInd).Interface().(PaginationArgs)
+	}
 
-	result, err := getConnection(nodeKey, castSlice(out[0].Interface()), connectionArgs)
+	result, err := getConnection(nodeKey, castSlice(out[0].Interface()), paginationArgs)
```
stephen: interesting. i think if we are passing down the pagination args, we should assume that the underlying resolver has handled passing back the exact correct results

graphql/connection_test.go
```diff
@@ -0,0 +1,628 @@
+package graphql_test
```
stephen: could you separate this commit into two? one that moves existing tests, then one that adds your new tests.

graphql/schemabuilder/pagination.go
```diff
@@ -298,23 +306,41 @@ func (o *Object) PaginateFieldFunc(name string, f interface{}) {
 		})
 }
 
-func (funcCtx *funcContext) consumePaginatedArgs(sb *schemaBuilder, in []reflect.Type) (*argParser, graphql.Type, []reflect.Type, error) {
+func embeddedPaginationArgs(argType reflect.Type) bool {
```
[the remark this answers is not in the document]
shahryart: done

graphql/schemabuilder/pagination.go
shahryart: Connection args and pagination args are used differently. Connection args has an interface that is used to parse the user facing args whereas Pagination args are user facing (passed into the resolver).

graphql/schemabuilder/pagination.go
shahryart: done

graphql/schemabuilder/pagination.go
shahryart: done

graphql/schemabuilder/pagination.go
shahryart: done

graphql/schemabuilder/pagination.go
shahryart: oh okay i thought it would be better since all branches return an error

shahryart:
I think if we are passing pagination args down we cannot support listing the entire set of pages, however, I feel that having a set of 10 pages is normal across most interfaces hence we should except the DB request to return a list of 5 pages before or after the requested cursor. Beyond a certain number of pages, however, it gets expensive to make the DB request as well.

graphql/connection_test.go
shahryart: done

graphql/schemabuilder/pagination.go
```diff
@@ -452,6 +557,96 @@ func castSlice(slice interface{}) []interface{} {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	pagObj, ok := pagArgType.(*graphql.InputObject)
+	if !ok {
+		panic("failed to cast paginated args to an input object")
+	}
+	for name, objField := range pagObj.InputFields {
+		if _, ok := argType.InputFields[name]; ok {
+			return nil, nil, fmt.Errorf("These arg names are restricted: First, After, Last and Before")
```
stephen: nit: start error with lowercase

graphql/connection_test.go
```diff
@@ -517,3 +517,112 @@ func TestPaginateNodeTypeFailure(t *testing.T) {
...
+		t.Error(err)
+	}
+	e := graphql.Executor{}
+	val, err := e.Execute(context.Background(), builtSchema.Query, nil, q)
+	assert.Nil(t, err)
+
+	assert.Equal(t, map[string]interface{}{
+		"inner": map[string]interface{}{
+			"innerConnection": map[string]interface{}{
+				"totalCount": int64(5),
```
stephen: as we talked about, i think `totalCount` is not quite right, since there might be more results that what is just passed through. What does it look like for us to support a custom resolver for `totalCount`?

shahryart: I'm leaning towards returning an integer in the resolver.

graphql/schemabuilder/pagination.go
```diff
@@ -281,9 +299,19 @@ func getConnection(key string, nodes []interface{}, args ConnectionArgs) (Connec
...
 
+	if returnsConnInfo {
+		connInfo := out[1].Interface().(PaginationInfo)
+		pageInfo := PageInfo{
+			HasNextPage: connInfo.HasNextPage,
+			HasPrevPage: connInfo.HasPrevPage,
+			StartCursor: startCursor,
+			EndCursor:   endCursor,
+		}
+		return Connection{TotalCount: connInfo.TotalCount(), Edges: edges, PageInfo: pageInfo}, nil
```
stephen: think we should invoke `TotalCount` only if the client requests it. i.e. it should be registered as a field func on `PageInfo`, rather than always invoking it.

graphql/schemabuilder/pagination.go
```diff
@@ -281,9 +299,19 @@ func getConnection(key string, nodes []interface{}, args ConnectionArgs) (Connec
...
+	}
 	pageInfo := PageInfo{HasNextPage: nextPage, EndCursor: endCursor, StartCursor: startCursor, HasPrevPage: prevPage, Pages: pages}
```
stephen: can we block out `pages` in the passthrough case?

graphql/schemabuilder/pagination.go
shahryart: you mean registered as a field func on connection?

graphql/schemabuilder/pagination.go
shahryart: Yes

graphql/schemabuilder/pagination.go
```diff
@@ -341,6 +393,43 @@ func (sb *schemaBuilder) getKeyFieldOnStruct(nodeType reflect.Type) (string, err
...
+	if len(out) > 0 && out[0] == reflect.TypeOf(PaginationInfo{}) {
+		retConnInfo = true
+		out = out[1:]
+	}
+
+	if len(out) > 0 && out[0] == errType {
+		funcCtx.hasError = true
+		out = out[1:]
+	}
+	log.Println(len(out))
```
bojdell: did you mean to log here?

graphql/schemabuilder/pagination.go
```diff
@@ -341,6 +393,43 @@ func (sb *schemaBuilder) getKeyFieldOnStruct(nodeType reflect.Type) (string, err
...
+	if len(out) != 0 {
+		err = fmt.Errorf("%s return values should [result][, error]", funcCtx.funcType)
```
bojdell: does this error message make sense?

graphql/schemabuilder/pagination.go
```diff
@@ -42,6 +43,21 @@ type ConnectionArgs struct {
...
 
+// PaginationArgs are embedded in a struct
+type PaginationArgs struct {
+	First  *int64
+	Last   *int64
+	After  *string
+	Before *string
+}
+
+//PaginationInfo can be returned in a PaginateFieldFunc
```
stephen: nit: space at beginning, period at the end

stephen: consider adding a comment similar to https://github.com/graphql/graphql-relay-js/issues/94#issuecomment-232410564 about how to populate this struct.

graphql/schemabuilder/pagination.go
```diff
@@ -135,6 +150,11 @@ func (funcCtx *funcContext) constructConnType(sb *schemaBuilder, typ reflect.Typ
 
 	pageInfoType, _ := reflect.TypeOf(Connection{}).FieldByName("PageInfo")
 	pageInfoField, err := sb.buildField(pageInfoType)
+	pageInfoNonNull, _ := pageInfoField.Type.(*graphql.NonNull)
+	pageInfoObj := pageInfoNonNull.Type.(*graphql.Object)
+	if returnsConnInfo {
```
stephen: can you add a comment for why decided on this?

graphql/schemabuilder/pagination.go
```diff
@@ -341,6 +400,42 @@ func (sb *schemaBuilder) getKeyFieldOnStruct(nodeType reflect.Type) (string, err
...
+		out = append(out, funcCtx.funcType.Out(i))
+	}
+
+	if len(out) > 0 && out[0] != errType {
+		funcCtx.hasRet = true
+		out = out[1:]
+	}
+
+	if len(out) > 0 && out[0] == reflect.TypeOf(PaginationInfo{}) {
+		retConnInfo = true
```
stephen: should we error if the resolver takes pagination args but doesn't return PaginationInfo?

stephen: also nit: can we keep this variable name consistent between `ConnInfo` and `PaginationInfo`?

graphql/schemabuilder/pagination.go
```diff
@@ -420,15 +518,41 @@ func (sb *schemaBuilder) buildPaginatedField(typ reflect.Type, f interface{}) (*
...
+		connectionArgs, _ := args.(ConnectionArgs)
+		paginationArgs = PaginationArgs{
+			First:  connectionArgs.First,
+			Last:   connectionArgs.Last,
+			After:  connectionArgs.After,
+			Before: connectionArgs.Before,
+		}
+	} else {
+		pagTyp := reflect.TypeOf(args)
+		fieldInd := -1 // will panic
```
stephen: nit: either remove the comment or expand to a sentence?

graphql/schemabuilder/pagination.go
```diff
@@ -420,15 +518,41 @@ func (sb *schemaBuilder) buildPaginatedField(typ reflect.Type, f interface{}) (*
...
+		for i := 0; i < pagTyp.NumField(); i++ {
+			field := pagTyp.Field(i)
+			if field.Type == reflect.TypeOf(PaginationArgs{}) {
+				fieldInd = i
+			}
+		}
+
+		paginationArgs, _ = reflect.ValueOf(args).Field(fieldInd).Interface().(PaginationArgs)
```
stephen: second return value is only necessary if you're going to use it

graphql/schemabuilder/pagination.go
stephen: ping on this

graphql/schemabuilder/pagination.go
```diff
@@ -298,23 +333,47 @@ func (o *Object) PaginateFieldFunc(name string, f interface{}) {
 		})
```
stephen: can we update the comment for `PaginateFieldFunc` to talk about the pagination modes available after this PR?

graphql/schemabuilder/pagination.go
stephen: oh, i see this is done in the third commit. it'd be nice to squash that down to here.

graphql/schemabuilder/pagination.go
shahryart: done

graphql/schemabuilder/pagination.go
shahryart: done

graphql/schemabuilder/pagination.go
shahryart: gotcha

graphql/schemabuilder/pagination.go
shahryart: done

PR #142, opened 2018-08-14
[Major Change] Introduce `types` package for type handling in sqlgen

berfarah:
## Goals

1. Support protobuf struct (de)serialization
2. Support custom scalars. For real. Like, types that aren't natively supported by SQL.
3. Keep supporting scalar aliases (string -> custom string etc).

 ## Changes

This PR adds a types package with the intention of replacing Scannables / TypeConverters. This unifies all of our logic in one place for determining how values are converted, while also supporting native interfaces.

 ### Implementation

1. Replace scannables with fields. Ensure that `Empty`/`Interface` is being called when scanning values.
2. Add integration tests for both native and custom types.
3. Deprecate `MustRegisterCustomScalar` and `MustRegisterSimpleScalar`.
4. ???
5. Profit

 ### Usage

```go
type Model struct {
  Foo ProtoModel `sql:",binary"` // (de)serializes to a binary blob via `Marshal`/`MarshalBinary`
  IP IP `sql:",primary,string"` // (de)serializes into a string via `MarshalText`
  UUID UUID // serializes via `sql/driver.Valuer` and deserializes via `sql.Scanner`
  Bar CustomString // (de)serializes the same way a native string would
  // int8, int16, int32 are converted to int64
  // uint8, uint16, uint32, uint64 are converted to int64
  // float32 is converted to float64
  // bool, float64, int64, string, []byte, time.Time and nil are natively supported
}
```

(but really, after this is implemented for SQL, look at where we could use it on the GraphQL side as well)


testfixtures/custom_type.go
```diff
@@ -5,25 +5,28 @@ import (
 	"fmt"
 )
 
-type CustomType []byte
+type CustomType [16]byte
```
stephen: could you explain this fixture change?

types/field.go
```diff
@@ -0,0 +1,88 @@
...
+type Field struct {
+	Tags  TagSet
+	value reflect.Value
+	Type  reflect.Type
+	Kind  reflect.Kind
+	Ptr   bool
+	IsNil bool
+}
+
+func FieldFromValue(in interface{}, tags []string) *Field {
```
stephen: ok, having read the whole PR now, I think I'd like to see this in `_test.go` only - it doesn't seem like there is a reason to use this outside of test.

berfarah:
Discounting tests, comments and test data, this only adds ~200 LOC. 😬

types/field.go
```diff
@@ -0,0 +1,72 @@
+package types
+
+import (
+	"reflect"
+)
+
+type TagSet map[string]struct{}
```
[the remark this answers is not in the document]
berfarah: Done

types/sql.go
```diff
@@ -74,3 +77,131 @@ func (f Field) Value() (driver.Value, error) {
...
+
+	// Null values are simply set to zero.
+	if src == nil {
+		return nil
+	}
+
+	// Handle native types.
+	switch i.(type) {
+	case *[]byte:
+		fmt.Println("bytes")
```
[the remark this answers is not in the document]
berfarah: Done

types/sql.go
```diff
@@ -0,0 +1,76 @@
...
+//    nil - for NULL values
+func (f Field) Value() (driver.Value, error) {
+	i := f.value.Interface()
+
+	if valuer, ok := i.(driver.Valuer); ok {
+		return valuer.Value()
+	}
+
+	// Handle sql type overrides as special cases.
+	switch f.TagType["sql"] {
```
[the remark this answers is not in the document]
berfarah: I've changed this

types/sql.go
```diff
@@ -0,0 +1,76 @@
...
+		if b, ok := i.([]byte); ok {
+			return b, nil
+		}
+		if iface, ok := i.(marshaler); ok {
+			return iface.Marshal()
+		}
+		if iface, ok := i.(encoding.BinaryMarshaler); ok {
+			return iface.MarshalBinary()
+		}
+	case "string", "text":
```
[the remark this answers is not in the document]
berfarah: The idea here is that you might have varchar or a text blob - though I'd be fine with changing it to just string

types/field.go
```diff
@@ -0,0 +1,50 @@
...
+
+	it.Kind = t.Kind()
+
+	return it
+}
+
+// Blank creates a blank copy of a field. This _must_ be used in order to create a
+// sql.Scanner from a field.
+// Example:
+// f := FieldFromValue(map[string]string{"sql": "binary", i)
```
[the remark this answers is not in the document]
berfarah: This won't be needed any more

types/field.go
```diff
@@ -0,0 +1,72 @@
...
+// scanner.Interface() // get back interface{}
+func (f *Field) Blank() *Field {
+	field := &Field{}
+	*field = *f
+	field.value = reflect.ValueOf(nil)
+	return field
+}
+
+// Interface returns the value as an interface.
+func (f Field) Interface() interface{} { return f.value.Interface() }
```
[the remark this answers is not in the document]
berfarah: I've tried to make intent clear where possible - only have pointers for methods that are meant to mutate.

types/sql.go
```diff
@@ -0,0 +1,93 @@
...
+func (f Field) Value() (driver.Value, error) {
+	// Our value should never be a pointer. driver.Value should always be implemented as a
+	// non-pointer method.
+	i := f.value.Interface()
+
+	// If our interface supports driver.Valuer we can immediately short-circuit as this is what the
+	// MySQL driver would do.
+	if valuer, ok := i.(driver.Valuer); ok {
+		return valuer.Value()
+	}
```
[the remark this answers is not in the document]
berfarah: Your second comment is correct - I'm not sure where/how I would want to document that. Do you have thoughts on where it would make the most sense?

types/sql.go
```diff
@@ -0,0 +1,93 @@
...
+		if b, ok := i.([]byte); ok {
+			return b, nil
+		}
+		if iface, ok := i.(marshaler); ok {
+			return iface.Marshal()
+		}
+		if iface, ok := i.(encoding.BinaryMarshaler); ok {
+			return iface.MarshalBinary()
+		}
+	case f.Tags.Contains("string"), f.Tags.Contains("text"):
```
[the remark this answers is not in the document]
berfarah: We can just use `string`

types/sql.go
```diff
@@ -0,0 +1,93 @@
...
+		if b, ok := i.([]byte); ok {
+			return b, nil
+		}
+		if s, ok := i.(string); ok {
+			return s, nil
+		}
```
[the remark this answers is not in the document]
berfarah: Good point. I can remove this.

types/field.go
```diff
@@ -68,5 +68,21 @@ func (f *Field) Blank() *Field {
 	return field
 }
 
+// CopyTo copies the reference onto the given value. Note: this uses pointers, so maybe this is a
+// bad name?
+func (f *Field) CopyTo(val reflect.Value) {
+	if f.Ptr {
+		if f.IsNil {
+			// val.Set(reflect.ValueOf(nil))
```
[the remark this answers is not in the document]
berfarah: I've changed that in the new code. I'm still a little bit wary with no value being set, but I'm not 100% sure of how to set nils.

sqlgen/deprecated.go
```diff
@@ -3,6 +3,8 @@ package sqlgen
 import (
 	"database/sql/driver"
 	"fmt"
+
+	_ "github.com/go-sql-driver/mysql"
```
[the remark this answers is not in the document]
berfarah: I'm going to move this into tests.

sqlgen/reflect.go
```diff
@@ -246,85 +223,25 @@ func (s *Schema) buildDescriptor(table string, primaryKeyType PrimaryKeyType, ty
...
-	return nil
-}
-
-func (s *Schema) MustRegisterCustomScalar(scalar interface{}, makeScannable func() Scannable) {
-	if err := s.RegisterCustomScalar(scalar, makeScannable); err != nil {
-		panic(err)
-	}
-}
-
-func (s *Schema) RegisterSimpleScalar(scalar interface{}) error {
```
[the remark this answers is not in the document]
berfarah: The migration _should_ be as simple as deleting these calls everywhere. My strategy for migrating this would be deleting all the calls and checking to see if it works.

sqlgen/reflect.go
```diff
@@ -51,40 +49,43 @@ const (
...
-// BuildStruct constructs a struct value defined by table and based on scannables
-func BuildStruct(table *Table, scannables []interface{}) interface{} {
+// BuildStruct constructs a struct value defined by table and field values.
+func BuildStruct(table *Table, fields []*types.Field) interface{} {
 	ptr := reflect.New(table.Type)
 	elem := ptr.Elem()
 
 	for i, column := range table.Columns {
-		value, _ := scannables[i].(driver.Valuer).Value()
+		// These values are all non-pointers and have been copied.
```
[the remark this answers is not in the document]
berfarah: Added a comment here

types/field.go
berfarah: I've removed this entirely

testfixtures/custom_type.go
berfarah: Yep: I wanted this to no longer be a natively supported SQL type

fields/sql.go
```diff
@@ -0,0 +1,106 @@
...
+//    float64
+//    bool
+//    []byte
+//    string
+//    time.Time
+//    nil - for NULL values
+func (f Valuer) Value() (driver.Value, error) {
+	// Ideally, all the interfaces below would be implemented as normal methods, not indirect
+	// methods. However, in order to be safe we take the address of pointers where we can.
+	// This specifically supports protobuf.
```
stephen: is this comment referring to the next line? i am having trouble understanding what this implies for my read here

fields/sql.go
```diff
@@ -0,0 +1,106 @@
...
+	// struct {
+	//   Blob proto.Blob `sql:",binary"`       // ensures that Marshal or MarshalBinary is used.
+	//   IP IP `sql:",string"`                 // ensures that its MarshalText method
+	//	                                       // is used for serialization.
+	//   IP IP `sql:",text"`                   // is equivalent to the former field
+	//   JSON map[string]string `sql:",json"`  // ensures that json.Marshal is used on the value.
+	// }
+	switch {
+	case f.Tags.Contains("binary"):
+		if iface, ok := i.(marshaler); ok {
```
stephen: could you help me understand the difference between marshaler and binary marshaler? why would one implement one vs. the other?

fields/sql.go
```diff
@@ -0,0 +1,106 @@
...
+			return iface.MarshalBinary()
+		}
+	case f.Tags.Contains("string"), f.Tags.Contains("text"):
+		if iface, ok := i.(encoding.TextMarshaler); ok {
+			return iface.MarshalText()
+		}
+	case f.Tags.Contains("json"):
+		if b, ok := i.([]byte); ok {
+			return b, nil
+		}
```
stephen: why would one use the `json` tag if the value is a string? i.e. there is nothing for the marshaler to do

fields/interface.go
```diff
@@ -0,0 +1,9 @@
+package fields
+
+type marshaler interface {
```
stephen: could you point to some places in the standard library where this is used but not exported?

fields/sql.go
```diff
@@ -104,3 +107,157 @@ func (f Valuer) Value() (driver.Value, error) {
...
+		if _, ok := src.([]byte); ok {
+			s.value.Set(reflect.ValueOf(src))
+			return nil
+		}
+	case *time.Time:
+		if _, ok := src.(time.Time); ok {
+			s.value.Set(reflect.ValueOf(src))
+			return nil
+		}
+	}
```
stephen: could you comment on why these need to be handled in this codepath, but not in the valuer path?

example/main.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"sort"
 
+	_ "github.com/go-sql-driver/mysql"
```
stephen: what is this for?

stephen: there is a lot going on in this last commit. it'd be great to split it up into a few different changes - maybe by valuer/scanner/livesql.

fields/sql.go
berfarah: I've attempted to clarify this by the interface. This specifically supports protobuf and other libraries that implement Marshal() (I know I've seen it before but I can't remember one off the top of my head)

fields/sql.go
```diff
@@ -0,0 +1,106 @@
...
+	// }
+	switch {
+	case f.Tags.Contains("binary"):
+		if iface, ok := i.(marshaler); ok {
+			return iface.Marshal()
+		}
+		if iface, ok := i.(encoding.BinaryMarshaler); ok {
+			return iface.MarshalBinary()
+		}
+	case f.Tags.Contains("string"), f.Tags.Contains("text"):
```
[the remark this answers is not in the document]
berfarah: Done

fields/sql.go
berfarah: Removed

fields/interface.go
berfarah: There are similar but not exactly the same patterns in the stdlib. We looked at https://github.com/gogo/protobuf/issues/28 together. I'm still not clear on why it wasn't implemented - the issue was fairly confusing and from over 3 years ago.

example/main.go
berfarah: This is in order to "install" the mysql driver.

berfarah:
@stephen Hopefully these commits are a little cleaner

sqlgen/reflect.go
```diff
@@ -319,7 +332,7 @@ func makeWhere(table *Table, filter Filter) (*SimpleWhere, error) {
 			return nil, fmt.Errorf("unknown column %s", name)
 		}
 
-		l = append(l, whereElem{column: column, value: value})
+		l = append(l, whereElem{column: column, value: column.Descriptor.Valuer(reflect.ValueOf(value))})
```
stephen: to verify - this is a departure from the existing behavior because the filters values are now converted vs. used directly.

sqlgen/reflect_test.go
```diff
@@ -226,37 +212,18 @@ func TestMakeWhere(t *testing.T) {
...
-	if err != nil {
-		t.Error(err)
-	}
-	if !reflect.DeepEqual(where, &SimpleWhere{
-		Columns: []string{"id"},
-		Values:  []interface{}{10},
-	}) {
-		t.Error("bad select")
-	}
+	assert.NoError(t, err)
```
stephen: did the tests from here down in this file change due to api changes, or as cleanup? if the latter, it'd be nice to have them separate/noted that way for reviewing.

fields/sql.go
```diff
@@ -0,0 +1,99 @@
...
+			return iface.Marshal()
+		}
+		if iface, ok := i.(encoding.BinaryMarshaler); ok {
+			return iface.MarshalBinary()
+		}
+	case f.Tags.Contains("string"):
+		if iface, ok := i.(encoding.TextMarshaler); ok {
+			return iface.MarshalText()
+		}
+	case f.Tags.Contains("json"):
```
stephen: mentioned irl, but i think it would have been nice to see string/json support as a follow up vs. in this initial implementation.

fields/sql.go
```diff
@@ -97,3 +100,157 @@ func (f Valuer) Value() (driver.Value, error) {
 
 var _ driver.Valuer = Valuer{}
 var _ driver.Valuer = &Valuer{}
+
+// Scanner fulfills the sql.Scanner interface which deserializes SQL values
+// into the type dictated by our descriptor.
+type Scanner struct {
+	*Descriptor
+	value   reflect.Value
+	IsValid bool
```
stephen: does `IsValid` need to be persisted on the struct? I see that it is only referenced within `Scan`

fields/sql.go
stephen: Thanks, I think that helps.

sqlgen/reflect.go
```diff
@@ -258,17 +203,9 @@ func (s *Schema) buildDescriptor(table string, primaryKeyType PrimaryKeyType, ty
...
-			scalarType = field.Type.Elem()
-		}
-
-		scannable, ok := s.scalarTypes[scalarType]
-		if !ok {
-			scannable, ok = coerceToScannable(scalarType)
-		}
-		if !ok {
-			return nil, fmt.Errorf("bad type %s: field %s has unsupported type %s", typ, field.Name, field.Type)
+		d := fields.New(field.Type, tags[1:])
```
stephen: is it safe for us to reuse the `sql` tag in this manner? should we use a different tag?

sqlgen/reflect.go
berfarah: Yes

fields/sql.go
berfarah: Yep, since we treat the value as a non-pointer value, IsValid tells us whether or not something is nil

sqlgen/reflect.go
berfarah: The sql tag is explicitly being used by only us in sqlgen. I think we should just use `sql` everywhere.

stephen: ah, yep, that makes sense. i had it in my head that other software used that tag too

fields/sql.go
stephen: could you point to where that is used outside of `Scan`?

sqlgen/reflect_test.go
berfarah: 😬 it was admittedly cleanup. I tried to extract these all into a separate PR but some of them got away from me

fields/sql.go
berfarah: Yeah :|

fields/sql.go
stephen: ah, thanks i missed that. making it package private sounds good

berfarah:
Includes https://github.com/samsarahq/thunder/pull/150

Allocations went up:
```
goos: darwin
goarch: amd64
pkg: github.com/samsarahq/thunder/sqlgen
Benchmark/Read-8         	    5000	    291292 ns/op	    2239 B/op	      60 allocs/op
Benchmark/Create-8       	    2000	    683282 ns/op	    1568 B/op	      39 allocs/op
Benchmark/Update-8       	    2000	    576854 ns/op	    1976 B/op	      49 allocs/op
Benchmark/Delete-8       	    2000	    643999 ns/op	    1000 B/op	      28 allocs/op
```

fields/sql.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Value satisfies the sql/driver.Valuer interface.
+// The value should be one of the following:
+//    int64
+//    float64
+//    bool
+//    []byte
+//    string
+//    time.Time
```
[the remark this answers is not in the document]
berfarah: That's where we can use `Scan`/`Value` to convert stuff more effectively

fields/sql.go
```diff
@@ -0,0 +1,99 @@
...
+	// Return early if the value is nil. Ideally we would do a `i == nil` comparison here, but
+	// unfortunately for us, `nil` is typed and that would always return false.
+	if f.value.Kind() == reflect.Ptr && f.value.IsNil() {
+		return nil, nil
+	}
+
+	// If our interface supports driver.Valuer we can immediately short-circuit as this is what the
+	// MySQL driver would do.
+	if valuer, ok := i.(driver.Valuer); ok {
+		return valuer.Value()
```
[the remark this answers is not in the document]
berfarah: For valuer, the nil case will panic (I checked). In scanner we can and do handle it.

fields/sql.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// If we can't figure out what the type is supposed to be, we pass it straight through to SQL,
+	// which will return an error if it can't handle it.
+	// This means we don't have to handle []byte or time.Time specially, since they'll just pass on
+	// through.
+	return f.value.Interface(), nil
+}
+
+var _ driver.Valuer = Valuer{}
+var _ driver.Valuer = &Valuer{}
```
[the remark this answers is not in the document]
berfarah: Nope, but it's nice to know that we are fulfilling the interface and the Go stdlib introduces this pattern. It only runs once so allocations are constant time (and presumably garbage collected)

fields/sql.go
```diff
@@ -97,3 +100,162 @@ func (f Valuer) Value() (driver.Value, error) {
...
+// The src value will be one of the following:
+//    int64
+//    float64
+//    bool
+//    []byte
+//    string
+//    time.Time
+//    nil - for NULL values
+func (s *Scanner) Scan(src interface{}) error {
+	// Get a value of the pointer of the pointer of our type. The Scanner and Unmarshalers should
```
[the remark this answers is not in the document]
berfarah: I think this comment is actually just wrong. `of the pointer` should be there once. Will update.

fields/sql.go
```diff
@@ -97,3 +100,162 @@ func (f Valuer) Value() (driver.Value, error) {
 
 var _ driver.Valuer = Valuer{}
 var _ driver.Valuer = &Valuer{}
+
+// Scanner fulfills the sql.Scanner interface which deserializes SQL values
+// into the type dictated by our descriptor.
+type Scanner struct {
+	*Descriptor
+	value   reflect.Value
+	isValid bool
```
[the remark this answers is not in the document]
berfarah: `isValid` is more following the style of `NullBytes`/`NullBool`/etc where it means "is not nil"

berfarah:
@stephen since this is a pretty big breaking change, I'm thinking we should tag this with a 1.0 release. What do you think?

fields/sql_test.go
```diff
@@ -0,0 +1,217 @@
...
+		// Type aliases:
+		{Out: likeString("foo"), In: "foo"},
+		{Out: int8(5), In: int64(5)},
+		{Out: int16(5), In: int64(5)},
+		{Out: int32(5), In: int64(5)},
+		{Out: likeInt(5), In: int64(5)},
+		{Out: float32(5), In: float64(5)},
+		{Out: likeFloat(5), In: float64(5)},
+		// Interfaces without tags:
+		{Out: ifaceScanner{[]byte("scan me")}, In: []byte("scan me")},
```
changpingc: pointer ifaceScanner test case?

PR #144, opened 2018-08-15
go livesql: add integration test

berfarah:
Add an integration on LiveSQL so we know when we're breaking it.

### DIFF
--- a/livesql/live.go
+++ b/livesql/live.go
@@ -89,7 +89,11 @@ func (t *dbTracker) registerDependency(ctx context.Context, table string, tester
 	t.add(r)
 }
 
-// LiveDB is a SQL client that supports live updating queries
+// LiveDB is a SQL client that supports live updating queries.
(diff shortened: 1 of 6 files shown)


berfarah:
@stephen I think I had originally left out some stuff from the integration test. I've updated this. I'm fine with leaving out the integration test for now and having the example serve as that.

livesql/live.go
```diff
@@ -111,6 +115,7 @@ type queryCacheKey struct {
 
 // query reactively performs a SelectQuery
 func (ldb *LiveDB) query(ctx context.Context, query *sqlgen.BaseSelectQuery) ([]interface{}, error) {
+	// Fall back to sqlgen querying if there is no reactive rerunner present.
```
stephen: or if we're in a tx

berfarah: Updated - also updated integration test

Issue #148, opened 2018-08-20
Errors handling and GraphQL spec


adelcasse:
Hello,

I'm new to this library so please excuse me if I'm out of scope or didn't understand something.

As far as I've understood, GraphQL errors are returned as strings : 

- "name.something: error message" when they are path errors (ie. we gave a name to the query and declared them with errors.New in resolvers)
- "error message" : when we use a sanitized error or if we didn't declare a name for the object.

But the GraphQL specification is quite different : http://facebook.github.io/graphql/June2018/#sec-Errors

Errors messages should look like : 

    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ]
    }

This leads to issues with client libraries expecting the right format.

Tell me if I missed something. If so, I'm sorry.

In case I'm right, is there any interest in fixing that (which would probably be a breaking change for existing users, and Samsara in the first place) ? If so, we could probably contribute to it as we're definitely interested in using this library in our products (everything else is just perfect ;)).

stephen:
Hi @adelcasse, you're correct in that thunder does not currently adhere to the graphql error handling mechanism. Here's some work that @changpingc did that looks similarly: https://github.com/samsara-dev/backend/pull/12845/files

We could also try to run flate (compression) ourselves on that json blob: https://golang.org/pkg/compress/flate/

adelcasse:
Hi @stephen,

Thanks for the quick answer. I don't have access to samsara-dev/backend (probably private organization/repo ?)

adelcasse:
@stephen I've tried something to avoid breaking changes but still be able to return valid GraphQL errors. What do you thing about that and is there interest in pursuing this implementation ?

ccbrown:
This is kind of a big deal imo. I would call spec violations like this a deal-breaker for any API that hopes to see widespread use.

That said, I'm not a big fan of that PR as-is. Setting the location and path shouldn't be the library consumer's responsibility; it should be the library's. Imo there should simply be a flag on the schema that enables spec-compliant errors. And extensions should be facilitated via interface rather than concrete error type.

adelcasse:
I agree. Without any answer here, we finally switched to another library.

sijad:
@stephen can you please share your inputs about this one?

PR #149, opened 2018-08-21
Support GraphQL errors as in GraphQL spec

adelcasse:
Following issue #148, what about something like that that could avoid introducing a breaking change and still handle valid GraphQL errors as in official GraphQL specification ? 

I've implemented it on HTTP only for the moment (as a proof of concept) and it could definitely be improved by doing things similar to path errors ...

If it's something interesting for you, I'll pursue on that direction.

### DIFF
--- a/graphql/http.go
+++ b/graphql/http.go
@@ -29,15 +29,34 @@ type httpPostBody struct {
 }
 
 type httpResponse struct {
-	Data   interface{} `json:"data"`
-	Errors []string    `json:"errors"`
+	Data   interface{}   `json:"data"`
+	Errors []interface{} `json:"errors"`
 }
 
 func (h *httpHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	writeResponse := func(value interface{}, err error) {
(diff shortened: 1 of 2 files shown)


ccbrown:
Setting the location and path shouldn't be the library consumer's responsibility; it should be the library's. Imo there should simply be a flag on the schema that enables spec-compliant errors. And extensions should be facilitated via interface rather than concrete error type.

ctnieves:
Is this going anywhwere? Or has an alternative PR been made/merged? This has been open since Aug 2018.

gobaldia:
@CodeBrew28  I know this has been inactive for a while, but I'd like to know if you're considering to support errors as in GraphQL spec.
Thunder is a great library, but it's a pity to manage errors in the current way.

adelcasse:
I don't think they will answer. In https://github.com/samsarahq/thunder/issues/148 they gave a first answer (which pointed me to a private repository I coudn't read) and then, instead of explaining more, deleted this answer and never got back to me. (They could just have said "no we don't want to change that")

PR #154, opened 2018-08-25
go sqlgen: support nullable (pointer) sql.Scanners

berfarah:
Currently, the only way to make sql.Scanner nullable is either by using the sql.NullBool pattern (where we break up the value into `IsValid` and `Bool`/whatever type).

This is because the single de-reference method can't overwrite the pointer to be a null pointer. We also can't define two methods (one for single de-reference and one for double de-reference), so the developer story starts looking a little messy.

In order to support nullable custom types (for real), we will handle `nil` on behalf of the scanner if it's a pointer.

**Side Effect: Remove `Scanner.Interface()`**

This method was exclusively being used in test at this point and won't properly set `nil` for pointers. In order to prevent bad behavior, we need to remove this method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## [Unreleased]
 
+## [0.3.0] - 2018-08-27
+
+### Changed
+
+#### `sqlgen`
+
+- Pointer scanners are no longer allowed to handle `nil`, are forcefully set to
+  `nil` instead.
+
+### Removed
+
+#### `fields`
+
+- Removed `Scanner.Interface()`: this should never have been exposed and
+  wouldn't work the way you would expect it to. Instead, you can only copy to an
+  existing `reflect.Value`.
+
+
 ## [0.2.0] - 2018-08-23
 
 ### Added
--- a/fields/sql.go
+++ b/fields/sql.go
@@ -117,14 +117,6 @@ type Scanner struct {
 	isValid bool
 }
 
-// Interface returns the value deserialized into the scanner.
-func (s Scanner) Interface() interface{} {
-	if s.value.IsValid() {
-		return s.value.Interface()
-	}
-	return nil
-}
-
 // CopyTo copies the scanner value to another reflect.Value. This is used for setting structs.
 func (s *Scanner) CopyTo(to reflect.Value) {
 	s.copy(s.value, to, s.isValid)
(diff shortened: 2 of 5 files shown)


berfarah:
Looks like I may have missed some `Interface()` removals.

fields/sql.go
```diff
@@ -117,14 +117,6 @@ type Scanner struct {
 	isValid bool
 }
 
-// Interface returns the value deserialized into the scanner.
-func (s Scanner) Interface() interface{} {
```
changpingc: 👍

fields/sql_test.go
```diff
@@ -131,53 +131,69 @@ func TestField_Value(t *testing.T) {
...
+		// Pointer interfaces with tags with nil:
+		{Type: &ifaceMarshal{}, Out: (*ifaceMarshal)(nil), In: nil, Tag: "binary"},
+		{Type: &ifaceBinaryMarshal{}, Out: (*ifaceBinaryMarshal)(nil), In: nil, Tag: "binary"},
+		{Type: &ifaceTextMarshal{}, Out: (*ifaceTextMarshal)(nil), In: nil, Tag: "string"},
+		{Type: &ifaceJSONMarshal{}, Out: (*ifaceJSONMarshal)(nil), In: nil, Tag: "json"},
+		// Pointer interfaces with tags with value:
+		{Type: &ifaceMarshal{}, Out: &ifaceMarshal{"binary_one"}, In: []byte("binary_one"), Tag: "binary"},
+		{Type: &ifaceBinaryMarshal{}, Out: &ifaceBinaryMarshal{"binary_two"}, In: []byte("binary_two"), Tag: "binary"},
+		{Type: &ifaceTextMarshal{}, Out: &ifaceTextMarshal{"text"}, In: []byte("text"), Tag: "string"},
+		{Type: &ifaceJSONMarshal{}, Out: &ifaceJSONMarshal{[]string{"json"}}, In: []byte("[\"json\"]"), Tag: "json"},
```
changpingc: curious to see what happens if for some bizarre reason someone defines field type to be `Type: interface{}(nil)`

CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+### Changed
+
+#### `sqlgen`
+
+- Pointer scanners are no longer allowed to handle `nil`, are forcefully set to
+  `nil` instead.
+
+### Removed
+
+- `fields.Scanner.Interface()` - this should never have been exposed and
```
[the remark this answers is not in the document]
stephen: oh, yep. i missed the heading. seems fine to me

PR #155, opened 2018-08-27
go fields: handle time like we did before, with mysql.NullTime

berfarah:
See https://github.com/samsarahq/thunder/pull/142/files#diff-ec296ed9b760f4157d5adf8255a5c5c8L102

I looked through our other scalars, and we handle the rest of them (including bytes with our own behavior). This unfortunately couples us back to MySQL.

### DIFF
--- a/fields/sql.go
+++ b/fields/sql.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"reflect"
 	"time"
+
+	"github.com/go-sql-driver/mysql"
 )
 
 // Valuer fulfills the sql/driver.Valuer interface which deserializes our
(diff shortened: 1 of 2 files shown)


fields/sql.go
```diff
@@ -186,10 +188,12 @@ func (s *Scanner) Scan(src interface{}) error {
...
 		}
 	case *time.Time:
-		if _, ok := src.(time.Time); ok {
-			s.value.Set(reflect.ValueOf(src))
-			return nil
+		t := mysql.NullTime{}
+		if err := t.Scan(src); err != nil {
+			return err
 		}
+		s.value.Set(reflect.ValueOf(t.Time))
```
changpingc: i'm surprised we need to call into mysql driver here?

berfarah: We _could_ handle the various permutations of how time is stored ourselves to decouple this from the mysql driver.

changpingc: I don't think this scans nil correctly? We should check t.Valid and set s.isValid accordingly.

berfarah: Null is already handled on line 170

changpingc: Can we have a test case that shows we scan nil MySQL time into nil time.Time correctly?

berfarah: (non-pointer converts into zero value)

berfarah:
@stephen added

stephen:
thanks.

@changpingc can you articulate here why you'd prefer to not have callers handle this in the sql driver?

PR #160, opened 2018-08-29
go graphql: allow overriding min rerun interval per query

changpingc:


graphql/server.go
```diff
@@ -573,6 +575,12 @@ func WithSubscriptionLogger(logger SubscriptionLogger) ConnectionOption {
 	}
 }
 
+func WithMinRerunIntervalFunc(fn RerunIntervalFunc) ConnectionOption {
```
stephen: can we deprecate this

changpingc: done

PR #165, opened 2018-09-12
serialize dependencies into protobuf

changpingc:
We'd like to ship dependencies out of thunder server process, which allows us to restore a subscription and avoid expensive initial computation when a client reconnects and resumes prior subscriptions. This also enables external dependency invalidator that can make thunder server stateless.


berfarah:
@changpingc Do you mind putting the reverts in a separate PR? 😬

changpingc:
revert landed separately in #167.

sqlgen/reflect.go
```diff
@@ -815,3 +815,42 @@ func (t *Table) extractRow(row interface{}) Filter {
...
+	case string:
+		return string(v)
+	case []byte:
+		return []byte(v)
+	case float32:
+		return float64(v)
+	case float64:
+		return float64(v)
+	case time.Time:
+		return time.Time(v)
```
berfarah: There are some types here that we're doing nothing with. Do we need to handle them?

livesql/marshal.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/samsarahq/thunder/sqlgen"
+	"github.com/samsarahq/thunder/thunderpb"
+)
+
+func valueToField(value interface{}) (*thunderpb.Field, error) {
+	switch column := sqlgen.MakeCanonical(value).(type) {
```
berfarah: What happens for custom types?

reactive/rerunner.go
```diff
@@ -131,14 +133,43 @@ func (r *Resource) Cleanup(f func()) {
...
-func AddDependency(ctx context.Context, r *Resource) {
+type dependencySetKey struct{}
+
+type dependencySet struct {
+	mu           sync.Mutex
+	dependencies []Serializable
+}
+
+type Serializable interface {
+	proto.Message
```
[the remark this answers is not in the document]
changpingc: sure it can turn this into Marshal(), Unmarshal() probably

sqlgen/reflect.go
changpingc: this handles all driver.Value.

livesql/marshal.go
changpingc: Talked offline. We are going to support it by storing driver.Value. See last commit!

reactive/rerunner.go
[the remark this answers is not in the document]
changpingc: done

sqlgen/reflect.go
berfarah: This handles more than `driver.Value` - for example, `uint32` is not a `driver.Value`, only `int64` is

reactive/rerunner.go
```diff
@@ -131,14 +131,46 @@ func (r *Resource) Cleanup(f func()) {
...
 		r.node.addOut(&node{released: true})
 		return
 	}
 
 	computation := ctx.Value(computationKey{}).(*computation)
 	r.node.addOut(&computation.node)
+
+	if serializable != nil {
+		depSet, ok := ctx.Value(dependencySetKey{}).(*dependencySet)
+		if ok && depSet != nil {
```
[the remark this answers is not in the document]
changpingc: done

reactive/rerunner.go
```diff
@@ -131,14 +131,46 @@ func (r *Resource) Cleanup(f func()) {
...
+type dependencySetKey struct{}
+
+type dependencySet struct {
+	mu           sync.Mutex
+	dependencies []Serializable
+}
+
+type Serializable interface {
+	// proto.Marshaler and proto.Unmarshaler.
+	Marshal() ([]byte, error)
```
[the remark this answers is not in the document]
changpingc: I don't particularly like that..

sqlgen/reflect.go
changpingc: opened a separate PR https://github.com/samsarahq/thunder/pull/170

changpingc:
@berfarah I think this is ready to land! It should not introduce any additional error in existing API.

berfarah:
@changpingc given the amount of changes here, do you think a second pair of eyes might be useful?

reactive/rerunner.go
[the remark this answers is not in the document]
berfarah: Yeah but would also be a built-in interface. I'm ok with not wrapping.

livesql/marshal_test.go
```diff
@@ -0,0 +1,97 @@
...
+				"id": int32(1),
+			},
+			unmarshaled: sqlgen.Filter{
+				"id": int64(1),
+			},
+		},
+		{
+			name: "id int64 ptr to int64",
+			filter: sqlgen.Filter{
+				"id": &one,
```
berfarah: nil test case?

thunderpb/dependency.proto
```diff
@@ -0,0 +1,41 @@
...
+  Bool = 2;
+  Int = 3;
+  Uint = 4;
+  String = 5;
+  Bytes = 6;
+  Float64 = 7;
+  Time = 8;
+}
+
+message Field {
```
willhug: ahh, think I found what's causing a lot of cloudserver OOMs, I think this field is quite large and taking up much more memory than usual whenever we create a dependency.

Have we thought of using the proto oneof annotation here?  Not sure if it actually reduces the allocation in code, just a thought.

willhug: here's one of our heap profiles: https://s3.internal.samsara.com/s3/samsara-prod-profile/2018-10-02/cloudserver/1438-prod-23996c8a-f0f5-47be-960a-d96d639a2f26-2wHXqesa-heap.pprof.svg

PR #166, opened 2018-09-12
Reduce allocations for read path

berfarah:
Two commits here.

First one reduces allocations by 4 because we're no longer copying references for new scanners/valuers.

Second one reduces allocations by not copying our value, but rather allocating straight to a struct.


berfarah:
Note: I need to update some tests to pass. However, I'm not prioritizing this until I confirm on a staging server that this runs much more efficiently.

fields/sql.go
```diff
@@ -134,17 +129,21 @@ func (s *Scanner) CopyTo(to reflect.Value) {
...
+
+	if s.isValid && s.Ptr {
+		s.value.Set(reflect.New(s.Type))
+	}
+
 	// Get a value of the pointer of our type. The Scanner and Unmarshalers should
 	// only be implemented as dereference methods, since they would do nothing otherwise. Therefore
 	// we can safely assume that we should check for these interfaces on the pointer value.
-	s.value = reflect.New(s.Type)
+	// s.value = s.va
```
willhug: le comment

berfarah: le grazie

livesql/binlog.go
```diff
@@ -188,24 +188,30 @@ func buildColumnMap(conn *sql.DB, database string, table *sqlgen.Table) (*column
...
 
-	scanners := make([]*fields.Scanner, len(columnMap.source))
 	for i, j := range columnMap.source {
 		if j == -1 {
 			continue
 		}
-		scanner := table.Columns[i].Descriptor.Scanner()
+		column := table.Columns[i]
+		field := elem.FieldByIndex(column.Index)
+		if field.Kind() != reflect.Ptr && field.CanAddr() {
```
willhug: this this alright when it's not a pointer and we can't Addr it?  Or is that not physically possible?

berfarah: It _should_ not be possible. However, you're right - if it's not a pointer and we can't addr it, we should return an error here.

fields/descriptor.go
```diff
@@ -39,7 +39,12 @@ func (d *Descriptor) Valuer(val reflect.Value) Valuer {
 }
 
 // Scanner creates a sql.Scanner from the descriptor.
-func (d *Descriptor) Scanner() *Scanner { return &Scanner{Descriptor: d} }
+func (d *Descriptor) Scanner(value reflect.Value) *Scanner {
```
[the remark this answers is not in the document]
berfarah: AFAIK I removed this - but definitely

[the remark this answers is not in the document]
berfarah: This is gone now

fields/descriptor.go
```diff
@@ -51,23 +53,17 @@ func (d Descriptor) ValidateSQLType() error {
...
-	// Set non-pointer by setting reference
-	if !d.Ptr {
-		to.Set(from)
-		return
-	}
 
-	if !isValid {
-		to.Set(reflect.Zero(to.Type()))
-		return
+	// We need to hold onto this pointer-pointer in order to make the value addressable.
```
changpingc: reflect law 3 https://blog.golang.org/laws-of-reflection 👍

fields/sql.go
```diff
@@ -119,9 +119,8 @@ type Scanner struct {
 	isValid bool
 }
 
-// CopyTo copies the scanner value to another reflect.Value. This is used for setting structs.
-func (s *Scanner) CopyTo(to reflect.Value) {
-	s.copy(s.value, to, s.isValid)
+func (s *Scanner) To(value reflect.Value) {
```
changpingc: Not sure if `To` name makes sense. Maybe `Reset` makes more sense?

One example I see often is having a sync.Pool of compression writers, like https://github.com/ungerik/go-pool/blob/master/gzip.go#L11:31. They almost always have a Reset method to replace underlying writer/destination.

berfarah: I'm open to renaming this. Maybe it should be `Target`? Also, now that I think about it, maybe I should clear the `value` at the end of `Scan`.

livesql/binlog.go
berfarah: Hmm, actually, maybe I should remove `field.CanAddr()` and just have it panic if it can't. This should _always_ be able to Addr - the struct is even defined in this method

berfarah:
Benchmark before/after:

```
goos: darwin
goarch: amd64
pkg: github.com/samsarahq/thunder/sqlgen
Benchmark/Read-8         	    5000	    291292 ns/op	    2239 B/op	      60 allocs/op
Benchmark/Create-8       	    2000	    683282 ns/op	    1568 B/op	      39 allocs/op
Benchmark/Update-8       	    2000	    576854 ns/op	    1976 B/op	      49 allocs/op
Benchmark/Delete-8       	    2000	    643999 ns/op	    1000 B/op	      28 allocs/op
```

```
goos: darwin
goarch: amd64
pkg: github.com/samsarahq/thunder/sqlgen
Benchmark/Read-8         	    5000	    312561 ns/op	    1670 B/op	      43 allocs/op
Benchmark/Read_Where-8   	    2000	    606886 ns/op	    2348 B/op	      56 allocs/op
Benchmark/Create-8       	    2000	    529853 ns/op	    1120 B/op	      26 allocs/op
Benchmark/Update-8       	    3000	    509400 ns/op	    1496 B/op	      32 allocs/op
Benchmark/Delete-8       	    3000	    457721 ns/op	     808 B/op	      25 allocs/op
```

Improvements across the board 😄

sqlgen/reflect.go
```diff
@@ -51,16 +51,20 @@ const (
 )
 
 // UnbuildStruct extracts SQL values from a struct
-func UnbuildStruct(table *Table, strct interface{}) []interface{} {
+func UnbuildStruct(table *Table, strct interface{}) ([]interface{}, error) {
```
willhug: from the commit message:
```
Don't hold onto valuers, instead only use them to get values back. By
doing this, we're allocating the simpler (and more likely to be used)
value onto the heap, while only allocating Valuer onto the heap.
```

do we mean:

```
Don't hold onto valuers, instead only use them to get values back. By
doing this, we're allocating the simpler (and more likely to be used)
value onto the heap, while only allocating Valuer onto the **stack**.
```
?

willhug: If not maybe we should reword this.

berfarah: Cheers

berfarah: Corrected

sqlgen/reflect.go
```diff
@@ -51,16 +51,20 @@ const (
...
 	elem := reflect.ValueOf(strct).Elem()
 	values := make([]interface{}, len(table.Columns))
 
 	for i, column := range table.Columns {
 		val := elem.FieldByIndex(column.Index)
-		values[i] = column.Descriptor.Valuer(val)
+		var err error
+		values[i], err = column.Descriptor.Valuer(val).Value()
```
stephen: my takeaway: escaping a value is much cheaper than escaping a big struct?

berfarah: Both that and `sql` was going to call `Value` anyway before it sent the value to the sql driver, which means that the other value was going to be escaped either way - might as well do it as soon as possible so we don't allocate two different values

fields/descriptor.go
```diff
@@ -54,7 +54,7 @@ func (d Descriptor) ValidateSQLType() error {
 	return d.Scanner().Scan(val)
 }
 
-func (d Descriptor) copy(from, to reflect.Value, isValid bool) {
+func (d *Descriptor) copy(from, to reflect.Value, isValid bool) {
```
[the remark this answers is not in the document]
berfarah: My takeaway here is: everything that's passed into a function, including pointers, is pass by value in go. If we have `Descriptor`, we are copying a value (all struct fields) and then pointing to it. If we have `*Descriptor`, we are copying a pointer (just one thing pointing to the value). The second is lighter because we don't have to re-allocate the struct and its children, we just have to allocate the pointer.

PR #172, opened 2018-09-20
go {livesql,sqlgen,internal}: support []byte arguments for live queries

berfarah:
## Problem

We have UUIDs which use or convert to `[]byte`. Both livesql and the batcher rely on comparable/hashable values (values that are put into a `map`) to work and panic otherwise. This means that we can't use those custom types with livesql or batching.

### LiveSQL
The `ToArray` function has been used for creating hashable/comparable arguments in the past. Renamed it to be more explicit, and also handle `[]byte` arguments, converting them into strings.

The primary use-case for this PR is to support UUIDs. We also get support for anything else that turns into `[]byte`.

The reason we want to serialize our arguments/filters is because (unfortunately) we support type aliases in our filters, and therefore allow you to query a `int64` with `int`/`int8`/etc. In order to maintain backward compatibility there, we have to assume that we're dealing with a SQL value for caching/rerunning.

### Batching

TL;DR: do nothing

Batching deals with go struct types (not go sql types), which means we have many possible permutations of types. We _could_ take our filter values and convert them to their sql values, using that to compare. This would take more time than simply using a `[16]byte` for our UUIDs.

It would properly implement 100% compatibility with custom types. However, I have some concerns about performance in doing so, and given deadline constraints, we can add that on later rather than doing so now.

**Slices in filters will continue not to be supported for batching.**

### DIFF
--- a/internal/reflect.go
+++ b/internal/reflect.go
@@ -31,9 +31,9 @@ func init() {
 	interfaceTyp = reflect.TypeOf(&x).Elem()
 }
 
-// ToArray converts a []interface{} slice into an equivalent fixed-length array
+// MakeHashable converts a []interface{} slice into an equivalent fixed-length array
 // [...]interface{} for use as a comparable map key
-func ToArray(s []interface{}) interface{} {
+func MakeHashable(s []interface{}) interface{} {
 	d := make([]interface{}, len(s))
 	// Convert byte slices into strings as they are otherwise not comparable/hashable.
 	for i, elem := range s {
--- a/internal/reflect.go
+++ b/internal/reflect.go
@@ -33,32 +33,42 @@ func init() {
 
 // ToArray converts a []interface{} slice into an equivalent fixed-length array
 // [...]interface{} for use as a comparable map key
-//
 func ToArray(s []interface{}) interface{} {
-	switch len(s) {
+	d := make([]interface{}, len(s))
+	// Convert byte slices into strings as they are otherwise not comparable/hashable.
+	for i, elem := range s {
+		if b, ok := elem.([]byte); ok {
+			d[i] = string(b)
+		} else {
+			d[i] = elem
+		}
+	}
+
+	// Return arrays as they are comparable/hashable.
+	switch len(d) {
 	// fast code paths for short arrays:
 	case 0:
 		return [...]interface{}{}
 	case 1:
-		return [...]interface{}{s[0]}
+		return [...]interface{}{d[0]}
 	case 2:
-		return [...]interface{}{s[0], s[1]}
+		return [...]interface{}{d[0], d[1]}
 	case 3:
-		return [...]interface{}{s[0], s[1], s[2]}
+		return [...]interface{}{d[0], d[1], d[2]}
 	case 4:
-		return [...]interface{}{s[0], s[1], s[2], s[3]}
+		return [...]interface{}{d[0], d[1], d[2], d[3]}
 	case 5:
-		return [...]interface{}{s[0], s[1], s[2], s[3], s[4]}
+		return [...]interface{}{d[0], d[1], d[2], d[3], d[4]}
 	case 6:
-		return [...]interface{}{s[0], s[1], s[2], s[3], s[4], s[5]}
+		return [...]interface{}{d[0], d[1], d[2], d[3], d[4], d[5]}
 	case 7:
-		return [...]interface{}{s[0], s[1], s[2], s[3], s[4], s[5], s[6]}
+		return [...]interface{}{d[0], d[1], d[2], d[3], d[4], d[5], d[6]}
 	case 8:
-		return [...]interface{}{s[0], s[1], s[2], s[3], s[4], s[5], s[6], s[7]}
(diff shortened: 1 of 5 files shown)


livesql/integration_test.go
```diff
@@ -90,3 +90,76 @@ func TestIntegrationBasic(t *testing.T) {
...
+	result, err := db.InsertRow(context.Background(), &User{Name: name, Uuid: uuid, Mood: &mood})
+	assert.NoError(t, err)
+	userId, err := result.LastInsertId()
+	assert.NoError(t, err)
+
+	// We use a channel to pass around query updates for testing.
+	users := make(chan *User)
+	rerunner := reactive.NewRerunner(context.Background(), func(ctx context.Context) (interface{}, error) {
+		user := &User{}
+		if err := db.QueryRow(ctx, &user, sqlgen.Filter{"mood": &mood}, nil); err != nil {
```
berfarah: This means UUIDs should definitely work

livesql/integration_test.go
```diff
@@ -90,3 +90,76 @@ func TestIntegrationBasic(t *testing.T) {
...
+	newUuid := testfixtures.CustomTypeFromString("11232481203")
+
+	result, err := db.InsertRow(context.Background(), &User{Name: name, Uuid: uuid, Mood: &mood})
+	assert.NoError(t, err)
+	userId, err := result.LastInsertId()
+	assert.NoError(t, err)
+
+	// We use a channel to pass around query updates for testing.
+	users := make(chan *User)
+	rerunner := reactive.NewRerunner(context.Background(), func(ctx context.Context) (interface{}, error) {
```
changpingc: try using batch.WithBatching(ctx)

changpingc:
cute idea: check reflect.Type.Comparable()

internal/reflect.go
```diff
@@ -31,10 +31,17 @@ func init() {
...
 }
 
-// ToArray converts a []interface{} slice into an equivalent fixed-length array
+// MakeHashable converts a []interface{} slice into an equivalent fixed-length array
 // [...]interface{} for use as a comparable map key
-//
-func ToArray(s []interface{}) interface{} {
+func MakeHashable(s []interface{}) interface{} {
+	// Convert byte slices into strings as they are otherwise not comparable/hashable.
+	for i, elem := range s {
```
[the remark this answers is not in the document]
berfarah: Hmm, I think I didn't quite get pass by value here - I'll change this so it makes a new slice and check some benchmarks.

internal/reflect.go
```diff
@@ -31,10 +31,17 @@ func init() {
 	interfaceTyp = reflect.TypeOf(&x).Elem()
 }
 
-// ToArray converts a []interface{} slice into an equivalent fixed-length array
+// MakeHashable converts a []interface{} slice into an equivalent fixed-length array
 // [...]interface{} for use as a comparable map key
-//
-func ToArray(s []interface{}) interface{} {
+func MakeHashable(s []interface{}) interface{} {
```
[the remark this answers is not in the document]
berfarah: My concern would be breaking things at the SQL layer if I put it higher up. If I were to pick out a refactor that could be worth doing, it's making the filter more responsible for its various conversions so that we don't have to hold in our heads what these things ^ are used for.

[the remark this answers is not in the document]
berfarah: Looks like batching doesn't actually use the `driver.Value`. Need to find a way to do so.

berfarah:
@stephen for batching, right now the code compares non-sql-value types, which means identifiers would have to be comparable. This means that without changing our uuids from `[]byte` to `[16]byte`, batching is broken for UUIDs. I think what I want to do is go with the UUID change for now and take more time to consider our various (de)serialization layers.

berfarah:
@stephen I'm ready to merge this as soon as I have a 👍

internal/reflect.go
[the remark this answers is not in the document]
berfarah: This no longer modifies the original slice

internal/reflect.go
```diff
@@ -31,34 +31,44 @@ func init() {
...
+	}
+
+	// Return arrays as they are comparable/hashable.
+	switch len(d) {
 	// fast code paths for short arrays:
 	case 0:
 		return [...]interface{}{}
 	case 1:
-		return [...]interface{}{s[0]}
+		return [...]interface{}{d[0]}
```
stephen: super nit: consider changing out `s` -> `d` in the arguments instead so that these references stay the same.

stephen: (i also checked _find all references_ so it seems fine to leave)

internal/reflect.go
```diff
@@ -31,34 +31,44 @@ func init() {
 	interfaceTyp = reflect.TypeOf(&x).Elem()
 }
 
-// ToArray converts a []interface{} slice into an equivalent fixed-length array
+// MakeHashable converts a []interface{} slice into an equivalent fixed-length array
```
stephen: nit: consider doing the rename in a separate step to make the semantic change easier to read

berfarah: Done

PR #173, opened 2018-09-25
fields: fix sql type validation for pointer receiver methods

stephen:
This PR fixes a bug where we don't use a pointer to struct when verifying that the input type is a valid sql type. This occurs when the method receiver is a pointer AND the type passed is a pointer. The cases for non-pointer method receivers already works fine.

### DIFF
--- a/internal/fields/sql_test.go
+++ b/internal/fields/sql_test.go
@@ -219,6 +219,7 @@ func TestField_ValidateSQLType(t *testing.T) {
 	integer := int64(0)
 	cases := []struct {
 		In    interface{}
+		Tags  []string
 		Error bool
 	}{
 		{In: "foo", Error: false},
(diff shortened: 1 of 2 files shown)


internal/fields/sql_test.go
```diff
@@ -219,17 +219,23 @@ func TestField_ValidateSQLType(t *testing.T) {
...
 		Error bool
 	}{
 		{In: "foo", Error: false},
 		{In: int16(0), Error: false},
 		{In: &integer, Error: false},
 		{In: [4]byte{}, Error: true},
 		{In: ifaceBinaryMarshal{}, Error: true},
+		{In: ifaceBinaryMarshal{}, Tags: []string{"binary"}, Error: false},
+		{In: &ifaceBinaryMarshal{}, Tags: []string{"binary"}, Error: false},
+		// Non-pointer proto does not work because the Marshal() method has a pointer receiver.
```
berfarah: It'd probably be nice to have the other SQL types in here too

stephen: let's do that as a follow up.

PR #176, opened 2018-09-27
sqlgen/livesql: Improve error logging

berfarah:
1. Allow for the destination of error output to be customized
2. Provide table and column information when parsing fails in `livesql` and `sqlgen`

### DIFF
--- a/livesql/binlog.go
+++ b/livesql/binlog.go
@@ -215,7 +215,7 @@ func parseBinlogRow(table *sqlgen.Table, binlogRow []interface{}, columnMap *col
 		scanner := scanners[i].(*fields.Scanner)
 		scanner.Target(field)
 		if err := scanner.Scan(binlogRow[j]); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("binlog: `%s`.`%s` error: %v", table.Name, table.Columns[i].Name, err)
 		}
 	}
 
--- a/livesql/binlog.go
+++ b/livesql/binlog.go
@@ -7,13 +7,13 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"log"
 	"reflect"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/samsarahq/thunder/internal/fields"
+	"github.com/samsarahq/thunder/logger"
 	"github.com/samsarahq/thunder/sqlgen"
 	"github.com/siddontang/go-mysql/mysql"
 	"github.com/siddontang/go-mysql/replication"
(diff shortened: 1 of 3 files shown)


livesql/binlog.go
```diff
@@ -211,7 +216,7 @@ func parseBinlogRow(table *sqlgen.Table, binlogRow []interface{}, columnMap *col
 		scanner := scanners[i].(*fields.Scanner)
 		scanner.Target(field)
 		if err := scanner.Scan(binlogRow[j]); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("binlog: `%s`.`%s` error: %v", table.Name, table.Columns[i].Name, err)
```
willhug: if we want things to be quoted, we can use %q and it will sanitize strings in quotes.

livesql/binlog.go
```diff
@@ -327,6 +332,12 @@ func (b *Binlog) parseBinlogRowsEvent(event *replication.BinlogEvent) (*update,
 	return update, nil
 }
 
+// SetOutput sets the destination for the error logger.
+func (b *Binlog) SetOutput(w io.Writer) {
+	b.logwriter = w
```
willhug: do we want to define the logwriter interface on an io.Writer?  Should we instead define a "Logger" interface and give it the functions we want to use.  In which case, we can hook up the "Logger" to a Zap (or other) logger under the hood.

For the base case, we can create a StdOutLogger that logs directly to stdout.

sqlgen/reflect.go
```diff
@@ -61,7 +61,7 @@ func UnbuildStruct(table *Table, strct interface{}) ([]interface{}, error) {
 		var err error
 		values[i], err = column.Descriptor.Valuer(val).Value()
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("sqlgen: serialization error for `%s`.`%s`: %v", table.Name, column.Name, err)
```
willhug: slightly worried about printf performance (since it does reflection), not a blocker though

livesql/binlog.go
berfarah: I was just trying to give a sense of what the tables look like

livesql/binlog.go
berfarah: Given that this is open source, I was hoping to provide an interface that's pretty flexible that anyone could use. Do you know of a widely used logging interface?

berfarah: Going to add an interface that supports `Error` and `Info`.

berfarah:
@willhug logger interface look right?

PR #177, opened 2018-10-02
use oneof for thunderpb.Field

changpingc:
To cut down memory usage. 

### Test plan

I will vendor this and make sure inflight snapshot still loads.


berfarah:
LGTM - but want to make sure @willhug has a look as well. Thanks for adding the changelog entries! 😄

changpingc:
Verified snapshot works still. Merging.

PR #181, opened 2018-10-04
go sqlgen: Add support for converting "Zero" value fields to NULL in the db

willhug:
Summary: Adds support to sqlgen to convert "Zero" value model fields to NULL in 
the database.  This only needs to be applied when inserting a value to the 
database because when we read from the db NULL values are noop (leaving the 
zero value).

I put some cleanups in the PR as well, would be easy to pull them out if we want.

Additionally, I wrote tests for the "Main" use cases, but I'm very open 
to add tests for other edge cases, let me know and I'll add them.

### DIFF
--- a/internal/fields/reflect.go
+++ b/internal/fields/reflect.go
@@ -1,0 +1,35 @@
+package fields
+
+import (
+	"reflect"
+)
+
+// isZero returns whether the value of the passed in reflect value is the
+// zero value of that type.
+// Mostly copied from https://github.com/golang/go/issues/7501
+// Works for mostly everything except zero initialized arrays (e.g.
+// `var ray [5]string` will not be "Zero")
+func isZero(v reflect.Value) bool {
+	switch v.Kind() {
+	case reflect.Array, reflect.String:
+		return v.Len() == 0
+	case reflect.Bool:
+		return !v.Bool()
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return v.Int() == 0
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		return v.Uint() == 0
+	case reflect.Float32, reflect.Float64:
+		return v.Float() == 0
+	case reflect.Interface, reflect.Map, reflect.Ptr, reflect.Slice:
+		return v.IsNil()
+	case reflect.Struct:
+		// This is a special case comparison for struct types.
+		return v.Interface() == reflect.Zero(v.Type()).Interface()
+	case reflect.Invalid:
+		// Invalid types are, by definition, "Zero" values
+		return true
+	}
+
+	return false
+}
--- a/internal/fields/sql.go
+++ b/internal/fields/sql.go
@@ -77,6 +77,10 @@ func (f Valuer) Value() (driver.Value, error) {
 			return iface.MarshalJSON()
 		}
 		return json.Marshal(i)
+	case f.Tags.Contains("implicitnull"):
+		if isZero(f.value) {
+			return nil, nil
+		}
 	}
 
 	// At this point we have already handled `nil` above, so we can assume that all
--- a/internal/fields/sql.go
+++ b/internal/fields/sql.go
@@ -115,8 +115,7 @@ var _ driver.Valuer = &Valuer{}
 // into the type dictated by our descriptor.
 type Scanner struct {
 	*Descriptor
-	value   reflect.Value
-	isValid bool
+	value reflect.Value
 }
 
 func (s *Scanner) Target(value reflect.Value) {
@@ -137,9 +136,9 @@ func (s *Scanner) Scan(src interface{}) error {
 	defer func() { s.value = reflect.Value{} }()
 
 	// Keep track of whether our value was empty.
-	s.isValid = src != nil
+	isValid := src != nil
 
-	if s.isValid && s.Ptr {
+	if isValid && s.Ptr {
 		s.value.Set(reflect.New(s.Type))
 	}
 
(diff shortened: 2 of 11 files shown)


internal/fields/sql.go
```diff
@@ -77,6 +77,10 @@ func (f Valuer) Value() (driver.Value, error) {
 			return iface.MarshalJSON()
 		}
 		return json.Marshal(i)
+	case f.Tags.Contains("zeroIsNull"):
```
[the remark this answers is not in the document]
berfarah: As per offline convo: `implicitnull`?

internal/fields/sql.go
```diff
@@ -115,8 +119,7 @@ var _ driver.Valuer = &Valuer{}
 // into the type dictated by our descriptor.
 type Scanner struct {
 	*Descriptor
-	value   reflect.Value
-	isValid bool
```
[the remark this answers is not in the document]
willhug: Yup!

sqlgen/zeronull_test.go
```diff
@@ -0,0 +1,187 @@
+package sqlgen
+
+import (
```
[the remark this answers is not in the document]
willhug: Yeah, adding some tests here and seeing edge cases, will update PR

willhug:
Adding some tests for filters to make sure they still work properly.

internal/fields/sql.go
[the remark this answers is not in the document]
berfarah: I would prefer all lowercase since that's how go's stdlib does it (`omitempty`).

willhug:
Updated!  Also added a db test thing I've been using locally to get tests to run without an EnvVar every time.

internal/testfixtures/db.go
```diff
@@ -28,6 +28,8 @@ var DefaultDBConfig = DBConfig{
 func init() {
 	if pw := os.Getenv("DB_PASSWORD"); pw != "" {
 		DefaultDBConfig.Password = pw
+	} else {
```
berfarah: We can change the password to just always be dev and update the docker container on here?

sqlgen/integration_test.go
```diff
@@ -201,10 +204,11 @@ func Benchmark(b *testing.B) {
...
 	user := &User{
-		Id:   1,
-		Name: "Bob",
-		Uuid: testfixtures.CustomType{'1', '1', '2', '3', '8', '4', '9', '1', '2', '9', '3'},
-		Mood: &mood,
+		Id:           1,
+		Name:         "Bob",
+		Uuid:         testfixtures.CustomType{'1', '1', '2', '3', '8', '4', '9', '1', '2', '9', '3'},
+		Mood:         &mood,
+		ImplicitNull: "test",
```
berfarah: Hmm - do we want this in our benchmark?

sqlgen/implicitnull_test.go
```diff
@@ -0,0 +1,238 @@
...
+	newRow.Id = 1
+
+	var rows []*ImplicitNull
+	require.NoError(t, db.Query(context.Background(), &rows, nil, nil))
+	assert.Equal(t, []*ImplicitNull{
+		newRow,
+	}, rows)
+}
+
+func GetCount(db *sql.DB, countQuery string) (int, error) {
```
stephen: nit: don't export this?

internal/fields/sql.go
```diff
@@ -77,6 +77,10 @@ func (f Valuer) Value() (driver.Value, error) {
 			return iface.MarshalJSON()
 		}
 		return json.Marshal(i)
+	case f.Tags.Contains("implicitnull"):
+		if IsZero(f.value) {
```
stephen: i'm excited by how small this change is that i missed it in my first read.

internal/fields/reflect.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"reflect"
+)
+
+// IsZero returns whether the value of the passed in reflect value is the
+// zero value of that type.
+// Mostly copied from https://github.com/golang/go/issues/7501
+// Works for mostly everything except zero initialized arrays (e.g.
+// `var ray [5]string` will not be "Zero")
+func IsZero(v reflect.Value) bool {
```
stephen: nit: should we make this private?

sqlgen/implicitnull_test.go
```diff
@@ -0,0 +1,238 @@
...
+	timeField := time.Date(2000, 10, 3, 0, 0, 0, 0, time.UTC)
+	newRow := &ImplicitNull{
+		NullStr:   "hello",
+		NullInt:   123,
+		NullFloat: 1.23,
+		NullBool:  true,
+		NullByte:  []byte("hello there"),
+		NullTime:  timeField,
+	}
+	if _, err := db.InsertRow(context.Background(), newRow); err != nil {
```
stephen: could we add a test for `UpdateRow` as well?

internal/fields/sql.go
```diff
@@ -77,6 +77,10 @@ func (f Valuer) Value() (driver.Value, error) {
 			return iface.MarshalJSON()
 		}
 		return json.Marshal(i)
+	case f.Tags.Contains("implicitnull"):
```
stephen: nit: what is implicit about this null? i kind of like `zeroisnull`, but i could be convinced otherwise

sqlgen/implicitnull_test.go
```diff
@@ -0,0 +1,238 @@
...
+			require.NoError(t, db.QueryRow(ctx, &imp, Filter{"null_byte": []byte("hello there")}, nil))
+			require.NoError(t, db.QueryRow(ctx, &imp, Filter{"null_time": timeField}, nil))
+
+			newRow.Id = 1
+			var rows []*ImplicitNull
+			require.NoError(t, db.Query(context.Background(), &rows, nil, nil))
+			assert.Equal(t, []*ImplicitNull{
+				newRow,
+			}, rows)
+		})
```
stephen: could we also add a test for the reading that a `null` in the db reads out as the zero value?

sqlgen/implicitnull_test.go
willhug: 👍

internal/fields/reflect.go
willhug: 👍

internal/fields/sql.go
willhug: No strong feelings either way, @berfarah @stephen whichever preference you guys want.

sqlgen/implicitnull_test.go
willhug: 👍

internal/testfixtures/db.go
willhug: Yes

willhug:
Not to self: binlog and Delete still need tests.

PR #184, opened 2018-10-08
go diff: don't panic on nil array value

berfarah:
## What changed?

When passed an array with nil values, `diff.Diff` would panic. This no longer happens.

## Test plan

Updated unit tests to add breaking case, the diff seems reasonable. The tests also worked without updating the tests file, which means there is no breaking behavior here.

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -182,6 +182,9 @@ func diffMap(old map[string]interface{}, newAny interface{}) interface{} {
 
 // reoderKey returns the key to use for a
 func reorderKey(i interface{}) interface{} {
+	if i == nil {
+		return i
+	}
 	if object, ok := i.(map[string]interface{}); ok {
 		if key, ok := object["__key"]; ok {
 			return key
--- a/diff/diff_test.go
+++ b/diff/diff_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/samsarahq/thunder/diff"
 	"github.com/samsarahq/thunder/internal"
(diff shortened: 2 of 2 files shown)


diff/diff_test.go
```diff
@@ -144,11 +146,13 @@ func TestDiffListOrder(t *testing.T) {
 		map[string]interface{}{"__key": "1"},
 		map[string]interface{}{"__key": "2"},
 		map[string]interface{}{"__key": "3"},
+		nil,
```
changpingc: have you thought about whether nil should be here or at the beginning? should we think about whether `merge` reverses this correctly? should we worry about client-side implementation of merge as well?

berfarah: If I put the `nil` at the beginning the test fails. I didn't consider this at all. Is there a way we can prevent `nil` from making it in here?

diff/diff.go
```diff
@@ -182,6 +182,9 @@ func diffMap(old map[string]interface{}, newAny interface{}) interface{} {
 
 // reoderKey returns the key to use for a
 func reorderKey(i interface{}) interface{} {
+	if i == nil {
```
berfarah: @jellevandenhooff per offline convo: the panic originates at `reflect.TypeOf`. If we have a typed nil (eg: `(*string)(nil)`) `reflect.TypeOf` does not panic.

diff/diff_test.go
berfarah: @changpingc this has been addressed.

PR #185, opened 2018-10-09
go fields: explicity support non-pointer marshalers

berfarah:
Because we sometimes might want to use non-pointers for protobuf, check our non-pointer types.

Test plan:
- Create benchmark for the above - maybe we should have benchmarks per supported value
- Run simulation traffic if the benchmarks look significantly different (more than 5%)

### DIFF
--- a/internal/fields/sql.go
+++ b/internal/fields/sql.go
@@ -19,6 +19,17 @@ type Valuer struct {
 	value reflect.Value
 }
 
+var marshalerType = reflect.TypeOf((*marshaler)(nil)).Elem()
+
+func nonPointerMarshal(d *Descriptor, val reflect.Value) (reflect.Value, bool) {
+	if !d.Ptr && reflect.PtrTo(d.Type).Implements(marshalerType) {
+		v := reflect.New(d.Type)
+		v.Elem().Set(val)
+		return v, true
+	}
+	return val, false
+}
+
 // Value satisfies the sql/driver.Valuer interface.
 // The value should be one of the following:
 //    int64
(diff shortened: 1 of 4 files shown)


internal/fields/sql.go
```diff
@@ -19,6 +19,18 @@ type Valuer struct {
 	value reflect.Value
 }
 
+var marshalerType = reflect.TypeOf((*marshaler)(nil)).Elem()
+
+func nonPointerMarshal(d *Descriptor, val reflect.Value) (reflect.Value, bool) {
```
berfarah: TODO: add a comment here explaining exactly what this is doing

internal/fields/sql.go
```diff
@@ -19,6 +19,18 @@ type Valuer struct {
...
+	if !d.Ptr && reflect.PtrTo(d.Type).Implements(marshalerType) {
+		fmt.Printf("typ: %+v, intf %v, intftyp: %T\n", d, val.Interface(), val.Interface())
```
stephen: nit: remove print?

internal/fields/sql_test.go
```diff
@@ -229,8 +229,7 @@ func TestField_ValidateSQLType(t *testing.T) {
 		{In: ifaceBinaryMarshal{}, Error: true},
 		{In: ifaceBinaryMarshal{}, Tags: []string{"binary"}, Error: false},
 		{In: &ifaceBinaryMarshal{}, Tags: []string{"binary"}, Error: false},
-		// Non-pointer proto does not work because the Marshal() method has a pointer receiver.
```
stephen: love it

internal/fields/sql.go
willhug: +1
Naming the return types might help too (especially the bool)

berfarah: 👍

berfarah:
As noted in my second commit with the benchmark, we see an increase of 1 allocation on the write path here (5 vs the typical 4). We're explicitly ok with this cost, given that:
1. We prefer being able to support non-pointers
2. It's opt-in (`sql:",binary"`)
3. It's on the write path

PR #186, opened 2018-10-10
graphql: oops wrap json errors with message

stephen:
Helpful for debugging c.handle errors logged. As a follow up, we should
fix up this code path so that errors can be handled by the client code.

### DIFF
--- a/graphql/server.go
+++ b/graphql/server.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
+	"github.com/samsarahq/go/oops"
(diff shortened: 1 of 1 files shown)


graphql/server.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/gorilla/websocket"
+	"github.com/samsarahq/go/oops"
```
berfarah: Is this a new dependency?

stephen: yes

stephen:
Think we can revisit if/when go2 takes a stance.

PR #187, opened 2018-10-11
graphql: make previous computation mutable by middlewares

zdylag:
Allow middlewares to mutate the previous computation output, for altering the diff sent back.

Also expose whether this is an initial computation to the middlewares.

### DIFF
--- a/graphql/server.go
+++ b/graphql/server.go
@@ -220,15 +220,18 @@ func (c *conn) handleSubscribe(in *inEnvelope) error {
 			return output
 		})
 
-		output := runMiddlewares(middlewares, &ComputationInput{
-			Ctx:         ctx,
-			Id:          id,
-			ParsedQuery: query,
-			Previous:    previous,
-			Query:       subscribe.Query,
-			Variables:   subscribe.Variables,
-			Extensions:  in.Extensions,
-		})
+		computationInput := &ComputationInput{
+			Ctx:                  ctx,
+			Id:                   id,
+			ParsedQuery:          query,
+			Previous:             previous,
+			IsInitialComputation: initial,
+			Query:                subscribe.Query,
+			Variables:            subscribe.Variables,
+			Extensions:           in.Extensions,
+		}
+
+		output := runMiddlewares(middlewares, computationInput)
 		current, err := output.Current, output.Error
 
 		c.logger.FinishExecution(ctx, tags, time.Since(start))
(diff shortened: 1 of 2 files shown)


graphql/server.go
```diff
@@ -269,7 +272,7 @@ func (c *conn) handleSubscribe(in *inEnvelope) error {
 			return nil, err
 		}
 
-		d := diff.Diff(previous, current)
+		d := diff.Diff(computationInput.Previous, current)
```
jellevandenhooff: this is getting pretty spooky but I am hopeful we'll fix it all up once we're done this with transformation

graphql/server.go
```diff
@@ -220,15 +220,18 @@ func (c *conn) handleSubscribe(in *inEnvelope) error {
...
-			Query:       subscribe.Query,
-			Variables:   subscribe.Variables,
-			Extensions:  in.Extensions,
-		})
+		computationInput := &ComputationInput{
+			Ctx:                  ctx,
+			Id:                   id,
+			ParsedQuery:          query,
+			Previous:             previous,
+			IsInitialComputation: previous == nil,
```
stephen: can we change the mutation handling codepath as well? It should always return `true` I think.

graphql/server.go
```diff
@@ -269,19 +272,26 @@ func (c *conn) handleSubscribe(in *inEnvelope) error {
...
 				ID:       id,
 				Type:     "update",
 				Message:  d,
 				Metadata: output.Metadata,
 			})
+		} else if initial {
+			c.writeOrClose(outEnvelope{
+				ID:       id,
+				Type:     "update",
+				Message:  struct{}{},
```
changpingc: can we change diff to return struct{}{} if there's no diff, and have a unit test that shows merge(any, struct{}{}) returns any?

changpingc: talked offline, we probably don't want to change diff behavior in case there's any external dependency, especially given there's a test case showing the desired behavior of diff(x, x) is nil.

changpingc: on that note, what if we do
```
if d == nil && initial {
  d = struct{}{}
}
```

graphql/server.go
```diff
@@ -377,6 +392,7 @@ func (c *conn) handleMutate(in *inEnvelope) error {
 
 		go c.rerunSubscriptionsImmediately()
 
+		initial = false
```
stephen: i don't think this variable is necessary. mutations never re-run, so they will never have a non-initial computation

PR #191, opened 2018-10-23
graphql: export ErrorCause helper

stephen:
This PR exports `ErrorCause` which allows client code to extract the root, unwrapped error for checking. It is useful for checking root case, e.g.:

```golang
if graphql.ErrorCause(err) == context.Canceled {
  // ...
}
```

### DIFF
--- a/graphql/executor.go
+++ b/graphql/executor.go
@@ -38,7 +38,7 @@ func nestPathError(key string, err error) error {
 	}
 }
 
-func extractPathError(err error) error {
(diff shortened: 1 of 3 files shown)


stephen:
how did test coverage decrease??

berfarah:
@stephen Can you add this to the changelog? Maybe we should make a new thunder release

stephen:
I added an entry to the changelog

PR #194, opened 2018-11-01
client: port merge test from go

amonks:
it's scary that this wasn't tested, but, hey! it passes!

### DIFF
--- a/client/src/tests/merge.spec.ts
+++ b/client/src/tests/merge.spec.ts
@@ -1,0 +1,72 @@
+import { merge } from "../merge";
+
+const cases: Array<{
+  title: string;
+  base: any;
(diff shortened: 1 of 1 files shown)


client/src/tests/merge.spec.ts
```diff
@@ -0,0 +1,72 @@
+import { merge } from "../merge";
+
+const cases: Array<{
+  title: string;
+  left: any;
+  right: any;
```
stephen: nit: consider `base` and `diff` or `patch` may be clearer

amonks: yeah you're right

amonks: done

PR #198, opened 2018-11-15
go reactive: dependency does not need to be serializable

changpingc:
We see keeping a dependency set of protobuf message uses
a lot of memory. We think building protobuf messages has
significant overhead. Instead, we will export a QueryDependency
type in livesql, and it will be up the user to call method in livesql
to turn them into protobuf messages.

### DIFF
--- a/livesql/marshal.go
+++ b/livesql/marshal.go
@@ -69,8 +69,8 @@ func FieldToValue(field *thunderpb.Field) (driver.Value, error) {
 	}
 }
 
-// filterToProto takes a sqlgen.Filter, runs Valuer on each filter value, and returns a thunderpb.SQLFilter.
-func filterToProto(schema *sqlgen.Schema, tableName string, filter sqlgen.Filter) (*thunderpb.SQLFilter, error) {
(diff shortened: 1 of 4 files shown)


livesql/marshal.go
```diff
@@ -69,8 +69,8 @@ func FieldToValue(field *thunderpb.Field) (driver.Value, error) {
 	}
 }
 
-// filterToProto takes a sqlgen.Filter, runs Valuer on each filter value, and returns a thunderpb.SQLFilter.
-func filterToProto(schema *sqlgen.Schema, tableName string, filter sqlgen.Filter) (*thunderpb.SQLFilter, error) {
+// FilterToProto takes a sqlgen.Filter, runs Valuer on each filter value, and returns a thunderpb.SQLFilter.
```
stephen: could you show an example for expected usage on these by the client code?

changpingc:
```
deps := reactive.Dependencies(ctx)
for _, dep := range deps {
  if sqldep, ok := dep.(livesql.QueryDependency); ok {
    p, err := livesql.FilterToProto(schema, sqldep.Table, sqldep.Filter)
    // ship p elsewhere.
  }
}
```

PR #201, opened 2018-11-20
Fix for "Bug: FieldFunc selectionSet argument is always nil"

### LINKED ISSUE #200: FieldFunc selectionSet argument is always nil
FieldFunc() should receive SelectionSet for particular node so that the resolver could build the appropriate query based on user's selection, but selectionSet is always nil.

Code example:

`query.FieldFunc("viewer", func(ctx context.Context, selectionSet *graphql.SelectionSet) *User {
	// selectionSet is always nil
})`

Dorokhov:
Fix #200


kvap:
Would be nice to have a unit test for that.

uripre:
@Dorokhov @kvap Would be great to see this crossing the finish line

PR #199, opened 2018-11-17
Pagination: Refactor

berfarah:
I refactored this a little bit as part of the process for understanding what's going on. I've added documentation and extracted individual components to keep cognitive load a little lighter when reading this.

Next step: adding filters / sorting

PR #202, opened 2018-11-20
Implement Sort and Filter

berfarah:
This adds in the ability to sort and filter on any thunder-managed paginated FieldFunc.

The first two commits are quality of life changes that make testing easier. They account for the biggest chunk of the diff here. The actual diff ends up being 326 LOC of non-test code.


berfarah:
Note: I've put the merge branch as `bf/pagination-filter` for now, but I'll change it to master once https://github.com/samsarahq/thunder/pull/199 has been merged.

berfarah:
Also it seems like the coveralls coverage here isn't accurate since it's counting zero pagination code as covered.

graphql/schemabuilder/pagination.go
```diff
@@ -422,15 +422,15 @@ func (c *connectionContext) applyTextFilter(nodes []interface{}, args Pagination
 		// For each possible field we're matching against, check if there's a match.
 		for _, filterField := range c.FilterTextFields {
 			// Resolve the graphql.Field made for sorting.
-			text, err := filterField.Resolve(context.TODO(), node, nil, nil)
```
berfarah: This context ensures we keep live data / caching going

graphql/schemabuilder/pagination.go
```diff
@@ -422,15 +422,15 @@ func (c *connectionContext) applyTextFilter(nodes []interface{}, args Pagination
...
+			text, err := filterField.Resolve(ctx, node, nil, nil)
 			if err != nil {
 				return nil, err
 			}
 
 			// Only strings are allowed for FilterText fields.
 			textString, ok := text.(string)
 			if !ok {
-				return nil, errors.New("Text was not a string")
+				return nil, errors.New("filter field didn't return a string")
```
berfarah: I can move this into a different commit

graphql/schemabuilder/pagination.go
```diff
@@ -461,15 +461,15 @@ func (c *connectionContext) applySort(nodes []interface{}, args PaginationArgs)
 	sortField, ok := c.SortFields[*args.SortBy]
 	// If the field wasn't registered, it's an unknown sort field.
 	if !ok {
-		return nil, fmt.Errorf("Unknown sort field %s", *args.SortBy)
+		return nil, fmt.Errorf("unknown sort field %s", *args.SortBy)
```
berfarah: Same

graphql/schemabuilder/pagination.go
```diff
@@ -372,6 +380,62 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
 
+var matchGroups = regexp.MustCompile(`(?:([^\s"]+)|"([^"]*)"?)+`)
+
+// Split terms by spaces, except for spaces within quotes.
+// Like http://stackoverflow.com/questions/16261635/javascript-split-string-by-space-but-ignore-space-in-quotes-notice-not-to-spli
+// except:
+// - Treat the query ["san fran] as having the single term ["san fran"] instead of ["san", "fran"]
+// - Removes the delimiting quotes.
+func matchString(str, query string) bool {
+	matches := matchGroups.FindAllStringSubmatch(query, -1)
```
willhug: This is returning  results in a different format when we use the quotes.  We have have three elements in the match instead of two.  The result seems to be that we start filtering on "" (which always has a match) so we won't get any filtering at all: https://play.golang.org/p/xlcQ73hpZth

graphql/schemabuilder/pagination.go
```diff
@@ -434,6 +506,34 @@ func indexOfPaginationArgs(argType reflect.Type) int {
...
+	typ := reflect.TypeOf(fn)
+	return typ.Out(0)
+}
+
+func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ reflect.Type) error {
+	c.FilterTextFields = make(map[string]*graphql.Field)
+
+	for name, fn := range m.TextFilterFuncs {
+		funcTyp := getFuncReturnValue(fn)
+		if funcTyp != reflect.TypeOf("") {
```
willhug: can we make a `var typeOfString = reflect.TypeOf("")`

or just use the "Kind" of the funcTyp?

graphql/connection_test.go
```diff
@@ -65,31 +65,20 @@ func TestConnection(t *testing.T) {
...
-		retList[2] = Item{Id: 3}
-		retList[3] = Item{Id: 4}
-		retList[4] = Item{Id: 5}
-		return retList
+		return []Item{{1}, {2}, {3}, {4}, {5}}
 	}, schemabuilder.Paginated)
+	inner.FieldFunc("innerConnectionWithFilter", func() []Item {
+		return []Item{{1}, {2}, {3}, {4}, {5}}
+	}, schemabuilder.Paginated, schemabuilder.TextFilterFields{
+		"foo": func(ctx context.Context, i Item) string {
```
willhug: Do we have to define this for all filterable fields?  Or is there a way to point to a field on the item direclty?  Is this to allow filtering on multiple fields at the same time?

willhug: Also, it looks like "foo" is not actually used anywhere in the query.  Is that just the name of the field?  Does it have value?  Can we reference it from the client?

graphql/schemabuilder/pagination.go
```diff
@@ -372,6 +380,62 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
+			// Resolve the graphql.Field made for sorting.
+			text, err := filterField.Resolve(context.TODO(), node, nil, nil)
+			if err != nil {
+				return nil, err
+			}
+
+			// Only strings are allowed for FilterText fields.
+			textString, ok := text.(string)
+			if !ok {
+				return nil, errors.New("Text was not a string")
```
willhug: include the name of the node here?  Or some other identifier to indicate that a filter on field "Blah" was not valid because it's not a string.

graphql/schemabuilder/pagination.go
willhug: I guess this is not on the request path, so nbd.

graphql/schemabuilder/pagination.go
willhug: Should probably add a test for this case as well.

graphql/connection_test.go
```diff
@@ -292,6 +305,107 @@ func TestConnection(t *testing.T) {
...
 		}
 	}`)
+
+	snap.SnapshotQuery("Pagination, sorts", `{
+		inner {
+			numbersAsc: innerConnectionWithSort(sortBy: "numbers", sortOrder: "asc", first: 5, after: "") {
+				totalCount
+				edges {
+					node {
+						id
```
willhug: Same thing about being confusing to read the snapshots, can we add the fields we're sorting by here.

graphql/connection_test.go
```diff
@@ -74,6 +74,19 @@ func TestConnection(t *testing.T) {
 			return []string{"can", "man", "cannot", "soban", "socan"}[i.Id-1]
 		},
 	})
+	inner.FieldFunc("innerConnectionWithSort", func() []Item {
+		return []Item{{1}, {2}, {3}, {4}, {5}}
+	}, schemabuilder.Paginated, schemabuilder.SortFields{
+		"numbers": func(ctx context.Context, i Item) int64 {
```
willhug: Why do these need to be extra fields?  Can we build sorting on top of the existing field names?  Maybe just a whitelist of the "sortable" fields.

doc/pagination.md
```diff
@@ -0,0 +1,249 @@
...
+
+```go
+schema := schemabuilder.NewSchema()
+object := schema.Object("user", User{})
+object.FieldFunc("addresses", func(ctx context.Context, user *User, args struct {}) ([]*Address, error) {
+  return GetAddresses(ctx, user, args)
+}, schemabuilder.Paginated,
+// Each field in this map will be checked when sorting:
+schemabuilder.SortFields{
+	"houseNumber": func(ctx context.Context, a *Address) int64 {
```
willhug: why would we use houseNumber here instead of the "number" field that will be returned on the address node?  It seems a bit odd that we need to do the mapping between the two field names.

doc/pagination.md
```diff
@@ -0,0 +1,249 @@
...
+		return a.Number
+	},
+	"userName": func(ctx context.Context, a *Address) (string, error) {
```
willhug: would this already be a fieldfunc on the Address object?

graphql/connection_test.go
berfarah: Ok, this is admittedly not great API right now - my thoughts here are that we want to add `filterBy` (where only certain fields are filtered) and this allows for that. Other than that, there's currently no use / no way the client can reference it. Currently the default would be to filter using all defined fields.

graphql/schemabuilder/pagination.go
berfarah: That's fair - I can provide the name of the text filter. It can be something like:
```go
"filter %s returned %T, must be a string", name, text
```

graphql/connection_test.go
berfarah: For the snapshot? I've named the parent node for what we're sorting. Unfortunately, since these are virtual values, they're not super deuggable. If we were sorting actual fields that are available for querying, it would be easier. I think this is another argument to not add additional virtual fields here, and opt instead for a shorthand to existing sub-fields.

graphql/connection_test.go
berfarah: Yes, but that's phase 3 of the plan.

berfarah: Maybe it's worth taking the time to do that now, given this work has taken less time than anticipated -- cc @stephen

doc/pagination.md
berfarah: I honestly was just trying to come up with a loosely tied (somewhat contrived) example. Would love to swap it out for a better example.

doc/pagination.md
berfarah: We don't get access to the FieldFunc with the current API.

graphql/connection_test.go
```diff
@@ -28,6 +30,27 @@ type Item struct {
 	Id int64
 }
 
+type Snapshotter struct {
```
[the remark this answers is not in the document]
berfarah: Makes sense, I can move this out into our testing package - I think we have one

graphql/connection_test.go
```diff
@@ -266,6 +255,43 @@ func TestConnection(t *testing.T) {
...
 		}
 	}`)
+
+	snap.SnapshotQuery("Pagination, filter", `{
+		inner {
+			filterByCan: innerConnectionWithFilter(filter: "can", first: 5, after: "") {
+				totalCount
+				edges {
+					node {
+						id
```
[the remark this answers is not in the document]
berfarah: Unfortunately it's not queriable :|

graphql/schemabuilder/pagination.go
berfarah: Agreed - I think this is worth extracting into a separate internal package where it can be tested.

graphql/schemabuilder/pagination.go
berfarah: done

graphql/connection_test.go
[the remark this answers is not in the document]
berfarah: (lmk if this is resolved)

graphql/schemabuilder/pagination.go
```diff
@@ -372,6 +380,62 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
+		return nodes, nil
+	}
+
+	var filteredNodes []interface{}
+	for _, node := range nodes {
+		keep := false
+		// For each possible field we're matching against, check if there's a match.
+		for _, filterField := range c.FilterTextFields {
+			// Resolve the graphql.Field made for sorting.
+			text, err := filterField.Resolve(context.TODO(), node, nil, nil)
```
stephen: let's figure out how to pipe the context through `getConnection()` so that this shares a cache with real resolvers

graphql/schemabuilder/pagination.go
```diff
@@ -434,6 +506,34 @@ func indexOfPaginationArgs(argType reflect.Type) int {
 	return -1
 }
 
+func getFuncReturnValue(fn interface{}) reflect.Type {
```
stephen: nit: `getFnReturnType` vs `Value`?

graphql/schemabuilder/pagination.go
```diff
@@ -434,6 +506,34 @@ func indexOfPaginationArgs(argType reflect.Type) int {
...
+	return typ.Out(0)
+}
+
+func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ reflect.Type) error {
+	c.FilterTextFields = make(map[string]*graphql.Field)
+
+	for name, fn := range m.TextFilterFuncs {
+		funcTyp := getFuncReturnValue(fn)
+		if funcTyp != reflect.TypeOf("") {
+			return fmt.Errorf("%s: Sort field must be a string", name)
```
stephen: nit: lower lower lower 😛

graphql/schemabuilder/pagination.go
```diff
@@ -434,6 +506,34 @@ func indexOfPaginationArgs(argType reflect.Type) int {
...
+		}
+
+		// Build a GraphQL field for the function.
+		field, err := sb.buildFunction(typ, &method{MarkedNonNullable: true, Fn: fn})
```
stephen: `MarkedNonNullable` unnecessary i think, since we already checked that the return is a string and not a pointer.

graphql/schemabuilder/pagination.go
```diff
@@ -434,6 +506,34 @@ func indexOfPaginationArgs(argType reflect.Type) int {
...
+		if err != nil {
+			return err
+		}
+		if field.Args != nil && len(field.Args) > 0 {
+			return fmt.Errorf("%s: Text filter fields can't take arguments", name)
```
stephen: can we also add tests for these bad registrations?

graphql/schemabuilder/pagination.go
```diff
@@ -99,6 +100,10 @@ type ConnectionArgs struct {
 	Args interface{}
 	// filter: "text search"
 	Filter *string
+	// sortBy: "fieldName"
+	SortBy *string
```
stephen: can we use an enum for these based on the keys in the map?

graphql/schemabuilder/pagination.go
```diff
@@ -436,6 +447,51 @@ func (c *connectionContext) applyTextFilter(nodes []interface{}, args Pagination
...
+	if !ok {
+		return nil, fmt.Errorf("Unknown sort field %s", *args.SortBy)
+	}
+
+	// sortValues is the slice we'll be sorting (with the sorted values) in order to figure out
+	// node order.
+	sortValues := make([]sortReference, len(nodes))
+	for i, node := range nodes {
+		// Resolve the graphql.Field made for sorting.
+		sortValue, err := sortField.Resolve(context.TODO(), node, nil, nil)
```
stephen: same note here: let's figure out how to share a context for the computation cache

graphql/schemabuilder/pagination.go
```diff
@@ -380,6 +444,14 @@ func (c *connectionContext) getConnection(out []reflect.Value, args PaginationAr
 		return Connection{}, nil
 	}
 
+	if !c.IsExternallyManaged() {
+		var err error
+		nodes, err = c.applyTextFilter(nodes, args)
+		if err != nil {
+			return Connection{}, nil
```
stephen: why do we drop the error here?

graphql/schemabuilder/pagination.go
```diff
@@ -380,6 +444,14 @@ func (c *connectionContext) getConnection(out []reflect.Value, args PaginationAr
 		return Connection{}, nil
 	}
 
+	if !c.IsExternallyManaged() {
+		var err error
```
stephen: why not `nodes, err := ` below?

graphql/schemabuilder/pagination.go
```diff
@@ -450,6 +506,10 @@ func (c *connectionContext) getConnection(out []reflect.Value, args PaginationAr
 		if err != nil {
 			return Connection{}, nil
 		}
+		nodes, err = c.applySort(nodes, args)
+		if err != nil {
+			return Connection{}, nil
```
stephen: same question: why drop the error?

graphql/schemabuilder/pagination.go
```diff
@@ -511,6 +571,117 @@ func getFuncReturnValue(fn interface{}) reflect.Type {
...
+		return sorts[reflect.Int64]
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return sorts[reflect.Uint64]
+	case reflect.Float32, reflect.Float64:
+		return sorts[reflect.Float64]
+	case reflect.String:
+		return sorts[reflect.String]
+	default:
+		// This should literally be impossible.
+		panic(fmt.Sprintf("Unexpected type %v used in superSmartSort", typ))
```
stephen: `getSort`? 😛

graphql/schemabuilder/pagination.go
```diff
@@ -511,6 +571,117 @@ func getFuncReturnValue(fn interface{}) reflect.Type {
...
+	}
+}
+
+var sorts = map[reflect.Kind]func([]sortReference, SortOrder){
+	reflect.Int64: func(slice []sortReference, order SortOrder) {
+		sort.Slice(slice, func(i, j int) bool {
```
stephen: should we slice stable these?

graphql/schemabuilder/pagination.go
```diff
@@ -511,6 +571,117 @@ func getFuncReturnValue(fn interface{}) reflect.Type {
...
+			a := reflect.ValueOf(slice[i].value)
+			b := reflect.ValueOf(slice[j].value)
+			if order == SortOrder_Ascending {
+				return a.Int() > b.Int()
```
stephen: should this be flipped? (`<`)

graphql/schemabuilder/pagination.go
```diff
@@ -511,6 +571,117 @@ func getFuncReturnValue(fn interface{}) reflect.Type {
...
+}
+
+func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ reflect.Type) error {
+	c.SortFunctions = make(map[string]func([]sortReference, SortOrder))
+	c.SortFields = make(map[string]*graphql.Field)
+
+	for name, fn := range m.SortFuncs {
+		sortableTyp := getFuncReturnValue(fn)
+		if !supportedSort(sortableTyp) {
+			return errors.New("Sort must be supported (err) - probably move this err into supportedsort")
```
stephen: is this error intended?

graphql/schemabuilder/pagination.go
```diff
@@ -109,7 +114,9 @@ type PaginationArgs struct {
 	After  *string
 	Before *string
 
-	Filter *string
+	Filter    *string
```
stephen: should we put these in a different embeddable struct? i.e. if you want pagination you embed `PaginationArgs`, `SortArgs`, etc.

graphql/schemabuilder/pagination.go
stephen: maybe enum makes the embedding thing hard

graphql/schemabuilder/pagination.go
stephen: oh great please ignore my earlier comments

graphql/schemabuilder/pagination.go
stephen: (same thing for sorts)

graphql/connection_test.go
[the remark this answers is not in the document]
berfarah: Done

graphql/schemabuilder/pagination.go
berfarah: Done

graphql/schemabuilder/pagination.go
```diff
@@ -104,6 +108,8 @@ type PaginationArgs struct {
 	Last   *int64
 	After  *string
 	Before *string
+
+	Filter *string
```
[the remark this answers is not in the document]
berfarah: 👍

graphql/schemabuilder/pagination.go
stephen: (my preference comes from https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

graphql/schemabuilder/pagination.go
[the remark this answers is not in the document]
berfarah: This is done

graphql/schemabuilder/pagination.go
berfarah: This is done in a later commit

graphql/schemabuilder/pagination.go
berfarah: Done

graphql/schemabuilder/pagination.go
berfarah: Done (I think)

graphql/schemabuilder/pagination.go
berfarah: Done

graphql/schemabuilder/pagination.go
berfarah: Ehhhhhhhhhhhhhh we can ~dynamically~ do it maybe

graphql/schemabuilder/pagination.go
berfarah: Whoops

graphql/schemabuilder/pagination.go
berfarah: This is done in another commit

graphql/schemabuilder/pagination.go
berfarah: We're overwriting the higher scoped nodes. I'm pretty sure that's how it works.

graphql/schemabuilder/pagination.go
berfarah: Whoops

graphql/schemabuilder/pagination.go
berfarah: Done

graphql/schemabuilder/pagination.go
berfarah: Done

graphql/schemabuilder/pagination.go
berfarah: I don't know anymore

graphql/schemabuilder/pagination.go
berfarah: Addressed

graphql/schemabuilder/pagination.go
berfarah: I'm not sure I like that

graphql/schemabuilder/pagination.go
berfarah: Or rather, this is a useful error now

graphql/schemabuilder/pagination.go
berfarah: Absolutely, yes

graphql/schemabuilder/pagination.go
berfarah: ^ this is done btw :D

graphql/schemabuilder/pagination.go
berfarah: Maybe this could be a phase 2?

graphql/connection_test.go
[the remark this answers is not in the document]
willhug: guess we're tackling this in one of the next steps for the pagination stuff, sounds good to me.

graphql/connection_test.go
willhug: hrm, maybe we can add a "dummy" node field that is the value of the sortable/filterable field (which would be the exact same as the field we're sorting/filtering by) just to make reading the snapshots easier.

graphql/schemabuilder/pagination.go
```diff
@@ -417,6 +428,51 @@ func (c *connectionContext) applyTextFilter(nodes []interface{}, args Pagination
...
+	// Default to ascending sort.
+	sortOrder := SortOrder_Ascending
+	if args.SortOrder != nil {
+		sortOrder = *args.SortOrder
+	}
+
+	sortField, ok := c.SortFields[*args.SortBy]
+	// If the field wasn't registered, it's an unknown sort field.
+	if !ok {
+		return nil, fmt.Errorf("Unknown sort field %s", *args.SortBy)
```
willhug: caps

graphql/schemabuilder/pagination.go
```diff
@@ -492,6 +552,121 @@ func getFuncReturnType(fn interface{}) reflect.Type {
...
+type SortOrder int64
+
+const (
+	SortOrder_Ascending SortOrder = iota
+	SortOrder_Descending
+)
+
+type sortReference struct {
+	index int
+	value interface{}
```
willhug: can we put the `reflect.Value` directly here?

doc/pagination.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+```graphql
+{
+  user(id: 5) {
+    addresses(
+      # Arguments conforming to the Relay spec.
+      first: 2,
+      after: "",
+			# Also available:
```
willhug: indentation issues.

doc/pagination.md
```diff
@@ -0,0 +1,249 @@
...
+```
+
+### Sorting
+
+You can define sorts for a field in Thunder. The return value of the sort must
+be one of the following (aliases are supported):
+
+- `int`/`int8`/`int16`/`int32`/`int64`
+- `uint`/`uint8`/`uint16`/`uint32`/`uint64`
+- `string`
```
willhug: floats?

doc/pagination.md
```diff
@@ -0,0 +1,250 @@
...
+  },
+  "state": func(ctx context.Context, a *Address) (string, error) {
+    state, err := GetState(ctx, a.StateId)
+    if err != nil {
+      return "", err
+    }
+    return state.Name, nil
+  },
+},
+)
```
willhug: some odd indentation here.

doc/pagination.md
berfarah: Done

doc/pagination.md
berfarah: Done

doc/pagination.md
berfarah: Done

graphql/schemabuilder/pagination.go
berfarah: Done

graphql/schemabuilder/pagination.go
berfarah: Fixed in last commit >.>

internal/filter/filter_test.go
```diff
@@ -0,0 +1,31 @@
...
+	testcases := []struct {
+		String  string
+		Query   string
+		Matches bool
+	}{
+		{"hi, san francisco", `"san fran"`, true},
+		{"hi, san francisco", `"san fran" and`, true},
+		{"hi, sandy francisco", `"san fran"`, false},
+		{"hi, sandy francisco", `"san fran" and`, true},
+		{"hi, sandy francisco", ``, false},
```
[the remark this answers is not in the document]
berfarah: Done

graphql/connection_test.go
```diff
@@ -73,336 +75,176 @@ func TestConnection(t *testing.T) {
...
-							id
-						}
-						cursor
-					}
-					pageInfo {
-						hasNextPage
-						hasPrevPage
-						startCursor
-						endCursor
+	snap := testgraphql.Snapshotter{
```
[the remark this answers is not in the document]
berfarah: Done

graphql/connection_test.go
```diff
@@ -73,336 +75,176 @@ func TestConnection(t *testing.T) {
...
-					}
-					pageInfo {
-						hasNextPage
-						hasPrevPage
-						startCursor
-						endCursor
+		}
+	}`)
+
+	snap.RecordError = true
```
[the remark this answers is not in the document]
berfarah: Done

graphql/connection_test.go
```diff
@@ -44,31 +44,20 @@ func TestConnection(t *testing.T) {
...
+	})
 	inner.FieldFunc("innerConnectionNilArg", func() []Item {
-		retList := make([]Item, 5)
-		retList[0] = Item{Id: 1}
-		retList[1] = Item{Id: 2}
-		retList[2] = Item{Id: 3}
-		retList[3] = Item{Id: 4}
-		retList[4] = Item{Id: 5}
-		return retList
+		return []Item{{1}, {2}, {3}, {4}, {5}}
```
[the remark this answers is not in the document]
berfarah: Done

graphql/connection_test.go
berfarah: I've added that

doc/pagination.md
berfarah: ^ lmk if you want me to address this

graphql/schemabuilder/pagination.go
stephen: i think this would be enough of a win (with autocompletion, validation) to do now. if the tradeoff is against embeddable (passthrough) sorts, I think I'd rather have the enum support over the passthrough api.

berfarah:
@stephen as per offline conversation, will punt on enums until later in a separate PR

PR #205, opened 2018-11-29
Bugfix: sort/filter fields now support pointer sources

berfarah:
When pointers are passed into buildFunction it gets things wrong.

When determining if the function is a ptr-func, it checks to see if the ptr representation of a type
matches what is passed in as an argument to the function.

Because we are passing in a pointer, that pointer representation ends up being a dual pointer.

Added an error so that we can't shoot ourselves in the foot this way again.

Also in this PR:
- Nodes on connections are marked as non-null

### DIFF
--- a/graphql/connection_test.go
+++ b/graphql/connection_test.go
@@ -62,6 +62,9 @@ func TestConnection(t *testing.T) {
 		"foo": func(ctx context.Context, i Item) string {
 			return i.FilterText
 		},
+		"bar": func(ctx context.Context, i *Item) string {
+			return ""
+		},
 	})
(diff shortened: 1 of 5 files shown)


graphql/connection_test.go
```diff
@@ -59,7 +59,7 @@ func TestConnection(t *testing.T) {
 			{Id: 5, FilterText: "socan"},
 		}
 	}, schemabuilder.Paginated, schemabuilder.TextFilterFields{
-		"foo": func(ctx context.Context, i Item) string {
```
stephen: let's have both cases

berfarah: Done!

bojdell:
nice 🙌

PR #206, opened 2018-11-30
go graphql: expose runMiddlewares

changpingc:
To use it in a custom HTTP endpoint.


changpingc:
We want to have more control over how we parse and execute query, and this helps with having a custom implementation of http endpoint.

PR #214, opened 2019-01-09
go graphql: Add ability to automatically TextMarshal/Unmarshal types

willhug:
Summary: Updates the GraphQL package so that if we encounter a type when
parsing the graph that implements the encoding.TextMarshaler/TextUnmarshaler 
interfaces we will automatically convert that type to and from a string in the 
graphql input and output types.

### DIFF
--- a/graphql/executor.go
+++ b/graphql/executor.go
@@ -90,6 +90,11 @@ func PrepareQuery(typ Type, selectionSet *SelectionSet) error {
 			return NewClientError("enum field must have no selections")
 		}
 		return nil
+	case *CustomScalar:
+		if selectionSet != nil {
+			return NewClientError("scalar field must have no selections")
+		}
+		return nil
 	case *Union:
 		if selectionSet == nil {
 			return NewClientError("object field must have selections")
@@ -373,6 +378,8 @@ func (e *Executor) execute(ctx context.Context, typ Type, source interface{}, se
 			return mapVal, nil
 		}
 		return nil, errors.New("enum is not valid")
+	case *CustomScalar:
+		return typ.Unwrapper(source)
 	case *Union:
 		return e.executeUnion(ctx, typ, source, selectionSet)
 	case *Object:
--- a/graphql/schemabuilder/input.go
+++ b/graphql/schemabuilder/input.go
@@ -1,6 +1,7 @@
 package schemabuilder
 
 import (
+	"encoding"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -171,6 +172,10 @@ func (sb *schemaBuilder) makeArgParserInner(typ reflect.Type) (*argParser, graph
 		return parser, argType, nil
 	}
 
+	if reflect.PtrTo(typ).Implements(textUnmarshalerType) {
+		return sb.makeTextUnmarshalerParser(typ)
+	}
+
 	switch typ.Kind() {
 	case reflect.Struct:
 		parser, argType, err := sb.makeStructParser(typ)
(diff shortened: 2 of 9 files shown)


graphql/schemabuilder/input.go
```diff
@@ -230,6 +235,26 @@ func (sb *schemaBuilder) getEnumArgParser(typ reflect.Type) (*argParser, graphql
...
+// value as a string and insert it into the destination type using the
+// encoding.TextUnmarshaler API.
+func (sb *schemaBuilder) makeTextUnmarshalerParser(typ reflect.Type) (*argParser, graphql.Type, error) {
+	return &argParser{
+		FromJSON: func(value interface{}, dest reflect.Value) error {
+			asString, ok := value.(string)
+			if !ok {
+				return errors.New("not a string")
+			}
+			unmarshalable, ok := dest.Addr().Interface().(encoding.TextUnmarshaler)
```
berfarah: Is this guaranteed to be addressable? Can this panic?

berfarah:
This looks great!

graphql/schemabuilder/input.go
willhug: Added a check that it's addressable (and a nicer error if it is not).

graphql/textmarshal_test.go
```diff
@@ -0,0 +1,169 @@
...
+	o.FieldFunc("uuidfunc", func(inner Inner) Uuid {
+		return inner.Uuid
+	})
+	o.FieldFunc("uuidslicefunc", func(inner Inner) []Uuid {
+		return inner.UuidSlice
+	})
+	o.FieldFunc("ptruuidslicefunc", func(inner Inner) []*Uuid {
+		return inner.PtrUuidSlice
+	})
+	o.FieldFunc("invaliduuidfunc", func(inner Inner) Uuid {
```
stephen: nit: camelCase each of these field funcs?

graphql/types.go
```diff
@@ -43,6 +43,18 @@ func (e *Enum) enumValues() []string {
 	return e.Values
 }
 
+// CustomScalar is a leaf value with a custom Unwrapper function.
+type CustomScalar struct {
+	TypeName  string
+	Unwrapper func(interface{}) (interface{}, error)
```
stephen: should we make this part of `Scalar`?

how does introspection handle this? can we add an introspection test case?

graphql/types.go
```diff
@@ -43,6 +43,18 @@ func (e *Enum) enumValues() []string {
 	return e.Values
 }
 
+// CustomScalar is a leaf value with a custom Unwrapper function.
+type CustomScalar struct {
```
stephen: if we go down this path, check that this implements `Type` (see line 106)

graphql/types.go
willhug: It was already a part of the introspection, but I think I agree that it would be nicer/safer to not have to change all those locations.

graphql/types.go
willhug: It does! Otherwise we wouldn't be able to return it from the getType func.

willhug:
Removed CustomScalar, Added Unwrapper to Scalar type and made function responses camelCase.

graphql/executor.go
```diff
@@ -366,6 +366,9 @@ func (e *Executor) execute(ctx context.Context, typ Type, source interface{}, se
 	}
 	switch typ := typ.(type) {
 	case *Scalar:
+		if typ.Unwrapper != nil {
```
berfarah: I think this commit should be further up the chain. Looks like the dates might be messed up on the commits?

willhug: it's the correct order in the stack, but github is stupid.

PR #215, opened 2019-01-10
go graphql: Add support for Default/ZeroValue parameter defaults

willhug:
Summary: Adds support for graphql parameter fields to be noop fields.
Previously if you wanted to add a non-pointer field as a new input arg
to one of your functions this would break all existing callers (since 
they would not have the proper field).  Now if you have an existing
input struct like:
```
type Input struct {
  Arg1 string
  Arg2 string
}
```

You can safely add a new type like so:
```
type Input struct {
  Arg1 string
  Arg2 string
  Arg3 string `graphql:",optional"`
}
```
And the Arg3 string will get filled with "" if no args are provided.

### DIFF
--- a/graphql/schemabuilder/reflect.go
+++ b/graphql/schemabuilder/reflect.go
@@ -49,6 +49,10 @@ type graphQLFieldInfo struct {
 
 	// KeyField indicates that this field should be treated as a Object Key field.
 	KeyField bool
+
+	// OptionalInputField indicates that this field should be treated as an optional
+	// field on graphQL input args.
+	OptionalInputField bool
 }
 
 // parseGraphQLFieldInfo parses a struct field and returns a struct with the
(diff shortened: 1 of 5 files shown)


berfarah:
Changelog entry and :shipit:

graphql/schemabuilder/reflect.go
```diff
@@ -49,6 +49,10 @@ type graphQLFieldInfo struct {
 
 	// KeyField indicates that this field should be treated as a Object Key field.
 	KeyField bool
+
+	// OptionalArg indicates that this field should be treated as an optional
+	// field on graphQL input args.
+	OptionalArg bool
```
stephen: nit: how about `OptionalInputField` to match the language graphql uses to refer to this

willhug: Done!

PR #218, opened 2019-01-23
go schemabuilder: Make Type annotation for "optional" fields optional.

willhug:
Summary: The type annotation that would be linked with an optional field
would not be "optional" meaning that new fields that would be added
would have issues getting parsed by the frontend.  This change modifies
the graphql schema view to have "optional" fields be truly optional
from graphql's perspective.

### DIFF
--- a/graphql/schemabuilder/input.go
+++ b/graphql/schemabuilder/input.go
@@ -132,7 +132,7 @@ func (sb *schemaBuilder) getStructObjectFields(typ reflect.Type) (*graphql.Input
 			return nil, nil, err
 		}
 		if fieldInfo.OptionalInputField {
-			parser = wrapWithZeroValue(parser)
+			parser, fieldArgTyp = wrapWithZeroValue(parser, fieldArgTyp)
 		}
 
 		fields[fieldInfo.Name] = argField{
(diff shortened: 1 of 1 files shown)


graphql/schemabuilder/input.go
```diff
@@ -133,6 +133,10 @@ func (sb *schemaBuilder) getStructObjectFields(typ reflect.Type) (*graphql.Input
 		}
 		if fieldInfo.OptionalInputField {
 			parser = wrapWithZeroValue(parser)
+			// Make sure the "fieldArgType" we expose in graphQL is a Nullable field.
+			if f, ok := fieldArgTyp.(*graphql.NonNull); ok {
```
willhug: Other option is to add the fieldArgTyp to the wrapWithZeroValue function as well.

graphql/schemabuilder/input.go
```diff
@@ -230,7 +234,7 @@ func wrapWithZeroValue(inner *argParser) *argParser {
 			return inner.FromJSON(value, dest)
 		},
 		Type: inner.Type,
-	}
+	}, fieldArgTyp
```
stephen: can we add an introspection test for this?

willhug: Done: https://github.com/samsarahq/thunder/pull/219

PR #225, opened 2019-02-23
Add support of setting proper Content-Type header, small optimisation

victor-perov:
I would like to add support of setting more relevant than `text/plain` Content-Type header for GraphQL. Regarding [specification](https://graphql.org/learn/serving-over-http/), response should has JSON format, so, the more suitable Content-Type is `application/json`.

This PR replaces using `http.Error` with `w.Write` for sending output to the client.

Why I think it is important.

Overriding Content-Type header inside `http.Error` is not only one thing, made me do this PR. In order to pass JSON to `http.Error`, you have to convert it to string. This is totally redundant operation for this case. Moreover, `http.Error` adds additional `\n` (because of `fmt.Fprintln()`) to each response of service, based on this framework. Though there is not any conventions in specifications of [HTTP/2](https://http2.github.io/http2-spec/) and [HTTP/1.1](https://tools.ietf.org/html/rfc2616#section-4.4) about setting `\n` in the end of message-body, but other huge GraphQL APIs, such as github, facebook and others do not set `\n` in the end of JSON response. I also do believe that `\n` in the end on JSON  response is useless and should be reworked.

This PR also removes default irrelevant for GraphQL setting `X-Content-Type-Options` header with `nosniff` value. However, you still able to set it (if needed) before starting processing `GraphQL.HTTPHandler`.

Because of `http.Write()` there is no need to send `http.StatusOk` with `w.WriteHeader()`. It will be don automatically.

In summary, after merging this PR, by default, response of GraphQL API based on this framework will use `Content-Type` header as `application/json`, will not has additional `\n` in the end of payload, will not has `X-Content-Type-Options` header, and will stop doing conversion bytes array to string on each response.

I would like to add an example, how still possible to set necessary for your application headers. For example, if you're using http router in the top of GraphQL API, you are able to do the next (sample with [go-chi](https://github.com/go-chi/chi) http router):

    ...
    r := chi.NewRouter()
    r.Group(func(router chi.Router) {
        router.Mount("/graphql", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
            w.Header().Set("X-Content-Type-Options", "nosniff")
            ...
            graphql.HTTPHandler(schema, func(input *graphql.ComputationInput, next graphql.MiddlewareNextFunc) *graphql.ComputationOutput {
                ...
            }).ServeHTTP(w, r)
        }))
    })
    ...

### DIFF
--- a/graphql/http.go
+++ b/graphql/http.go
@@ -47,8 +47,10 @@ func (h *httpHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
-
-		http.Error(w, string(responseJSON), http.StatusOK)
+		if w.Header().Get("Content-Type") == "" {
+			w.Header().Set("Content-Type", "application/json")
+		}
+		w.Write(responseJSON)
 	}
 
 	if r.Method != "POST" {
--- a/graphql/http_test.go
+++ b/graphql/http_test.go
@@ -41,7 +41,7 @@ func TestHTTPMustPost(t *testing.T) {
 		t.Errorf("expected 200, but received %d", rr.Code)
 	}
 
-	if diff := pretty.Compare(rr.Body.String(), "{\"data\":null,\"errors\":[\"request must be a POST\"]}\n"); diff != "" {
+	if diff := pretty.Compare(rr.Body.String(), "{\"data\":null,\"errors\":[\"request must be a POST\"]}"); diff != "" {
 		t.Errorf("expected response to match, but received %s", diff)
 	}
 }
@@ -58,7 +58,7 @@ func TestHTTPParseQuery(t *testing.T) {
 		t.Errorf("expected 200, but received %d", rr.Code)
 	}
 
-	if diff := pretty.Compare(rr.Body.String(), "{\"data\":null,\"errors\":[\"request must include a query\"]}\n"); diff != "" {
+	if diff := pretty.Compare(rr.Body.String(), "{\"data\":null,\"errors\":[\"request must include a query\"]}"); diff != "" {
 		t.Errorf("expected response to match, but received %s", diff)
 	}
 }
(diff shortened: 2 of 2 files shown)


victor-perov:
I have executed all of them as cURL from cli. `victor@` in the same line with response means response without trailing `\n`.

## github graphql api
```
< Server: Cowboy
< Date: Sat, 23 Feb 2019 14:51:33 GMT
< Connection: keep-alive
< Content-Type: application/json; charset=utf-8
< Cache-Control: max-age=0, private, must-revalidate
...
< Strict-Transport-Security: max-age=631138519
< X-Frame-Options: ALLOW-FROM https://developer.github.com
< X-Content-Type-Options: nosniff
< X-Xss-Protection: 1; mode=block
< X-Download-Options: noopen
< X-Permitted-Cross-Domain-Policies: none
< Content-Security-Policy: default-src 'none'; base-uri 'self'; block-all-mixed-content; connect-src 'self'; font-src 'self' data:; frame-ancestors developer.github.com; img-src 'self' data:; script-src 'self'; style-src 'self' 'unsafe-inline'
< Transfer-Encoding: chunked
< Via: 1.1 vegur
< 
* Connection #0 to host graphql-explorer.githubapp.com left intact
{"data":{"viewer":{"login":"victor-perov"}}}victor@
```

## facebook graph api
```
< HTTP/2 200 
< x-app-usage: {"call_count":5,"total_cputime":0,"total_time":0}
< content-type: application/json; charset=UTF-8
< facebook-api-version: v3.2
< strict-transport-security: max-age=15552000; preload
< pragma: no-cache
< x-fb-rev: 4791517
< access-control-allow-origin: *
< cache-control: private, no-cache, no-store, must-revalidate
...
< expires: Sat, 01 Jan 2000 00:00:00 GMT
< content-length: 47
< date: Sat, 23 Feb 2019 14:41:21 GMT
< 
* Connection #0 to host graph.facebook.com left intact
{"id":"<id is here>","name":"Viktor Perov"}victor@
```

## microsoft azure demo graph
```
< HTTP/1.1 200 OK
< Content-Length: 59
< Content-Type: application/json; charset=utf-8
...
< Server: Microsoft-IIS/10.0
< X-Powered-By: Express
< X-Powered-By: ASP.NET
< Date: Sat, 23 Feb 2019 15:44:14 GMT
< 
* Connection #0 to host graphql-demo.azurewebsites.net left intact
{"data":{"domains":[{"id":"M365x214355.onmicrosoft.com"}]}}victor@
```

victor-perov:
Hello, @stephen! Could you please take a look into my PR. Though we still can use our fork, but would be great to have it in upstream!

PR #230, opened 2019-03-29
btuan/count

btuan:
Implements a basic `Count` receiver that wraps `SELECT COUNT(*)` functionality in SQL databases.

### DIFF
--- a/sqlgen/db.go
+++ b/sqlgen/db.go
@@ -92,6 +92,23 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
 	return db
 }
 
+func (db *DB) baseCount(ctx context.Context, query *baseCountQuery) (int64, error) {
+	countQuery, err := query.makeCountQuery()
+	if err != nil {
+		return 0, err
+	}
+
+	clause, args := countQuery.ToSQL()
+
+	var count int64
+	err = db.QueryExecer(ctx).QueryRowContext(ctx, clause, args...).Scan(&count)
+	if err != nil {
+		return 0, err
+	}
+
+	return count, err
+}
+
 func (db *DB) BaseQuery(ctx context.Context, query *BaseSelectQuery) ([]interface{}, error) {
 	if query.Options == nil && !db.HasTx(ctx) && batch.HasBatching(ctx) {
 		rows, err := db.batchFetch.Invoke(ctx, query)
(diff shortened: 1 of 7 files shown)


sqlgen/db.go
```diff
@@ -123,6 +140,29 @@ func (db *DB) execWithTrace(ctx context.Context, query SQLQuery, operationName s
...
 
+// Count counts the number of relevant rows in a database, matching options in filter
+//
+// model should be a pointer to a struct, for example:
+//
+//   var res int64
+//   user := &User{Id; 10}
+//   if err := db.Count(ctx, user, &res, Filter{}); err != nil {
+//
+func (db *DB) Count(ctx context.Context, model interface{}, result *int64, filter Filter) error {
```
ryanr23: Drive by: I think there's no reason to use the pattern of a result pointer argument since the return type is fixed.  I'd return (int64, error).  It's a less awkward API.

sqlgen/db.go
```diff
@@ -92,6 +92,23 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
 	return db
 }
 
+func (db *DB) BaseCount(ctx context.Context, query *BaseCountQuery) (*int64, error) {
```
ryanr23: Ryan's opinion.  Return (int64, error).  Removes pointer handling from the non-error case.

sqlgen/db.go
btuan: totally agreed, I want to get some opinions from other maintainers, as all of the other API's return singleton error

btuan: @jellevandenhooff similarly prefers the less clunky direct return!

sqlgen/db.go
```diff
@@ -92,6 +92,23 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
 	return db
 }
 
+func (db *DB) BaseCount(ctx context.Context, query *BaseCountQuery) (int64, error) {
```
berfarah: Is there any reason we want this to be public API?

berfarah: +godoc

berfarah:
_hype_

sqlgen/db.go
btuan: unexported.

sqlgen/reflect.go
```diff
@@ -369,6 +369,55 @@ func makeWhere(table *Table, filter Filter) (*SimpleWhere, error) {
 	}, nil
 }
 
+type BaseCountQuery struct {
```
berfarah: Do we have to export anything we added here?

btuan: I don't believe so -- all of the other *basequeries are exported, but perhaps that's undesirable in hindsight.

btuan:
I plan on merging this Monday, barring any additional comments!

PR #231, opened 2019-04-15
[WIP] go graphql: Add basic BatchFieldFunc entrypoint for graphql objects

willhug:
Summary: Adds a BatchFieldFunc entry point for the Schemabuilder.
BatchFieldFuncs take a map[int]*ParentObject parameter as their "source"
and will return an equivalent response of map[int]*RespObject.  This allows
us to execute queries that would usually require a large amount of goroutines 
in a single function.

This version is a "Dummy" version of the functionality.  The Resolver we return
runs the BatchFunc with a single value, but all the code around it is built
to support passing in a slice of requests at the same time
(and returning a slice of results).

Also fixed an issue where regular SelectionSets would never set the value in
fieldFunc execution.

TODO:
- [ ] [maybe] consolidate logic between funcContext and BatchFuncContext
- [ ] Get some tests that test batching multiple inputs together.
- [ ] Add Changelog entry.
- [ ] \<Please comment for more\>

### DIFF
--- a/graphql/batch_test.go
+++ b/graphql/batch_test.go
@@ -1,0 +1,244 @@
+package graphql_test
+
+import (
+	"context"
+	"errors"
+	"reflect"
+	"testing"
+
+	"github.com/samsarahq/thunder/graphql"
+	"github.com/samsarahq/thunder/graphql/schemabuilder"
+	"github.com/samsarahq/thunder/internal"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBatchFieldFuncExecution(t *testing.T) {
+	type Object struct {
+		Key string
+	}
+	tests := []struct {
+		Name           string
+		GiveObjectFunc interface{}
+		GiveValueFunc  interface{}
+		GiveQuery      string
+		WantResultJSON string
+		WantError      string
+	}{
+		{
+			Name:           "good run with no value",
+			GiveObjectFunc: func(ctx context.Context) []*Object { return []*Object{&Object{Key: "key1"}} },
+			GiveValueFunc:  func(ctx context.Context, o map[int]Object) (map[int]string, error) { return nil, nil },
+			GiveQuery: `
+			{
+				objects {
+					key
+					value
+				}
+			}`,
+			WantResultJSON: `
+			{"objects": [
+			{"key": "key1", "value":null}
+			]}
+			`,
+		},
+		{
+			Name:           "good run with response",
+			GiveObjectFunc: func(ctx context.Context) []*Object { return []*Object{&Object{Key: "key1"}} },
+			GiveValueFunc: func(ctx context.Context, o map[int]Object) (map[int]string, error) {
+				myMap := make(map[int]string, len(o))
+				for idx, val := range o {
+					myMap[idx] = "valfor" + val.Key
+				}
+				return myMap, nil
+			},
+			GiveQuery: `
+			{
+				objects {
+					key
+					value
+				}
+			}`,
+			WantResultJSON: `
+			{"objects": [
+			{"key": "key1", "value": "valforkey1"}
+			]}
+			`,
+		},
+		{
+			Name:           "error run with response",
+			GiveObjectFunc: func(ctx context.Context) []*Object { return []*Object{&Object{Key: "key1"}} },
+			GiveValueFunc: func(ctx context.Context, o map[int]Object) (map[int]string, error) {
+				return nil, errors.New("my error here")
+			},
+			GiveQuery: `
+			{
+				objects {
+					key
+					value
+				}
+			}`,
+			WantError: "my error here",
+		},
+		{
+			Name:           "good run with pointer args",
+			GiveObjectFunc: func(ctx context.Context) []*Object { return []*Object{&Object{Key: "key1"}} },
+			GiveValueFunc: func(ctx context.Context, o map[int]*Object) (map[int]string, error) {
+				myMap := make(map[int]string, len(o))
+				for idx, val := range o {
+					myMap[idx] = "valfor" + val.Key
+				}
+				return myMap, nil
+			},
+			GiveQuery: `
+			{
+				objects {
+					key
+					value
+				}
+			}`,
+			WantResultJSON: `
+			{"objects": [
+			{"key": "key1", "value": "valforkey1"}
+			]}
+			`,
+		},
+		{
+			Name:           "good run with pointer response type",
+			GiveObjectFunc: func(ctx context.Context) []*Object { return []*Object{&Object{Key: "key1"}} },
+			GiveValueFunc: func(ctx context.Context, o map[int]*Object) (map[int]*string, error) {
+				myMap := make(map[int]*string, len(o))
+				for idx, val := range o {
+					p := "valfor" + val.Key
+					myMap[idx] = &p
+				}
+				return myMap, nil
+			},
+			GiveQuery: `
+			{
+				objects {
+					key
+					value
+				}
+			}`,
+			WantResultJSON: `
+			{"objects": [
+			{"key": "key1", "value": "valforkey1"}
+			]}
+			`,
+		},
+		{
+			Name:           "good run with pointer nil response type",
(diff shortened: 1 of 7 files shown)


willhug:
Added a batch.Func hookup in https://github.com/samsarahq/thunder/pull/233 that I don't plan to land (but can use as a POC for APIs).

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+		},
+		Args:           args,
+		Type:           retType,
+		ParseArguments: argParser.Parse,
+		Expensive:      funcCtx.hasContext,
+	}, nil
+}
+
+// funcContext is used to parse the function signature in buildFunction.
+type batchFuncContext struct {
```
jellevandenhooff: I find this context type confusing... I'll leave some comments on the methods.

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+	hasError        bool
+
+	funcType     reflect.Type
+	batchMapType reflect.Type
+	isPtrFunc    bool
+	parentTyp    reflect.Type
+}
+
+// getFuncVal returns a reflect.Value of an executable function.
+func (funcCtx *batchFuncContext) getFuncVal(m *method) (reflect.Value, error) {
```
jellevandenhooff: I find it confusing that this function both extracts something out of m and sticks on the funcCtx, and then returns something else.

I would inline functions this short in the caller.

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+	if fun.Kind() != reflect.Func {
+		return fun, fmt.Errorf("fun must be func, not %s", fun)
+	}
+	funcCtx.funcType = fun.Type()
+	return fun, nil
+}
+
+// getFuncInputTypes returns the input arguments for the function we're
+// representing.
+func (funcCtx *batchFuncContext) getFuncInputTypes() []reflect.Type {
```
jellevandenhooff: I would make this `func getFuncInputTypes(reflect.Type) []reflect.Type` and then invoke `getFuncInputTypes(funcCtx.funcType)`

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+	for i := 0; i < funcCtx.funcType.NumIn(); i++ {
+		in = append(in, funcCtx.funcType.In(i))
+	}
+	return in
+}
+
+// consumeContext reads in the input parameters for the provided
+// function and determines whether the function has a Context input parameter.
+// It returns the input types without the context parameter if it was there.
+func (funcCtx *batchFuncContext) consumeContext(in []reflect.Type) []reflect.Type {
```
jellevandenhooff: I think I'd either combine all the consume function into one, or add some other indicator they are grouped together. An intro comment for them might "// consume... figures out which of the optional arguments func implements". Then maybe I'd rename them "detect{ContextArg,SourceBatchArg,...}" to make the comment almost redundant.

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+	if len(in) > 0 && in[0] == contextType {
+		funcCtx.hasContext = true
```
jellevandenhooff: I'd make this function return (hasContext, in[1:]) so that the caller is forced to store hasContext. i like functions that don't modify state, if possible.

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+func (funcCtx *batchFuncContext) consumeSelectionSet(in []reflect.Type) []reflect.Type {
+	if len(in) > 0 && in[0] == selectionSetType {
+		in = in[1:]
+		funcCtx.hasSelectionSet = true
+	}
+	return in
+}
+
+func (funcCtx *batchFuncContext) getFuncOutputTypes() []reflect.Type {
+	out := make([]reflect.Type, 0, funcCtx.funcType.NumOut())
```
jellevandenhooff: this can be a shared helper

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+
+	batch := reflect.MakeMapWithSize(funcCtx.batchMapType, len(sources))
+	for idx, source := range sources {
+		idxVal := idx
+		sourceValue := reflect.ValueOf(source)
+		ptrSource := sourceValue.Kind() == reflect.Ptr
+		switch {
+		case ptrSource && !funcCtx.isPtrFunc:
+			batch.SetMapIndex(reflect.ValueOf(idxVal), sourceValue.Elem())
+		case !ptrSource && funcCtx.isPtrFunc:
```
jellevandenhooff: wonder if we should enforce something here to prevent unnecessary copying

graphql/schemabuilder/batch.go
jellevandenhooff: maybe call this batchFuncConfiguration? batchFuncShape?

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,300 @@
...
+	if len(out) > 0 && out[0] == errType {
+		funcCtx.hasError = true
+		out = out[1:]
+	}
+	return out
+}
+
+// prepareResolveArgs converts the provided source, args and context into the
+// required list of reflect.Value types that the function needs to be called.
+func (funcCtx *batchFuncContext) prepareResolveArgs(sources []interface{}, args interface{}, ctx context.Context, selectionSet *graphql.SelectionSet) []reflect.Value {
```
jellevandenhooff: i'd make this a function instead of a method because i read context as descriptor now and i'd use a descriptor but not ask it to do things... if that makes sense

graphql/schemabuilder/reflect_test.go
```diff
@@ -677,3 +677,122 @@ func TestObjectKeyMustBeScalar(t *testing.T) {
 	})
 
 }
+
+func TestBatchFieldFuncValidation(t *testing.T) {
```
jellevandenhooff: fantastic

graphql/batch_test.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestBatchFieldFuncExecution(t *testing.T) {
+	type Object struct {
+		Key string
+	}
+	tests := []struct {
+		Name           string
+		GiveObjectFunc interface{}
```
jellevandenhooff: i'd skip the give/want prefixes but up to you

graphql/batch_test.go
```diff
@@ -0,0 +1,244 @@
...
+		GiveValueFunc  interface{}
+		GiveQuery      string
+		WantResultJSON string
+		WantError      string
+	}{
+		{
+			Name:           "good run with no value",
+			GiveObjectFunc: func(ctx context.Context) []*Object { return []*Object{&Object{Key: "key1"}} },
+			GiveValueFunc:  func(ctx context.Context, o map[int]Object) (map[int]string, error) { return nil, nil },
```
jellevandenhooff: put some newlines please!

graphql/batch_test.go
```diff
@@ -0,0 +1,244 @@
...
+				}
+			}`,
+			WantResultJSON: `
+			{"objects": [
+			{"key": "key1", "value": {"key": "testkey1"}}
+			]}
+			`,
+		},
+	}
+	for _, tt := range tests {
```
jellevandenhooff: great

graphql/schemabuilder/batch.go
willhug: Yeah, this follows the pattern that has been used for the regular fieldFunc "funcCtx" in https://github.com/samsarahq/thunder/blob/master/graphql/schemabuilder/function.go#L78.  This file is mostly a copy+cleanup+adding batch functionality of that file.
Very happy to do a full refactor now that it's functional.

willhug: Would be great to make all the fieldFuncs (regular, paginated, batch) all use the same flexible infrastructure for defining input/output parameters dynamically.

PR #232, opened 2019-04-15
go schemabuilder/union_test: union builder doesn't err out on scalar …

shahryart:
…pointer members.

Shouldn't thunder complain on having non struct members in a union type? The error only shows up on frontend.


willhug:
Closing until we actually want to implement it.

PR #233, opened 2019-04-16
[WIP/POC] go batchFieldFunc: Allow running batch functions using batch.Func

willhug:
Summary: Stopgap/POC commit to prove that we can run batch functions
using batch.Func.  This is mostly a POC/cool thing we can use for
testing until we rewrite the executor to support batchFieldFuncs without
goroutines (I don't plan to land this).


willhug:
Overridden by #240

PR #235, opened 2019-04-19
go reactive: remove mutex in reactive.Cache

changpingc:
`reactive.Cache` call serializes callers with a mutex. If the computation function is written badly and blocks after context cancellation, we can have many orphan goroutines unnecessarily blocking on the mutex. This makes certain class of failures worse. 

Because this gets rid of the locker struct, it also helps with graphql.Executor rewrite for batch field func where it needs to do atomically update a set of cache keys.

### DIFF
--- a/reactive/rerunner.go
+++ b/reactive/rerunner.go
@@ -17,88 +17,37 @@ var (
 	WriteThenReadDelay = 200 * time.Millisecond
 )
 
-// locker is a collection of mutexes indexed by arbitrary keys
-type locker struct {
-	mu sync.Mutex
-	m  map[interface{}]*lock
-}
-
-// newLocker creates a new locker instance.
-func newLocker() *locker {
-	return &locker{
-		m: make(map[interface{}]*lock),
-	}
-}
-
-// lock is a single mutex in a locker
-type lock struct {
-	ref int
-	mu  sync.Mutex
-}
-
-// Lock locks a locker by (optionally) allocating, increasing the ref count,
-// and locking
-func (l *locker) Lock(k interface{}) {
-	l.mu.Lock()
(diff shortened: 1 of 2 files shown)


reactive/rerunner.go
```diff
@@ -225,24 +174,46 @@ func Cache(ctx context.Context, key interface{}, f ComputeFunc) (interface{}, er
...
-	computation := ctx.Value(computationKey{}).(*computation)
+	parent := ctx.Value(computationKey{}).(*computation)
 
-	cache.locker.Lock(key)
-	defer cache.locker.Unlock(key)
-
-	if child := cache.get(key); child != nil {
-		child.node.addOut(&computation.node)
-		return child.value, nil
+	cache.mu.Lock()
```
willhug: do we want to use a sync.Map or RWMutex here?  (Might be less contention when reading entries).

reactive/rerunner.go
```diff
@@ -225,24 +174,46 @@ func Cache(ctx context.Context, key interface{}, f ComputeFunc) (interface{}, er
...
+	if !ok {
+		// First cache call for this key.
+		entry = &cacheEntry{
+			done: make(chan struct{}),
+		}
+		cache.entries[key] = entry
+	}
+	cache.mu.Unlock()
+
+	if ok {
```
willhug: non-blocking

To reduce some of the logic here, should we invert the check here and use the same logic for lookup/return?
```
if !ok {
  entry.computation, entry.err = run(ctx, f) // close entry.Done in this func
}
select {
  case <-ctx.Done():
     ...
  case <-entry.done:
}
if entry.computation == nil {
  ...
}
if entry.err != nil {
 ...
}
//add out
return value
```

PR #236, opened 2019-04-22
go reactive: make reactive.Cache locker acquire take context

changpingc:
more conservative change than https://github.com/samsarahq/thunder/pull/235

### DIFF
--- a/reactive/ctxmutex.go
+++ b/reactive/ctxmutex.go
@@ -1,0 +1,38 @@
+package reactive
+
+import (
+	"context"
+	"sync"
+)
+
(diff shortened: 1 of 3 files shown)


willhug:
The repo used here does not inspire a lot of confidence: https://github.com/sasha-s/go-csync.  18 stars and no updates in the last 3 years... If we do this we should really make sure we're not causing a regression, check out some large org queries before and after (which are likely to have a lot of contention).

changpingc:
@willhug  @jellevandenhooff i rolled a simpler mutex implementation with a channel. ptal

PR #238, opened 2019-04-25
go graphql: Add BatchExecutor for controlled concurrency and Batching Capabilities

willhug:
Summary: Adds a new Executor option for Graphql execution to thunder.  This new
entrypoint has the ability to control the concurrency of requests through a 
"Scheduler" interface which owns how "ExecutionUnits" are executed on the graphql
graph.  In addition the new executor treats batches as a top-level concern, this 
makes it much easier to slot in graphQL batchFieldFuncs in a subsequent PR.

In the process of adding this new feature, there were a couple changes/fixes
that needed to go in to help with the migration:
- Fix passing SelectionSet into functions
- Change Object.Key from a Function to a fully-formed graphql.Field
- Add "External" flag on graphql.Field to control external concurrency.
- Clear reactive cache computations when a request errors.
- Fix Union type __typename fields so they are the type name of the sub-type (not the union)
- Fix a concurrency race condition that exists in pagination right now.

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -1,0 +1,396 @@
+package graphql
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"github.com/samsarahq/thunder/reactive"
+)
+
+// WorkUnit is a set of execution work that will be done when running
+// a graphql query.  For every source there is an equivalent destination
+// OutputNode that is used to record the result of running a section of the
+// graphql query.
+type WorkUnit struct {
+	ctx          context.Context
+	field        *Field
+	selection    *Selection
+	sources      []interface{}
+	destinations []*outputNode
+}
+
+// UnitResolver is a function that executes a function and returns a set of
+// new work units that need to be run.
+type UnitResolver func(*WorkUnit) []*WorkUnit
+
+// WorkScheduler is an interface that can be provided to the BatchExecutor
+// to control how we traverse the Execution graph.  Examples would include using
+// a bounded goroutine pool, or using unbounded goroutine generation for each
+// work unit.
+type WorkScheduler interface {
+	Run(resolver UnitResolver, startingUnits ...*WorkUnit)
+}
+
+func NewBatchExecutor(scheduler WorkScheduler) *BatchExecutor {
+	return &BatchExecutor{
+		scheduler: scheduler,
+	}
+}
+
+// BatchExecutor is a GraphQL executor.  Given a query it can run through the
+// execution of the request.
+type BatchExecutor struct {
+	scheduler WorkScheduler
+}
+
+// Execute executes a query by traversing the GraphQL query graph and resolving
+// or executing fields.  Any work that needs to be done is passed off to the
+// scheduler to handle managing concurrency of the request.
+// It must return a JSON marshallable response (or an error).
+func (e *BatchExecutor) Execute(ctx context.Context, typ Type, source interface{}, query *Query) (interface{}, error) {
+	queryObject, ok := typ.(*Object)
+	if !ok {
+		return nil, fmt.Errorf("expected query or mutation object for execution, got: %s", typ.String())
+	}
+
+	topLevelSelections := Flatten(query.SelectionSet)
+	topLevelRespWriter := newTopLevelOutputNode(query.Name)
+	initialSelectionWorkUnits := make([]*WorkUnit, 0, len(topLevelSelections))
+	writers := make(map[string]*outputNode)
+	for _, selection := range topLevelSelections {
+		field, ok := queryObject.Fields[selection.Name]
+		if !ok {
+			return nil, fmt.Errorf("invalid top-level selection %q", selection.Name)
+		}
+
+		writer := newOutputNode(topLevelRespWriter, selection.Alias)
+		writers[selection.Alias] = writer
+
+		initialSelectionWorkUnits = append(
+			initialSelectionWorkUnits,
+			&WorkUnit{
+				ctx:          ctx,
+				sources:      []interface{}{source},
+				field:        field,
+				destinations: []*outputNode{writer},
+				selection:    selection,
+			},
+		)
+	}
+
+	e.scheduler.Run(resolveWorkUnit, initialSelectionWorkUnits...)
+
+	if topLevelRespWriter.errRecorder.err != nil {
+		return nil, topLevelRespWriter.errRecorder.err
+	}
+	return writers, nil
+}
+
+// resolveWorkUnit executes/resolves a work unit and checks the
+// selections of the unit to determine if it needs to schedule more work (which
+// will be returned as new work units that will need to get scheduled.
+func resolveWorkUnit(unit *WorkUnit) []*WorkUnit {
+	// TODO Insert Batching code execution here if field is "Batch"
+
+	var units []*WorkUnit
+	for idx, src := range unit.sources {
+		if !unit.field.Expensive {
+			units = append(units, resolveNonBatchUnit(unit.ctx, src, unit.destinations[idx], unit)...)
+			continue
+		}
+		units = append(units, resolveNonBatchUnitWithCaching(src, unit.destinations[idx], unit)...)
+	}
+	return units
+}
+
+// resolveNonBatchUnitWithCaching wraps a resolve request in a reactive cache
+// call.
+// This function makes two assumptions:
+// - We assume that all the reactive cache will get cleared if there is an error.
+// - We assume that there is no "error-catching" mechanism that will stop an
+//   error from propagating all the way to the top of the request stack.
+func resolveNonBatchUnitWithCaching(src interface{}, dest *outputNode, unit *WorkUnit) []*WorkUnit {
+	var workUnits []*WorkUnit
+	subDestRes, err := reactive.Cache(unit.ctx, getWorkCacheKey(src, unit.field, unit.selection), func(ctx context.Context) (interface{}, error) {
+		subDest := newOutputNode(dest, "")
+		workUnits = resolveNonBatchUnit(ctx, src, subDest, unit)
+		return subDest.res, nil
+	})
+	if err != nil {
+		dest.Fail(err)
+	}
+	dest.Fill(subDestRes)
+	return workUnits
+}
+
+// getWorkCacheKey gets the work cache key for the provided source.
+func getWorkCacheKey(src interface{}, field *Field, selection *Selection) resolveAndExecuteCacheKey {
+	value := reflect.ValueOf(src)
+	// cache the body of resolve and execute so that if the source doesn't change, we
+	// don't need to recompute
+	key := resolveAndExecuteCacheKey{field: field, source: src, selection: selection}
+	// some types can't be put in a map; for those, use a always different value
+	// as source
+	if value.IsValid() && !value.Type().Comparable() {
+		// TODO: Warn, or somehow prevent using type-system?
+		key.source = new(byte)
+	}
+	return key
+}
+
+// resolveNonBatchUnit resolves a non-batch field in our graphql response graph.
+func resolveNonBatchUnit(ctx context.Context, src interface{}, dest *outputNode, unit *WorkUnit) []*WorkUnit {
+	fieldResult, err := safeResolve(ctx, unit.field, src, unit.selection.Args, unit.selection.SelectionSet)
+	if err != nil {
+		dest.Fail(err)
+		return nil
+	}
+	subFieldWorkUnits, err := resolveBatch(ctx, []interface{}{fieldResult}, unit.field.Type, unit.selection.SelectionSet, []*outputNode{dest})
+	if err != nil {
+		dest.Fail(err)
+		return nil
+	}
+	return subFieldWorkUnits
+}
+
+// resolveBatch traverses the provided sources and fills in result data and
+// returns new work units that are required to resolve the rest of the
+// query result.
+func resolveBatch(ctx context.Context, sources []interface{}, typ Type, selectionSet *SelectionSet, destinations []*outputNode) ([]*WorkUnit, error) {
+	switch typ := typ.(type) {
+	case *Scalar:
+		return nil, resolveScalarBatch(sources, typ, destinations)
+	case *Enum:
+		return nil, resolveEnumBatch(sources, typ, destinations)
+	case *List:
+		return resolveListBatch(ctx, sources, typ, selectionSet, destinations)
+	case *Union:
+		return resolveUnionBatch(ctx, sources, typ, selectionSet, destinations)
+	case *Object:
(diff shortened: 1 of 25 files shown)


graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+	queryObject, ok := typ.(*Object)
+	if !ok {
+		return nil, fmt.Errorf("expected query or mutation object for execution, got: %s", typ.String())
+	}
+
+	topLevelSelections := Flatten(query.SelectionSet)
+	topLevelRespWriter := newTopLevelObjectWriter(query.Name)
+	initialSelectionExecUnits := make([]*ExecutionUnit, 0, len(topLevelSelections))
+	writers := make(map[string]*objectWriter)
+	for _, selection := range topLevelSelections {
```
jellevandenhooff: could this use executeObject somehow?

graphql/executor_test.go
```diff
@@ -116,47 +121,47 @@ func makeQuery(onArgParse *func()) *Object {
...
+	"a": {
+		"value": 0,
+		"__key": 0,
+		"nested": {
+			"value": 1,
+			"__key": 1
+		}
+	},
+	"as": [
+					{"value": 0, "valuePtr": null, "__key": 0},
```
jellevandenhooff: wonky

graphql/schemabuilder/function.go
```diff
@@ -238,7 +238,7 @@ func (funcCtx *funcContext) argsTypeMap(argType graphql.Type) (map[string]graphq
 
 // prepareResolveArgs converts the provided source, args and context into the
 // required list of reflect.Value types that the function needs to be called.
-func (funcCtx *funcContext) prepareResolveArgs(source interface{}, args interface{}, ctx context.Context) []reflect.Value {
+func (funcCtx *funcContext) prepareResolveArgs(source interface{}, hasArgs bool, args interface{}, ctx context.Context, selectionSet *graphql.SelectionSet) []reflect.Value {
```
jellevandenhooff: confused about what should be an argument and what should live on funcCtx

graphql/schemabuilder/output.go
```diff
@@ -71,13 +71,13 @@ func (sb *schemaBuilder) buildStruct(typ reflect.Type) error {
...
 		if fieldInfo.KeyField {
-			if object.Key != nil {
+			if object.KeyField != nil {
 				return fmt.Errorf("bad type %s: multiple key fields", typ)
 			}
 			if !isScalarType(built.Type) {
 				return fmt.Errorf("bad type %s: key type must be scalar, got %T", typ, built.Type)
 			}
-			object.Key = built.Resolve
+			object.KeyField = built
```
jellevandenhooff: nice

graphql/schemabuilder/pagination.go
```diff
@@ -907,6 +907,7 @@ func (sb *schemaBuilder) buildPaginatedField(typ reflect.Type, m *method) (*grap
 		Type:           retType,
 		ParseArguments: argParser.Parse,
 		Expensive:      c.hasContext,
+		External:       true,
```
jellevandenhooff: maybe use Internal or Cheap instead?

(random question: would we benefit at all from codegen-ing scalar field accessors?)

graphql/testdata/TestTextMarshaling.snapshots.json
```diff
@@ -55,6 +87,14 @@
       }
     ]
   },
+  {
+    "Name": "invalid uuid func response",
```
jellevandenhooff: why don't either of these have batch executor in the name?

graphql/writer.go
```diff
@@ -0,0 +1,112 @@
...
+type errorRecorder struct {
+	sync.Once
+	err error
+}
+
+func (e *errorRecorder) record(err error) {
+	if err == nil {
+		return
+	}
+	e.Do(func() {
```
jellevandenhooff: is this cheaper or more efficient than a sync.Mutex? would prefer mutex because i find it simpler

graphql/writer.go
```diff
@@ -0,0 +1,112 @@
...
+		e.err = err
+	})
+}
+
+// newTopLevelObjectWriter creates a top-level object writer, this should be
+// the object writer that starts the graphql query.
+func newTopLevelObjectWriter(path string) *objectWriter {
```
jellevandenhooff: should we call this output node? or something else?

graphql/writer.go
```diff
@@ -0,0 +1,112 @@
...
+	path := o.getPath()
+	err = nestPathErrorMulti(path, err)
+	o.errRecorder.record(err)
+}
+
+// getPath traverses the parent list to get the current execution path.
+func (o *objectWriter) getPath() []string {
+	path := make([]string, 0)
+	cur := o
+	for {
```
jellevandenhooff: for cur != nil

graphql/writer.go
```diff
@@ -0,0 +1,112 @@
...
+		if cur.path != "" {
+			path = append(path, cur.path)
+		}
+		cur = cur.parent
+	}
+	return path
+}
+
+// Unwraps the object writer JSON map to a "regular" JSON comparable type.
+func unwrapObjectWriter(src interface{}) interface{} {
```
jellevandenhooff: outputNodeToJson?

reactive/rerunner.go
```diff
@@ -103,6 +103,15 @@ func (c *cache) cleanInvalidated() {
 	}
 }
 
+func (c *cache) cleanAll() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for key := range c.computations {
+		delete(c.computations, key)
```
jellevandenhooff: do we need to do any releasing here?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+		return unwrapBatchObjectResult(ctx, sources, typ, selectionSet, destinations)
+	case *NonNull:
+		return unwrapBatchResult(ctx, sources, typ.Type, selectionSet, destinations)
+	default:
+		panic(typ)
+	}
+}
+
+// Resolves the scalar type value for all the provided sources.
+func unwrapBatchScalarResult(sources []interface{}, typ *Scalar, destinations []*objectWriter) error {
```
jellevandenhooff: executeBatchScalar?

jellevandenhooff: or resolveBatchScalar?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+		dest.Fail(err)
+		return nil
+	}
+	return subFieldExecUnits
+}
+
+// unwrapBatchResult traverses the provided sources and fills in result data and
+// returns new execution units that are required to resolve the rest of the
+// query result.
+func unwrapBatchResult(ctx context.Context, sources []interface{}, typ Type, selectionSet *SelectionSet, destinations []*objectWriter) ([]*ExecutionUnit, error) {
```
jellevandenhooff: resolveBatch?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+	}
+	return nil
+}
+
+// Flattens the sources for the list type and calls into an unwrapper method for
+// the list's subtype.
+func unwrapBatchListResult(ctx context.Context, sources []interface{}, typ *List, selectionSet *SelectionSet, destinations []*objectWriter) ([]*ExecutionUnit, error) {
+	flattenedResps := make([]*objectWriter, 0, len(sources))
+	flattenedSources := make([]interface{}, 0, len(sources))
```
jellevandenhooff: allocate for sum len of sources?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+		} else {
+			destinations[i].Fill(mapVal)
+		}
+	}
+	return nil
+}
+
+// Flattens the sources for the list type and calls into an unwrapper method for
+// the list's subtype.
+func unwrapBatchListResult(ctx context.Context, sources []interface{}, typ *List, selectionSet *SelectionSet, destinations []*objectWriter) ([]*ExecutionUnit, error) {
```
jellevandenhooff: could we make sources [][]interface{} here and make something else responsible for doing the slice munging?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+			destinations[idx].Fill(nil)
+			continue
+		}
+
+		srcType := ""
+		if union.Kind() == reflect.Ptr && union.Elem().Kind() == reflect.Struct {
+			union = union.Elem()
+		}
+		for typString := range typ.Types {
+			inner := union.FieldByName(typString)
```
jellevandenhooff: wish this wasn't a loop. let's fix it later.

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+	for _, selection := range selections {
+		if field, ok := typ.Fields[selection.Name]; ok && field.Expensive {
+			numExpensive++
+		} else if ok && field.External {
+			numNonExpensive++
+		}
+	}
+
+	// For every object, create a "destination" map that we can fill with our
+	// result values.  Filter out invalid/nil objects.
```
jellevandenhooff: why don't slices have to do this?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+			}
+			continue
+		}
+		if field.External {
+			// External non-Expensive fields should be fast (so we can run them at the
+			// same time), but, since they are still external functions we don't want
+			// to run them where they could potentially block.
+			// So we create an execution unit with all the fields to execute
+			// asynchronously.
+			executionUnits = append(executionUnits, &ExecutionUnit{
```
jellevandenhooff: dedup struct with below?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+				ctx:          ctx,
+				field:        field,
+				sources:      nonNilSources,
+				destinations: destForSelection,
+				selection:    selection,
+			})...,
+		)
+	}
+
+	if typ.KeyField != nil {
```
jellevandenhooff: could we get this in the selection somehow? would be nice to have the same expensive etc logic for these

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+	for idx, src := range unit.sources {
+		units = append(units, safeResolveNonBatchUnit(src, unit.destinations[idx], unit)...)
+	}
+	return units
+}
+
+// safeResolveNonBatchUnit resolves a non-batch unit within a panic wrapper.
+func safeResolveNonBatchUnit(src interface{}, dest *objectWriter, unit *ExecutionUnit) []*ExecutionUnit {
+	defer func() {
+		if panicErr := recover(); panicErr != nil {
```
jellevandenhooff: share this with other code?

graphql/batch_executor.go
```diff
@@ -0,0 +1,409 @@
...
+	"strconv"
+
+	"github.com/samsarahq/thunder/reactive"
+)
+
+// ExecutionUnit is a set of execution work that will be done when running
+// a graphql query.  For every source there is an equivalent destination
+// ObjectWriter that is used to record the result of running a section of the
+// graphql query.
+type ExecutionUnit struct {
```
jellevandenhooff: WorkUnit?

graphql/batch_executor.go
willhug: We'd need to pass in a dummy value as the "source".  (Since we filter out nil sources when unwrapping objects).

Honestly the API is a bit odd because I think we always pass in a nil Source in the top-level `Execute` function.  Maybe we can just get rid of it?  Possibly a follow-up?

graphql/schemabuilder/function.go
willhug: Ideally, anything that doesn't change should be stored on the funcContext.  Things that change should be passed in.  Might be worth wrapping back around to this.

graphql/schemabuilder/pagination.go
willhug: Codegening scalar accessors would definitely be advantageous, not sure if the scale would matter that much yet since we have some lower hanging fruit.

graphql/testdata/TestTextMarshaling.snapshots.json
willhug: Whoops, wasn't adding the prefix for error recordings, fixed.

graphql/writer.go
willhug: Under the hood a sync.Once is a boolean atomic and a sync.Mutex.  If we did have a mutex we'd probably want to have the same functionality with a boolean atomic to quick-exit if the error had already been set (which would be more efficient than just a mutex.

reactive/rerunner.go
willhug: Yeah, in order to be sure we need to.

I took a closer look and it looks like we do already release the node in a goroutine (which should get cleared in the clearInvalidated section), but there's no guarantee that the goroutine is run before we start our next execution run.  So I'm going to keep the explicit deletion here which should be safer.

graphql/batch_executor.go
willhug: went with `resolveBatch` and `resolveScalarBatch`, etc

graphql/batch_executor.go
willhug: Would still need to be able handle a Lists of Lists I think.  Not sure if that's possible in graphql.

graphql/batch_executor.go
willhug: Added an iteration step to grab the proper length first.

graphql/batch_executor.go
willhug: I think we get away with it because slices become zero value sized.

graphql/batch_executor.go
willhug: We could probably do it at the parsing layer and add it to the selection set based on the same parameters we're using here.

willhug: Maybe a follow-up?

graphql/batch_executor.go
willhug: Moved the wrapper to be around the field.Resolve func directly.  (Nothing else should panic).

graphql/writer.go
```diff
@@ -0,0 +1,109 @@
...
+
+// errorRecorder is a concurrency-safe way where we can record the first error
+// we get from executing the graphql query.
+type errorRecorder struct {
+	sync.Once
+	err error
+}
+
+func (e *errorRecorder) record(err error) {
+	if err == nil {
```
changpingc: i don't think this is safe as record() might be invoked in another goroutine and setting error pointer

let's get rid of this line?

willhug: this checks the passed in error, should be safe I think?

changpingc: yes i misread

PR #240, opened 2019-04-26
go graphql: Add BatchFieldFuncs to new Executor

willhug:
Summary: Adds support for running fieldFuncs as a "Batch" instead of one at a time. 
The API surface of this forces passing in a Fallback/Non-batch function that has 
the same signature as the batch function.  This is mostly for the transition period,
eventually we'll go full-batch.

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -81,7 +81,7 @@ func (e *BatchExecutor) Execute(ctx context.Context, typ Type, source interface{
 		)
 	}
 
-	e.scheduler.Run(resolveWorkUnit, initialSelectionWorkUnits...)
+	e.scheduler.Run(executeWorkUnit, initialSelectionWorkUnits...)
 
 	if topLevelRespWriter.errRecorder.err != nil {
 		return nil, topLevelRespWriter.errRecorder.err
@@ -89,27 +89,27 @@ func (e *BatchExecutor) Execute(ctx context.Context, typ Type, source interface{
 	return writers, nil
 }
 
-// resolveWorkUnit executes/resolves a work unit and checks the
+// executeWorkUnit executes/resolves a work unit and checks the
 // selections of the unit to determine if it needs to schedule more work (which
 // will be returned as new work units that will need to get scheduled.
-func resolveWorkUnit(unit *WorkUnit) []*WorkUnit {
+func executeWorkUnit(unit *WorkUnit) []*WorkUnit {
 	if unit.field.Batch && unit.selection.UseBatch {
-		return resolveBatchUnit(unit)
+		return executeBatchWorkUnit(unit)
 	}
 
 	var units []*WorkUnit
 	for idx, src := range unit.sources {
 		if !unit.field.Expensive {
-			units = append(units, resolveNonBatchUnit(unit.ctx, src, unit.destinations[idx], unit)...)
+			units = append(units, executeNonBatchWorkUnit(unit.ctx, src, unit.destinations[idx], unit)...)
 			continue
 		}
-		units = append(units, resolveNonBatchUnitWithCaching(src, unit.destinations[idx], unit)...)
+		units = append(units, executeNonBatchWorkUnitWithCaching(src, unit.destinations[idx], unit)...)
 	}
 	return units
 }
 
-func resolveBatchUnit(unit *WorkUnit) []*WorkUnit {
-	results, err := safeResolveBatch(unit.ctx, unit.field, unit.sources, unit.selection.Args, unit.selection.SelectionSet)
+func executeBatchWorkUnit(unit *WorkUnit) []*WorkUnit {
+	results, err := safeExecuteBatchResolver(unit.ctx, unit.field, unit.sources, unit.selection.Args, unit.selection.SelectionSet)
 	if err != nil {
 		for _, dest := range unit.destinations {
 			dest.Fail(err)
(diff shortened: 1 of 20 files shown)


graphql/batch_executor.go
```diff
@@ -106,6 +108,24 @@ func resolveWorkUnit(unit *WorkUnit) []*WorkUnit {
...
 
+func resolveBatchUnit(unit *WorkUnit) []*WorkUnit {
+	results, err := safeResolveBatch(unit.ctx, unit.field, unit.sources, unit.selection.Args, unit.selection.SelectionSet)
+	if err != nil {
+		for _, dest := range unit.destinations {
+			dest.Fail(err)
+		}
+		return nil
+	}
+	unitChildren, err := resolveBatch(unit.ctx, results, unit.field.Type, unit.selection.SelectionSet, unit.destinations)
```
jellevandenhooff: nit: hmmm. seems we've now overloaded resolve. maybe we have executeWorkUnit (instead of resolveBatchUnit), safeExecute[Batch]Resolver, and then resolveBatch, resolveBatchList? although resolveBatch isn't quite what resolve means either in the graphql spec.

graphql/batch_executor.go
```diff
@@ -335,6 +355,16 @@ func resolveObjectBatch(ctx context.Context, sources []interface{}, typ *Object,
 		}
 
 		field := typ.Fields[selection.Name]
+		if field.Batch && selection.UseBatch {
+			workUnits = append(workUnits, &WorkUnit{
+				ctx:          ctx,
```
jellevandenhooff: WorkUnit here looks duplicated a bunch of times which distracts me from the actual logic. How about you create the WorkUnit once, and then later decide where/how to append it?

For the field.Expensive case, you could have a helper that splits a batch work unit into an array of work units.

graphql/batch_executor.go
```diff
@@ -335,6 +355,16 @@ func resolveObjectBatch(ctx context.Context, sources []interface{}, typ *Object,
 		}
 
 		field := typ.Fields[selection.Name]
+		if field.Batch && selection.UseBatch {
```
jellevandenhooff: Maybe add a comment "// Figure out how to resolve this field depending on the type of the resolver.

graphql/batch_test.go
```diff
@@ -0,0 +1,368 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestBatchFieldFuncExecution(t *testing.T) {
+	type Object struct {
+		Key string
+	}
+	tests := []struct {
+		Name                  string
+		GiveObjectFunc        interface{}
```
jellevandenhooff: idiomatic go (at least in the go compiler source) is to have all these fields be private, and to drop the give prefix (since it's implied by not having a want/desired prefix)

graphql/batch_test.go
```diff
@@ -0,0 +1,368 @@
...
+		GiveValueFunc         interface{}
```
jellevandenhooff: is this the resolver? can we call it resolver if it is?

graphql/batch_test.go
```diff
@@ -0,0 +1,368 @@
...
+	)
+	conditions := []string{OldExecutor, NewExecutorNoBatching, NewExecutorBatching}
+	for _, tt := range tests {
+		for _, cond := range conditions {
+			t.Run(fmt.Sprintf("%s %s", tt.Name, cond), func(t *testing.T) {
+				builder := schemabuilder.NewSchema()
+				builder.Query().FieldFunc("objects", tt.GiveObjectFunc)
+
+				obj := builder.Object("object", Object{})
+				obj.BatchFieldFunc("value", tt.GiveValueFunc, tt.GiveValueFallbackFunc, func(ctx context.Context) bool {
```
jellevandenhooff: could we make the batch field func API a little more forward facing? how about we ask that the fallback-resolver is registered through .FieldFunc, and the decision function is passed to .BatchFieldFunc with a special option schemabuilder.UseBatchFieldFuncFunction(...)?

alternatively, and this might be a little nutty, how about we make it the job of the application code to build multiple schemas, and with and one without batching?

or, we let folks register both, and we create a top-level function on the schema builder to supply the batch decision function which will take a type and field name?

graphql/executor.go
```diff
@@ -170,34 +170,42 @@ func PrepareQuery(typ Type, selectionSet *SelectionSet) error {
...
 	case *NonNull:
-		return PrepareQuery(typ.Type, selectionSet)
+		return PrepareQuery(ctx, typ.Type, selectionSet)
 
 	default:
 		panic("unknown type kind")
 	}
 }
 
-type panicError struct {
```
jellevandenhooff: is there any downside to getting rid of this struct?

willhug: Not being used & private, so... probably not?

graphql/batch_test.go
```diff
@@ -0,0 +1,368 @@
...
+					return
+				}
+				require.NoError(t, err)
+
+				wantParsedJSON := internal.ParseJSON(tt.WantResultJSON)
+				gotJSON := internal.AsJSON(res)
+
+				require.True(
+					t,
+					reflect.DeepEqual(gotJSON, wantParsedJSON),
```
changpingc: require.Equal(t, expected, actual)?

graphql/schemabuilder/batch.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	inType := in[0]
+	in = in[1:]
+
+	parentPtrType := reflect.PtrTo(funcCtx.parentTyp)
+	if inType.Kind() != reflect.Map ||
+		inType.Key().Kind() != reflect.Int ||
+		(inType.Elem() != parentPtrType && inType.Elem() != funcCtx.parentTyp) {
+		return nil, fmt.Errorf(
+			"invalid source batch type, expected one of map[int]*%s or map[int]%s, but got %s",
```
changpingc: 👍

graphql/batch_test.go
changpingc: i like separate register calls .FieldFunc, .BatchFieldFunc and a global decision func that takes type and field name. that feels the most user-friendly for porting to batch and for deprecating fallbacks

graphql/batch_executor.go
willhug: Moved this logic into a helper and added a comment.

willhug:
Resolved everything but the API question, there's some weirdness about how the API integrates with the graphql executor that might need some answers (especially if we haven't rolled out/deprecated the old executor yet).  Maybe we can talk in person.

willhug:
Everything should be resolved now, please take another look!

graphql/batch_executor.go
```diff
@@ -335,27 +381,23 @@ func resolveObjectBatch(ctx context.Context, sources []interface{}, typ *Object,
...
-			continue
-		}
 		unit := &WorkUnit{
 			ctx:          ctx,
 			field:        field,
 			sources:      nonNilSources,
 			destinations: destForSelection,
 			selection:    selection,
 		}
+		if shouldUseBatch(selection, field) {
```
jellevandenhooff:
```go
switch {
case shouldUseBatch(selection, field):
			workUnits = append(workUnits, unit)
case filed.Expensive:
			// Expensive fields should be executed as multiple "Units".  The scheduler
			// controls how the units are executed
			workUnits = append(workUnits, splitWorkUnit(unit)...)
case field.External:
			// External non-Expensive fields should be fast (so we can run them at the
			// same time), but, since they are still external functions we don't want
			// to run them where they could potentially block.
			// So we create an work unit with all the fields to execute
			// asynchronously.
			workUnits = append(workUnits, unit)
default:
		// If the fields are not expensive or external the work time should be
		// bounded, so we can resolve them immediately.
		workUnits = append(
			workUnits,
			executeWorkUnit(unit)...,
		)
}
```

jellevandenhooff:
hmmm maybe we want to make the decision function global still? can do that refactor later.

willhug:
I think we might want to make the entry point for the decision function global, but it will probably still be necessary to annotate every field with the existing "choose" function (Because PrepareQuery is built around the graphql typ, not a top-level object), so I don't think much of the internals will need to change too much.

PR #244, opened 2019-05-01
go graphql: Use flattening function for new executor results

willhug:
Summary: I wrote a graphql tree flattening function that would convert
an output node into a JSON tree of map[string]interface{},
[]interface{}, etc.  We didn't use it because we thought we wouldn't
necessarily need it (we could do JSON marshal/unmarshaling).  Since 
the thunder/diff package heavily depends on the map[string]interface{}
type structure to function properly this commit unwraps the new executor's
graphql result.

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -101,7 +101,7 @@ func (e *BatchExecutor) Execute(ctx context.Context, typ Type, source interface{
 	if topLevelRespWriter.errRecorder.err != nil {
 		return nil, topLevelRespWriter.errRecorder.err
 	}
-	return writers, nil
+	return outputNodeToJSON(writers), nil
 }
 
 // executeWorkUnit executes/resolves a work unit and checks the


graphql/batch_executor.go
```diff
@@ -101,7 +101,7 @@ func (e *BatchExecutor) Execute(ctx context.Context, typ Type, source interface{
 	if topLevelRespWriter.errRecorder.err != nil {
 		return nil, topLevelRespWriter.errRecorder.err
 	}
-	return writers, nil
+	return outputNodeToJSON(writers), nil
```
jellevandenhooff: Can tests now stop calling this?

willhug: don't know what you mean...

jellevandenhooff: for some reason I thought there were tests using this outputNodeToJSON before diffing which now might not be necessary anymore, but I can't find any tests doing that.

PR #246, opened 2019-05-01
go graphql: Fix bug with passing back nils/empty results for batchFuncs

willhug:
Summary: Fixes a bug that would occur when a batchFieldFunc would return
less objects than were passed in.  Adds a test which failed before and
passes afterwards.

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -350,7 +350,7 @@ func resolveObjectBatch(ctx context.Context, sources []interface{}, typ *Object,
 	originDestinations := make([]*outputNode, 0, len(destinations))
 	for idx, source := range sources {
 		value := reflect.ValueOf(source)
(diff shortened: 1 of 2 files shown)


graphql/batch_executor.go
```diff
@@ -131,9 +131,19 @@ func executeBatchWorkUnit(unit *WorkUnit) []*WorkUnit {
 		}
 		return nil
 	}
-	unitChildren, err := resolveBatch(unit.ctx, results, unit.field.Type, unit.selection.SelectionSet, unit.destinations)
+	nonNilSrcs := make([]interface{}, 0, len(results))
```
jellevandenhooff: why doesn't this get handled in eg the executing of an object or value? i worry that we have multiple places handling nil, and so eg also multiple places handling non-nullable fields.

willhug: I can make that change.

PR #248, opened 2019-05-03
go schemabuilder: Allow marking batchFieldFuncs as "NonNullable"

willhug:
Summary: Adds support for marking a BatchFieldFunc as "NonNullable". We
don't have the ability to enforce this explicitly through the function
signature, but we can enforce it by validating that responses are being
passed back properly for every node we passed in.  If we see a mismatch
we fail the entire query.

This allows us to have parity with the current non-nullable fieldFuncs
we have right now.  I've added some tests to validate that we fail if a
nil result is returned, and for the happy cases of having non-nil
strings and enum values.

### DIFF
--- a/graphql/schemabuilder/batch.go
+++ b/graphql/schemabuilder/batch.go
@@ -10,7 +10,7 @@ import (
 
 // buildBatchFunction corresponds to buildFunction for a batchFieldFunc
 func (sb *schemaBuilder) buildBatchFunctionWithFallback(typ reflect.Type, m *method) (*graphql.Field, error) {
-	fallbackField, fallbackFuncCtx, err := sb.buildFunctionAndFuncCtx(typ, &method{Fn: m.BatchArgs.FallbackFunc})
+	fallbackField, fallbackFuncCtx, err := sb.buildFunctionAndFuncCtx(typ, &method{Fn: m.BatchArgs.FallbackFunc, MarkedNonNullable: m.MarkedNonNullable})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


graphql/schemabuilder/batch.go
```diff
@@ -312,30 +319,37 @@ func (funcCtx *batchFuncContext) prepareResolveArgs(sources []interface{}, args
...
 
-	return in
+	return in, idxValues
 }
 
 // extractResultsAndErr converts the response from calling the function into
 // the expected type for the response object (as opposed to a reflect.Value).
 // It also handles reading whether the function ended with errors.
-func (funcCtx *batchFuncContext) extractResultsAndErr(numResps int, out []reflect.Value, retType graphql.Type) ([]interface{}, error) {
+func (funcCtx *batchFuncContext) extractResultsAndErr(out []reflect.Value, idxValues []reflect.Value, retType graphql.Type) ([]interface{}, error) {
```
jellevandenhooff: maybe i'm missing something, but the idxValues don't seem to be always reflect.ValueOf(0..n-1), so i think we don't have to do this list thing?

willhug: Wanted to reduce the number of calls we made to `reflect.Value` as much as possible.  Since we already had to set the map keys in the provider function thought it would be more efficient to keep the idxValues around. 🤷‍♂ Seemed kinda weird to remake calls into something we just created a couple of lines before.

willhug: Also didn't like passing in the length attribute, it felt odd.

PR #251, opened 2019-05-09
go reactive: Reset reactive cache on every run.

willhug:
Summary: Whenever we get a retry sentinel error we want to throw away
the current calculations we have to get a clean slate for the next run.

This is required to rollout the new graphql Executor which has a
possibility of caching promises that stored an error (instead of the
result we want).

### DIFF
--- a/reactive/rerunner.go
+++ b/reactive/rerunner.go
@@ -350,23 +350,27 @@ func (r *Rerunner) run() {
 	ctx := context.WithValue(r.ctx, cacheKey{}, r.cache)
 	ctx = context.WithValue(ctx, dependencySetKey{}, &dependencySet{})
 
-	computation, err := run(ctx, r.f)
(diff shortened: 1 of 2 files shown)


reactive/rerunner.go
```diff
@@ -350,23 +350,27 @@ func (r *Rerunner) run() {
...
+		if err != RetrySentinelError {
 			// If we encountered an error that is not the retry sentinel,
 			// we should stop the rerunner.
 			return
 		}
+		// Reset the cache for sentinel errors so we get a clean slate.
+		r.cache = &cache{
+			computations: make(map[interface{}]*computation),
+			locker:       newLocker(),
+		}
```
willhug: maybe need a `go oldCache.CleanupCache()` ?

jellevandenhooff: i think it'll jsut get GCd

PR #253, opened 2019-05-10
sqlgen: add WithShardLimit

jellevandenhooff:
WithShardLimit lets you point a sqlgen.DB add a shard where all values must
satisfy some constraint (eg. customer ID = x.) The check at the sqlgen.DB
layer enforces all operations actually specify the shard key to catch
confused queries going the wrong place.

### DIFF
--- a/sqlgen/db.go
+++ b/sqlgen/db.go
@@ -92,23 +92,6 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
 	return db
 }
 
-func (db *DB) baseCount(ctx context.Context, query *baseCountQuery) (int64, error) {
-	countQuery, err := query.makeCountQuery()
-	if err != nil {
-		return 0, err
-	}
-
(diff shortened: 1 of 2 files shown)


sqlgen/db.go
```diff
@@ -92,7 +94,65 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
...
+func (db *DB) checkFilterAgainstShardLimit(filter Filter) error {
+	if db.shardLimit == nil {
+		return nil
+	}
+	for k, v := range db.shardLimit {
+		filterV, ok := filter[k]
+		if !ok {
+			return fmt.Errorf("db is sharded to require %s = %v, but query does not filter on %s", k, v, k)
+		}
+		if filterV != v {
```
changpingc: oh... int64(orgid) != int(orgid). this might break things

jellevandenhooff: i think unit tests should catch all breakage

PR #254, opened 2019-05-10
go graphql: Add support for using batch.Index as key for all batchFieldFuncs

willhug:
Summary: Introduces a batch.Index type that we can optionally use in
batchFieldFuncs.  This is to make it obvious that we're using an index
and not necessarily a key or Id.

### DIFF
--- a/graphql/schemabuilder/batch.go
+++ b/graphql/schemabuilder/batch.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/samsarahq/thunder/batch"
 	"github.com/samsarahq/thunder/graphql"
 )
 
@@ -123,6 +124,8 @@ type batchFuncContext struct {
 	hasRet          bool
 	hasError        bool
 
+	usesBatchIndex bool
+
 	enforceNoNilResps bool
 
 	funcType     reflect.Type
@@ -184,6 +187,7 @@ func (funcCtx *batchFuncContext) consumeRequiredSourceBatch(in []reflect.Type) (
 			inType.String(),
 		)
 	}
+	funcCtx.usesBatchIndex = isBatchIndexType(inType.Key())
 
 	funcCtx.isPtrFunc = inType.Elem() == parentPtrType
 	funcCtx.batchMapType = inType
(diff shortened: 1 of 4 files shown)


graphql/schemabuilder/batch.go
```diff
@@ -297,7 +317,11 @@ func (funcCtx *batchFuncContext) prepareResolveArgs(sources []interface{}, args
 		idxVal := idx
 		sourceValue := reflect.ValueOf(source)
 		ptrSource := sourceValue.Kind() == reflect.Ptr
-		idxValues[idxVal] = reflect.ValueOf(idxVal)
+		if funcCtx.usesBatchIndex {
+			idxValues[idxVal] = reflect.ValueOf(batch.Index(idxVal))
+		} else {
+			idxValues[idxVal] = reflect.ValueOf(idxVal)
```
jellevandenhooff: why don't we have to do anything in the output processing?

willhug: Because we pass the reflect.Value of the keys directly to the output parsing function

graphql/schemabuilder/batch.go
```diff
@@ -274,6 +290,10 @@ func (funcCtx *batchFuncContext) consumeReturnValue(m *method, sb *schemaBuilder
 	return retType, out, nil
 }
 
+func isBatchIndexType(t reflect.Type) bool {
+	return t.PkgPath() == "github.com/samsarahq/thunder/batch" && t.String() == "batch.Index"
```
willhug: does this need to check for vendoring?

willhug: yuuuuuuup

willhug:
```
v := reflect.ValueOf(graphql.Executor{})
fmt.Println(v.Type().PkgPath())
```
`.../vendor/github.com/samsarahq/thunder/graphql`

willhug: Updated to a `strings.Contains` instead.

changpingc: can't it do a direct type equality check?

```
var batchIndexPointer *batch.Index
batchIndexTyp := reflect.New(reflect.TypeOf(batchIndexPointer)).Elem()
```

willhug: ah, yes, let's do that

willhug: done

PR #255, opened 2019-05-14
go graphql: Short-Circuit execution if no sources

willhug:
Summary: There was an edge-case where we'd schedule another round of
useless work for empty list responses.  This fixes the problem by
stopping the unwrapping if we have no sources/destinations to use.


graphql/batch_executor.go
```diff
@@ -195,6 +195,9 @@ func executeNonBatchWorkUnit(ctx context.Context, src interface{}, dest *outputN
 // returns new work units that are required to resolve the rest of the
 // query result.
 func resolveBatch(ctx context.Context, sources []interface{}, typ Type, selectionSet *SelectionSet, destinations []*outputNode) ([]*WorkUnit, error) {
+	if len(sources) == 0 {
+		return nil, nil
```
jellevandenhooff: please add a unit test

willhug: Added a test that fails without this fix.

PR #257, opened 2019-05-28
Implements Unwrap for SafeErrors

jsm:
We want to be able to extract stack traces from safeErrors. To do this, we're adding an `inner` field to the struct, and implementing `Unwrap` to extract this in order to get to an error type that provides a stack trace.

I've also moved errors out to errors.go for cleanliness.

### DIFF
--- a/graphql/errors.go
+++ b/graphql/errors.go
@@ -12,6 +12,7 @@ type SanitizedError interface {
 }
 
 type SafeError struct {
+	inner   error
 	message string
 }
 
@@ -33,12 +34,23 @@ func (e SafeError) SanitizedError() string {
 	return e.message
 }
 
+// Unwrap returns the wrapped error, implementing go's 1.13 error wrapping proposal.
+func (e SafeError) Unwrap() error {
+	return e.inner
+}
+
 func NewClientError(format string, a ...interface{}) error {
 	return ClientError{message: fmt.Sprintf(format, a...)}
 }
 
 func NewSafeError(format string, a ...interface{}) error {
 	return SafeError{message: fmt.Sprintf(format, a...)}
+}
+
+// WrapAsSafeError wraps an error into a "SafeError", and takes in a message.
+// This message can be used like fmt.Sprintf to take in formatting and arguments.
+func WrapAsSafeError(err error, format string, a ...interface{}) error {
+	return SafeError{inner: err, message: fmt.Sprintf(format, a...)}
 }
 
 func sanitizeError(err error) string {
--- a/graphql/errors.go
+++ b/graphql/errors.go
@@ -1,0 +1,54 @@
+package graphql
+
+import (
+	"fmt"
+
+	"github.com/gorilla/websocket"
+)
+
+type SanitizedError interface {
+	error
+	SanitizedError() string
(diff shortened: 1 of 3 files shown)


graphql/errors.go
```diff
@@ -12,6 +12,7 @@ type SanitizedError interface {
 }
```
stephen: nit: commit message: `graphql: implement Unwrap...`

graphql/errors.go
```diff
@@ -25,6 +26,10 @@ func (e ClientError) SanitizedError() string {
 	return e.message
 }
 
+func (e ClientError) Unwrap() error {
+	return e.inner
+}
```
stephen: maybe just `SafeError` needs it? `ClientError` is only ever emitted by thunder and never wraps

graphql/errors.go
```diff
@@ -41,6 +50,10 @@ func NewSafeError(format string, a ...interface{}) error {
 	return SafeError{message: fmt.Sprintf(format, a...)}
 }
 
+func WrapAsSafeError(err error, format string, a ...interface{}) error {
```
stephen: add a godoc for this?

graphql/errors.go
```diff
@@ -41,6 +50,10 @@ func NewSafeError(format string, a ...interface{}) error {
...
+	return SafeError{inner: err, message: fmt.Sprintf(format, a...)}
```
stephen: how is this message going to be used? if our impl is going to use a middleware to wrap other errors that already have messages, i think we don't need it?

jsm: We want to extract the message from the error, and use that. Thunder will still use `sanitizeError()` for the message, and thus SafeError will still need to provide that

stephen: can we use the underlying message from the err as `inner` instead of asking the caller to provide one?

jsm: Talked about this in person.

To have thunder extract messages itself, would require it to know and understand our error patterns and structure. I'd rather defer the logic of how what goes in the message and how we unwrap it to added middleware

graphql/errors.go
jsm: agreed, updating

graphql/errors.go
jsm: updated

graphql/errors.go
jsm: updated

graphql/errors_test.go
```diff
@@ -0,0 +1,28 @@
...
+	assert.True(t, ok)
+	assert.Nil(t, wrapperError.Unwrap())
+}
+
+func TestWrapAsSafeError(t *testing.T) {
+	sourceErr := errors.New("I am the source error.")
+	err := graphql.WrapAsSafeError(sourceErr, "This is an error.")
+	wrapperError, ok := err.(Wrapper)
+	assert.True(t, ok)
+	assert.NotNil(t, wrapperError.Unwrap())
```
stephen: `assert.Equal(t, sourceErr, wrapperError.Unwrap())`?

jsm: tru

PR #263, opened 2019-06-13
Deleted Old Executor

CodeBrew28:
Removed the old executor as everything can be directed to the batch executor


graphql/server.go
```diff
@@ -283,7 +283,7 @@ func (c *conn) handleMutate(in *inEnvelope) error {
 	}
 
 	initial := true
-	e := Executor{}
+	e := BatchExecutor{}
```
wedaly: I think you'll need to call `NewBatchExecutor()` here, otherwise the scheduler won't be set?  (Old executor doesn't have this problem because it's only field is a mutex, which has a sensible zero value)

graphql/server.go
```diff
@@ -496,7 +496,7 @@ func CreateConnection(ctx context.Context, socket JSONSocket, schema *Schema, op
 		ctx:                ctx,
 		schema:             schema,
 		mutationSchema:     schema,
-		executor:           &Executor{},
+		executor:           &BatchExecutor{},
```
wedaly: Same comment here: probably need to call the constructor

CodeBrew28:
@wedaly Absolutely agree, the PR (https://github.com/samsara-dev/backend/pull/27498) should make sure all queries are using the new executor, which should get merged before

graphql/batch_test.go
```diff
@@ -778,9 +778,6 @@ func TestBatchFieldFuncExecution(t *testing.T) {
 
 				var e graphql.ExecutorRunner
 				e = graphql.NewBatchExecutor(graphql.NewImmediateGoroutineScheduler())
-				if cond == OldExecutor {
```
willhug: Delete the "OldExecutor" Constant plz

graphql/schemabuilder/reflect_test.go
```diff
@@ -167,7 +167,7 @@ func TestExecuteGood(t *testing.T) {
 		t.Error(err)
 	}
 
-	e := graphql.Executor{}
+	e := graphql.NewBatchExecutor(graphql.NewImmediateGoroutineScheduler())
```
willhug: can we rename `NewBatchExecutor` to just `NewExecutor`?

willhug: and have that function return an ExecutorRunner instead of a `*BatchExecutor`

graphql/server.go
```diff
@@ -496,7 +496,7 @@ func CreateConnection(ctx context.Context, socket JSONSocket, schema *Schema, op
 		ctx:                ctx,
 		schema:             schema,
 		mutationSchema:     schema,
-		executor:           &Executor{},
+		executor:           (NewBatchExecutor(NewImmediateGoroutineScheduler())),
```
willhug: don't need the extra set of braces here.

graphql/http.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func HTTPHandler(schema *Schema, middlewares ...MiddlewareFunc) http.Handler {
-	return HTTPHandlerWithExecutor(schema, &Executor{}, middlewares...)
+	return HTTPHandlerWithExecutor(schema, (NewBatchExecutor(NewImmediateGoroutineScheduler())), middlewares...)
```
willhug: Don't need extra braces

graphql/batch_test.go
willhug: Actually, nvm, I'm going to land https://github.com/samsarahq/thunder/pull/264, you can just remove this change and rebase on top of it.

PR #265, opened 2019-06-15
sqlgen: WithShardLimit bug when filter is part of pk

dgiridharan:
When we use UpdateRow, we fail to check if the shard filter is in the
where columns. If the shard is part of PK, the column and value is
added to as part of the columns in the where component of the query.
As a result, we should check both the where and the regular columns
for the shard filter when we update a row.


dgiridharan:
I was missing a lot of test cases. I will rework the PR and send it later

PR #266, opened 2019-06-15
sqlgen: WithShardLimit bug when filter is part of pk

dgiridharan:
When we use UpdateRow, we fail to check if the shard filter is in the
where columns. If the shard is part of PK, the column and value is
added to as part of the columns in the where component of the query.
As a result, we should check both the where and the regular columns
for the shard filter when we update a row.

### DIFF
--- a/sqlgen/db.go
+++ b/sqlgen/db.go
@@ -311,7 +311,9 @@ func (db *DB) UpdateRow(ctx context.Context, row interface{}) error {
 		return err
 	}
 
-	if err := db.checkColumnValuesAgainstShardLimit(query.Columns, query.Values); err != nil {
+	if err := db.checkColumnValuesAgainstShardLimit(
+		append(query.Where.Columns, query.Columns...),
+		append(query.Where.Values, query.Values...)); err != nil {
 		return err
 	}
 
--- a/sqlgen/integration_test.go
+++ b/sqlgen/integration_test.go
@@ -344,6 +344,20 @@ func TestLimit(t *testing.T) {
 		t.Error(err)
 	}
 
+	// Check aliceDb can update alice if pk is part of the shard limit clause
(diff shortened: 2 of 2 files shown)


sqlgen/db.go
```diff
@@ -311,7 +311,9 @@ func (db *DB) UpdateRow(ctx context.Context, row interface{}) error {
 		return err
 	}
 
-	if err := db.checkColumnValuesAgainstShardLimit(query.Columns, query.Values); err != nil {
+	if err := db.checkColumnValuesAgainstShardLimit(
+		append(query.Where.Columns, query.Columns...),
```
dgiridharan: I need to include `Where.Columns` because of the way `MakeUpdateRow()` works in `reflect.go`. `MakeUpdateRow()` adds the `column.Name` and `column.Value` to `where` clause if the `column` is `pk`. As a result, if you shard by `pk` you need to check both the `query.Columns` and `query.Where.Columns`.

dgiridharan: Just for reference, `WithShardLimit` clause was introduced in:
commit: 
https://github.com/samsarahq/thunder/commit/3a53d564b5b2ebc9b3476f597941cea9db693dfe

PR: https://github.com/samsarahq/thunder/pull/253

jellevandenhooff: could this be just the where columns?

dgiridharan: I was thinking about that but unfortunately I don't think so. For example, in the test ⬇️ the first test case tries to shard with a non pk column. In that case `db.Schema.MakeUpdateRow(row)` adds the column to `column.Where` and we would get the same error.

Are you concerned about perf to append arrays?

Would it just be better to send both arrays and iterate over them instead of concentrating them?

PR #269, opened 2019-06-19
Batching Pagination Filters

CodeBrew28:
Added batch filtering and fallback methods for filtering during pagination

To run the tests, use `GOMAXPROCS=1 /Users/arshiamalkani/co/backend/opt/go1.11.5/bin/go test -benchmem -run=^$ github.com/samsarahq/thunder/graphql -bench=.`

Benchmark Test results
```
BenchmarkFiltersBatched10Items             30000             40715 ns/op           16149 B/op        247 allocs/op
BenchmarkFiltersNotBatched10Items          30000             58969 ns/op           15419 B/op        245 allocs/op
BenchmarkFiltersBatched100Items            10000            120814 ns/op           69603 B/op        930 allocs/op
BenchmarkFiltersNotBatched100Items          3000            436085 ns/op           68804 B/op       1064 allocs/op
BenchmarkFiltersBatched1000Items            2000           1114082 ns/op          752992 B/op       7599 allocs/op
BenchmarkFiltersNotBatched1000Items          300           5026686 ns/op          641897 B/op       9165 allocs/op
```

### DIFF
--- a/graphql/benchmark_test.go
+++ b/graphql/benchmark_test.go
@@ -1,0 +1,115 @@
+package graphql_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/samsarahq/thunder/batch"
+	"github.com/samsarahq/thunder/graphql"
+	"github.com/samsarahq/thunder/graphql/schemabuilder"
+	"github.com/stretchr/testify/require"
+)
+
+func pagintedQueryWithFilterBenchmark(b *testing.B, n int, batchFunc bool) {
+	schema := schemabuilder.NewSchema()
+	type Inner struct {
+	}
+	query := schema.Query()
+	query.FieldFunc("inner", func() Inner {
+		return Inner{}
+	})
+	inner := schema.Object("inner", Inner{})
+	item := schema.Object("item", Item{})
+	item.Key("id")
+	filterTexts := [5]string{"can", "man", "cannot", "soban", "socan"}
+	items := make([]Item, n, n)
+	for i := 0; i < n; i++ {
+		items[i] = Item{Id: int64(i), FilterText: filterTexts[i%5], String: "a"}
+	}
+	if !batchFunc {
+		inner.FieldFunc("innerConnectionWithFilter", func() []Item {
+			return items
+		}, schemabuilder.Paginated,
+			schemabuilder.FilterField("foo",
+				func(ctx context.Context, i Item) string {
+					return i.FilterText
+				},
+			),
+			schemabuilder.FilterField("bar",
+				func(ctx context.Context, i Item) string {
+					return i.String
+				},
+			),
+		)
+	} else {
+		inner.FieldFunc("innerConnectionWithFilter", func() []Item {
+			return items
+		}, schemabuilder.Paginated,
+			schemabuilder.BatchFilterField("foo",
+				func(ctx context.Context, items map[batch.Index]Item) (map[batch.Index]string, error) {
+					myMap := make(map[batch.Index]string, len(items))
+					for i, item := range items {
+						myMap[i] = item.FilterText
+					}
+					return myMap, nil
+				},
+			),
+			schemabuilder.BatchFilterField("bar",
+				func(ctx context.Context, items map[batch.Index]Item) (map[batch.Index]string, error) {
+					myMap := make(map[batch.Index]string, len(items))
+					for i, item := range items {
+						myMap[i] = item.String
+					}
+					return myMap, nil
+				},
+			),
+		)
+	}
+	builtSchema := schema.MustBuild()
+	q := graphql.MustParse(`
+		{
+			inner {
+				innerConnectionWithFilter(filterText: "can",first: 4, after: "") {
+					totalCount
+					edges {
+						node {
+							id
+						}
+						cursor
+					}
+				}
+			}
+		}`, nil)
+	graphql.PrepareQuery(context.Background(), builtSchema.Query, q.SelectionSet)
+	exeuctor := graphql.NewExecutor(graphql.NewImmediateGoroutineScheduler())
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := exeuctor.Execute(context.Background(), builtSchema.Query, nil, q)
+		require.NoError(b, err)
+	}
+
+}
+
+func BenchmarkFiltersBatched10Items(b *testing.B) {
+	pagintedQueryWithFilterBenchmark(b, 10, true)
+}
+
+func BenchmarkFiltersNotBatched10Items(b *testing.B) {
+	pagintedQueryWithFilterBenchmark(b, 10, false)
+}
+
+func BenchmarkFiltersBatched100Items(b *testing.B) {
+	pagintedQueryWithFilterBenchmark(b, 100, true)
+}
+
+func BenchmarkFiltersNotBatched100Items(b *testing.B) {
+	pagintedQueryWithFilterBenchmark(b, 100, false)
+}
+
+func BenchmarkFiltersBatched1000Items(b *testing.B) {
+	pagintedQueryWithFilterBenchmark(b, 1000, true)
+}
+
+func BenchmarkFiltersNotBatched1000Items(b *testing.B) {
+	pagintedQueryWithFilterBenchmark(b, 1000, false)
+}
--- a/graphql/connection_test.go
+++ b/graphql/connection_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/samsarahq/thunder/batch"
 	"github.com/samsarahq/thunder/graphql"
 	"github.com/samsarahq/thunder/graphql/schemabuilder"
 	"github.com/samsarahq/thunder/internal"
@@ -59,14 +60,28 @@ func TestConnection(t *testing.T) {
 			{Id: 4, FilterText: "soban"},
 			{Id: 5, FilterText: "socan"},
 		}
-	}, schemabuilder.Paginated, schemabuilder.TextFilterFields{
-		"foo": func(ctx context.Context, i Item) string {
-			return i.FilterText
-		},
-		"bar": func(ctx context.Context, i *Item) string {
-			return ""
-		},
-	})
+
+	}, schemabuilder.Paginated,
+		schemabuilder.BatchFilterField("foo",
+			func(ctx context.Context, i map[batch.Index]Item) (map[batch.Index]string, error) {
+				myMap := make(map[batch.Index]string, len(i))
+				for i, item := range i {
+					myMap[i] = item.FilterText
+				}
+				return myMap, nil
+			},
+		),
+		schemabuilder.BatchFilterField("bar",
+			func(ctx context.Context, i map[batch.Index]Item) (map[batch.Index]string, error) {
+				myMap := make(map[batch.Index]string, len(i))
+				for i := range i {
+					myMap[i] = ""
+				}
+				return myMap, nil
+			},
+		),
+	)
+
 	inner.FieldFunc("innerConnectionWithSort", func() []Item {
 		return []Item{
 			{Id: 1, Number: 1, String: "1", Float: 1.0},
@@ -487,9 +502,12 @@ func TestPaginateBuildFailure(t *testing.T) {
 		object.Key("id")
 		inner.FieldFunc("connection", func(ctx context.Context, args Args) ([]StructWithKey, error) {
 			return nil, nil
-		}, schemabuilder.Paginated, schemabuilder.TextFilterFields{
-			"noargs": func() {},
-		})
+		},
+			schemabuilder.Paginated,
+			schemabuilder.BatchFilterField(
+				"noargs",
+				func() {},
+			))
 
 		_, err := schema.Build()
 		require.Error(t, err)
@@ -511,9 +529,9 @@ func TestPaginateBuildFailure(t *testing.T) {
 		object.Key("id")
 		inner.FieldFunc("connection", func(ctx context.Context, args Args) ([]StructWithKey, error) {
 			return nil, nil
-		}, schemabuilder.Paginated, schemabuilder.TextFilterFields{
-			"intReturn": func() int64 { return 0 },
-		})
+		}, schemabuilder.Paginated, schemabuilder.BatchFilterField(
+			"intReturn", func() int64 { return 0 },
+		))
 
 		_, err := schema.Build()
 		require.Error(t, err)
@@ -533,11 +551,15 @@ func TestPaginateBuildFailure(t *testing.T) {
 		object.Key("id")
 		inner.FieldFunc("connection", func(ctx context.Context, i *Inner, args Args) ([]StructWithKey, error) {
 			return nil, nil
-		}, schemabuilder.Paginated, schemabuilder.TextFilterFields{
-			"someArgs": func(ctx context.Context, i *StructWithKey, args Args) string {
-				return ""
+		}, schemabuilder.Paginated, schemabuilder.BatchFilterField(
+			"someArgs", func(ctx context.Context, i map[batch.Index]*StructWithKey, args Args) (map[batch.Index]string, error) {
+				myMap := make(map[batch.Index]string, len(i))
+				for i := range i {
+					myMap[i] = ""
+				}
+				return myMap, nil
 			},
-		})
+		))
 
 		_, err := schema.Build()
 		require.Error(t, err)
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"sync"
 
+	"github.com/samsarahq/thunder/batch"
 	"github.com/samsarahq/thunder/graphql"
 	"github.com/samsarahq/thunder/internal/filter"
 	"golang.org/x/sync/errgroup"
@@ -21,6 +23,7 @@ type Connection struct {
 }
 
 var typeOfString = reflect.TypeOf("")
+var typeofMap = reflect.TypeOf(map[batch.Index]string{})
 
 // paginateManually applies the pagination arguments to the edges in memory and sets hasNextPage +
 // hasPrevPage. The behavior is expected to conform to the Relay Cursor spec:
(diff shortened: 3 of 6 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -401,34 +403,54 @@ func (c *connectionContext) applyTextFilter(ctx context.Context, nodes []interfa
...
+					if !ok {
+						return fmt.Errorf("filter %s returned %T, must be a string", name, text)
+					}
+
+					if filter.Match(textString, *args.FilterText) {
+						nodesToKeep[i] = true
+					}
 				}
+			} else {
+				for unscopedI, unscopedNode := range nodes {
```
willhug: this is probably going to have some negative performance implications.

graphql/schemabuilder/pagination.go
```diff
@@ -393,55 +395,96 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
 
 	g, ctx := errgroup.WithContext(ctx)
-	for unscopedI, unscopedNode := range nodes {
-		i, node := unscopedI, unscopedNode
-		g.Go(func() error {
-			keep := false
-			for name, filterField := range c.FilterTextFields {
-				// Resolve the graphql.Field made for sorting.
-				text, err := filterField.Resolve(ctx, node, nil, nil)
+	g.Go(func() error {
```
willhug: don't need the g.Go anymore, we only have one execution here.

graphql/schemabuilder/pagination.go
```diff
@@ -393,55 +395,96 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
+		for name, filterField := range c.FilterTextFields {
+			// Resolve the graphql.Field made for sorting.
```
willhug: maybe we could do it once for each filterField though.

willhug: Do a goroutine to get all the FilterField results in parallel, wait for them to finish, and filter all the text at the same time.

graphql/schemabuilder/pagination.go
```diff
@@ -393,55 +395,96 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
+	}
+
+	nodesToKeep := make([]bool, len(nodes))
+
+	g, ctx := errgroup.WithContext(ctx)
+	for name, filterField := range c.FilterTextFields {
+		if !filterField.Batch && !filterField.UseBatchFunc(ctx) {
+			for unscopedI, unscopedNode := range nodes {
+				g.Go(func() error {
+					i, node := unscopedI, unscopedNode
```
willhug: can't do this here, we need to do this outside of the goroutine function or the `unscopedI, unscopedNode` variables will all be the same "last" parameter for all the goroutine functions (we'll all override the same point and have race conditions).

graphql/schemabuilder/pagination.go
```diff
@@ -393,55 +395,96 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
 
-			return nil
-		})
 	}
 
 	if err := g.Wait(); err != nil {
 		return nil, err
 	}
 
+	batchedNodesToKeep, err := c.applyBatchTextFilter(ctx, nodes, args)
```
willhug: can we run this in a `g.Go(` sub func as well (so we run it async as well).

graphql/schemabuilder/pagination.go
```diff
@@ -393,55 +395,96 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
+	if err != nil {
+		return nil, err
+	}
 	var filteredNodes []interface{}
 
-	for i, keep := range nodesToKeep {
-		if keep {
+	for i, _ := range nodesToKeep {
```
willhug: `i := range nodesToKeep`

graphql/schemabuilder/pagination.go
```diff
@@ -702,23 +745,55 @@ func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ
...
+			if field.Args != nil && len(field.Args) > 0 {
+				return fmt.Errorf("invalid text filter field %s: text filter fields can't take arguments", name)
+			}
+			c.FilterTextFields[name] = field
+		} else {
+			field, err := sb.buildFunction(typ, &method{Fn: fn.FilterFunc, Batch: false})
+			if err != nil {
+				return err
+			}
+			if field.Args != nil && len(field.Args) > 0 {
```
willhug: this check is the same for all three statements here, can we share them?

graphql/schemabuilder/pagination.go
```diff
@@ -702,23 +745,55 @@ func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ
...
+				return fmt.Errorf("invalid text filter field %s: text filter fields can't take arguments", name)
+			}
+			c.FilterTextFields[name] = field
```
willhug: move to a lower line?
```
var field *Field
if ... {...}
else if... {}
else ... {}
If field.Args != nil ....
c.FilterTextFields[name] = field

File: graphql/schemabuilder/types.go [hunk 9]
```diff
@@ -44,10 +44,62 @@ var Paginated fieldFuncOptionFunc = func(m *method) {
...
+		Name:            name,
+		FilterFunc:      filter,
+		BatchFilterFunc: batchFilter,
+		FallbackFlag:    flag,
+	}
 
-func (s TextFilterFields) apply(m *method) {
-	m.TextFilterFuncs = s
+	var fieldFuncTextFilterFields fieldFuncOptionFunc = func(m *method) {
+
```
willhug: remove space.

graphql/schemabuilder/types.go
```diff
@@ -44,10 +44,62 @@ var Paginated fieldFuncOptionFunc = func(m *method) {
...
+		if m.TextFilterFuncs == nil {
+			m.TextFilterFuncs = map[string]Filter{}
+		}
+		m.TextFilterFuncs[name] = filterStruct
```
willhug: hrm, would be good to validate we're not registering the same name twice here.   Maybe add a panic?  This should only be on startup so it should be fine, and we run through this code in tests so we should crash as soon as possible.

graphql/schemabuilder/pagination.go
willhug: We should probably also write a test that would catch this.  Any test that has multiple nodes should probably work.

graphql/connection_test.go
```diff
@@ -511,9 +529,9 @@ func TestPaginateBuildFailure(t *testing.T) {
 		object.Key("id")
 		inner.FieldFunc("connection", func(ctx context.Context, args Args) ([]StructWithKey, error) {
 			return nil, nil
-		}, schemabuilder.Paginated, schemabuilder.TextFilterFields{
-			"intReturn": func() int64 { return 0 },
-		})
+		}, schemabuilder.Paginated, schemabuilder.BatchFilterField(
```
willhug: This shouldn't work right?

CodeBrew28: nope, I'll double check, but I'm pretty sure this is testing an error case

CodeBrew28: yup, it returns a `"unsupported return type int64"`

graphql/schemabuilder/pagination.go
```diff
@@ -393,39 +396,33 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
+	mux         sync.Mutex
+}
 
+func (c *connectionContext) applyBatchTextFilter(ctx context.Context, nodes []interface{}, args PaginationArgs, batchedFields map[string]*graphql.Field) ([]bool, error) {
 	g, ctx := errgroup.WithContext(ctx)
-	for unscopedI, unscopedNode := range nodes {
-		i, node := unscopedI, unscopedNode
+	matchStrings := filter.GetMatchStrings(*args.FilterText)
+	s := SafeBatchNodesToKeep{nodesToKeep: make([]bool, len(nodes), len(nodes))}
+	for unscopeName, unscopedFilterField := range c.FilterTextFields {
```
willhug: use `batchedFields`

graphql/benchmark_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"context"
+	"testing"
+
+	"github.com/samsarahq/thunder/batch"
+	"github.com/samsarahq/thunder/graphql"
+	"github.com/samsarahq/thunder/graphql/schemabuilder"
+	"github.com/stretchr/testify/require"
+)
+
+func pagintedQueryWithFilterBenchmark(n int, b *testing.B, batchFunc bool) {
```
willhug: `b *testing.B` as the first parameter

graphql/benchmark_test.go
```diff
@@ -0,0 +1,126 @@
...
+	for i := 0; i < b.N; i++ {
+		_, err := exeuctor.Execute(context.Background(), builtSchema.Query, nil, q)
+		require.NoError(b, err)
+	}
+
+}
+
+func BenchmarkFiltersBatched10Items(b *testing.B) {
+	pagintedQueryWithFilterBenchmark(10, b, true)
+	// pagintedQueryWithFilterBenchmark(100, b, true)
```
willhug: remove comments

graphql/schemabuilder/pagination.go
```diff
@@ -393,39 +396,33 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
+	nodesToKeep []bool
+	mux         sync.Mutex
+}
 
+func (c *connectionContext) applyBatchTextFilter(ctx context.Context, nodes []interface{}, args PaginationArgs, batchedFields map[string]*graphql.Field) ([]bool, error) {
 	g, ctx := errgroup.WithContext(ctx)
-	for unscopedI, unscopedNode := range nodes {
-		i, node := unscopedI, unscopedNode
+	matchStrings := filter.GetMatchStrings(*args.FilterText)
+	s := SafeBatchNodesToKeep{nodesToKeep: make([]bool, len(nodes), len(nodes))}
```
willhug: Instead of this wrapper object, can we just inline the lock and nodesToKeep object?

willhug: also, only need `make([]bool, len(nodes))`

graphql/schemabuilder/pagination.go
```diff
@@ -434,14 +431,74 @@ func (c *connectionContext) applyTextFilter(ctx context.Context, nodes []interfa
...
+	for name, filterField := range c.FilterTextFields {
+		if filterField.Batch && filterField.UseBatchFunc(ctx) {
+			filterTextFieldsBatched[name] = filterField
+		} else {
+			filterTextFieldsNotBatched[name] = filterField
+		}
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
+	matchString := filter.GetMatchStrings(*args.FilterText)
```
willhug: `s/matchString/matchStrings/`

graphql/schemabuilder/pagination.go
```diff
@@ -434,14 +431,74 @@ func (c *connectionContext) applyTextFilter(ctx context.Context, nodes []interfa
...
+					text, err := filterField.Resolve(ctx, node, nil, nil)
+					if err != nil {
+						return err
+					}
+					// Only strings are allowed for FilterText fields.
+					textString, ok := text.(string)
+					if !ok {
+						return fmt.Errorf("filter %s returned %T, must be a string", name, text)
+					}
+					// if filter.Match(textString, *args.FilterText) {
```
willhug: rm comment

graphql/schemabuilder/pagination.go
```diff
@@ -686,10 +743,7 @@ func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ refle
 
 		// Build a GraphQL field for the function.
 		field, err := sb.buildFunction(typ, &method{Fn: fn})
-		if err != nil {
-			return err
-		}
-		if field.Args != nil && len(field.Args) > 0 {
+		if err != nil || len(field.Args) > 0 {
```
willhug: why did we get rid of the error return here?  If we get an error from the build func we won't see what the issue was.

graphql/schemabuilder/pagination.go
```diff
@@ -702,23 +756,45 @@ func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ
...
+		if fn.FilterFunc != nil {
+			funcTyp := getFuncReturnType(fn.FilterFunc)
+			if funcTyp != typeOfString {
+				return fmt.Errorf("invalid text filter field %s: unsupported return type %v, must be a string", name, funcTyp)
+			}
+		}
 
-		if funcTyp != typeOfString {
-			return fmt.Errorf("invalid text filter field %s: unsupported return type %v, must be a string", name, funcTyp)
+		var field *graphql.Field = nil
```
willhug: don't need the `= nil` part here.

graphql/schemabuilder/pagination.go
```diff
@@ -702,23 +756,45 @@ func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ
...
+		var err error = nil
```
willhug: ditto

graphql/schemabuilder/types.go
```diff
@@ -44,10 +44,58 @@ var Paginated fieldFuncOptionFunc = func(m *method) {
...
+func BatchFilterField(name string, batchFilter interface{}) FieldFuncOption {
+	var filterStruct = Filter{
+		Name:            name,
+		BatchFilterFunc: batchFilter,
+	}
+	var fieldFuncTextFilterFields fieldFuncOptionFunc = func(m *method) {
+		if m.TextFilterFuncs == nil {
+			m.TextFilterFuncs = map[string]Filter{}
+		}
+		if filter, ok := m.TextFilterFuncs[name]; ok {
```
willhug: can we add this check to the other two FilterFields as well?

internal/filter/filter.go
```diff
@@ -29,8 +25,22 @@ func Match(str, query string) bool {
...
+}
 
-		if strings.Contains(strings.ToLower(str), strings.ToLower(matchString)) {
+// Split terms by spaces, except for spaces within quotes.
+// Like http://stackoverflow.com/questions/16261635/javascript-split-string-by-space-but-ignore-space-in-quotes-notice-not-to-spli
+// except:
+// - Treat the query ["san fran] as having the single term ["san fran"] instead of ["san", "fran"]
+// - Removes the delimiting quotes.
+func MatchText(str string, matchStrings []string, query string) bool {
+	if query == "" {
```
willhug: Let's get rid of this check.  Which should never get to it.  Also it allows us to remove the `query` variable from the entire stack.

internal/filter/filter.go
```diff
@@ -29,8 +25,22 @@ func Match(str, query string) bool {
...
+		return true
+	}
+	for _, matchString := range matchStrings {
+		if strings.Contains(strings.ToLower(str), strings.ToLower(matchString)) && matchString != "" {
```
willhug: can we just filter out empty matchStrings before we get here?

internal/filter/filter.go
```diff
@@ -29,8 +25,22 @@ func Match(str, query string) bool {
 		if matchString == "" {
 			matchString = match[2]
 		}
```
willhug: `if matchString == "" { continue }` ?

internal/filter/filter.go
willhug: or maybe we can replace it with a check if `matchStrings` is an empty list/nil (should be equivalent to "we have no query values)"

graphql/schemabuilder/pagination.go
```diff
@@ -393,39 +396,33 @@ func (c *connectionContext) pagesFromEdges(edges []Edge, limit int) (pages []str
...
-		return nodes, nil
-	}
-
-	nodesToKeep := make([]bool, len(nodes))
+type SafeBatchNodesToKeep struct {
+	nodesToKeep []bool
+	mux         sync.Mutex
+}
 
+func (c *connectionContext) applyBatchTextFilter(ctx context.Context, nodes []interface{}, args PaginationArgs, batchedFields map[string]*graphql.Field) ([]bool, error) {
```
willhug: instead of passing in the paginationArgs let's pass in the `matchStrings` directly.

graphql/schemabuilder/pagination.go
CodeBrew28: It needs to be defined beforehand, so that I can correctly initialize them later depending on if its a batch/nonbatch/fallback filter, and then I can do the error checking after. Since the error checking is the same for all 3 cases, its cleaner to just define these variables at the top

willhug: oh, yeah, but you can just do `var field *graphql.Field` it automatically defaults to nil

internal/filter/filter.go
```diff
@@ -7,32 +7,38 @@ import (
...
 
-		if strings.Contains(strings.ToLower(str), strings.ToLower(matchString)) {
-			return true
+func MatchText(str string, matchStrings []string) bool {
+	if len(matchStrings) == 0 {
+		return true
+	}
+	for _, matchString := range matchStrings {
+		if matchString != "" {
+			if strings.Contains(strings.ToLower(str), strings.ToLower(matchString)) && matchString != "" {
```
willhug: already checked the `matchString != ""` above

internal/filter/filter.go
```diff
@@ -7,32 +7,38 @@ import (
...
 			continue
 		}
-		// Get relevant match (one of these has to be a non-empty string, otherwise this wouldn't be
-		// a match).
+		// Get relevant match (one of these has to be a non-empty string, otherwise this wouldn't be a match).
 		matchString := match[1]
 		if matchString == "" {
 			matchString = match[2]
 		}
+		matchStrings[i] = matchString
```
willhug: can we do:
```
if matchString != "" {
  matchStrings[i] = matchString
}
```
?

graphql/schemabuilder/pagination.go
```diff
@@ -702,23 +755,45 @@ func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ
...
-		// Build a GraphQL field for the function.
-		field, err := sb.buildFunction(typ, &method{Fn: fn})
 		if err != nil {
 			return err
 		}
 		if field.Args != nil && len(field.Args) > 0 {
 			return fmt.Errorf("invalid text filter field %s: text filter fields can't take arguments", name)
 		}
 		c.FilterTextFields[name] = field
-	}
```
willhug: don't need the extra space.

internal/filter/filter.go
willhug: actually I guess we don't need it because of the check above, we have a guarantee that both match[1] and match[2] are not empty strings.

internal/filter/filter.go
```diff
@@ -7,32 +7,38 @@ import (
...
-func Match(str, query string) bool {
+func GetMatchStrings(query string) []string {
 	if query == "" {
-		return true
+		return nil
 	}
-
 	matches := matchGroups.FindAllStringSubmatch(query, -1)
-	for _, match := range matches {
+	matchStrings := make([]string, len(matches), len(matches))
```
willhug: woah, this should not be preallocated.

We want 
```
matchStrings := make([]string, 0, len(matches))
for _, match := range matches {
 ...
  matchStrings = append(matchStrings, matchString)
```

This is probably why we're seeing nils everywhere.

internal/filter/filter.go
```diff
@@ -7,30 +7,34 @@ import (
...
 			continue
 		}
-		// Get relevant match (one of these has to be a non-empty string, otherwise this wouldn't be
-		// a match).
+		// Get relevant match (one of these has to be a non-empty string, otherwise this wouldn't be a match).
 		matchString := match[1]
 		if matchString == "" {
 			matchString = match[2]
 		}
+		// matchStrings[i] = matchString
```
willhug: remove commented code

PR #270, opened 2019-06-19
go batch: Make batch.Index a struct (not an int)

willhug:
Summary: It's too easy right now to accidentally make a batch.Index from a
device Id or equivalent.  This commit changes the batch.Index to be a
struct type which cannot be explicitly created, you must use a function to
create it.  This should help reduce all the invalid usages of the
batch.Index.

Also removes the ability to create map[int]Blah batches, because that's more error-prone.


batch/index.go
```diff
@@ -1,5 +1,15 @@
...
+// Index is a "key" mapper for batch requests.  It's key is not special and
+// should be treated the same as an index in a slice.
+type Index struct {
+	key int
+}
+
+// NewIndexForBatchCreationOnly creates an index.  This should only be used by
+// generated code or the creator of batches,  The "Key" has no purpose and should
+// not be used elsewhere.
+func NewIndexForBatchCreationOnly(key int) Index {
```
changpingc: nit: NewIndex() is fine

batch/index.go
```diff
@@ -1,5 +1,15 @@
 package batch
 
-// Index is an alias for the int type, it should be used in all batch field
-// funcs.
-type Index int
+// Index is a "key" mapper for batch requests.  It's key is not special and
```
wedaly: nit: small typo here, "it's" --> "its"

batch/index.go
```diff
@@ -1,5 +1,15 @@
...
+// should be treated the same as an index in a slice.
+type Index struct {
+	key int
+}
+
+// NewIndexForBatchCreationOnly creates an index.  This should only be used by
+// generated code or the creator of batches,  The "Key" has no purpose and should
+// not be used elsewhere.
+func NewIndexForBatchCreationOnly(key int) Index {
+	// TODO generate a hash to really mess with people trying to recreate batches?
```
wedaly: I wonder if this is something we could catch with a lint check?

PR #279, opened 2019-06-29
go diff: Fix long-standing diffing bug 

willhug:
Summary: Noticed and brought up by Edwin last week. This bug has existed
for a long time.  The way it would manifest, is that when a field would
get replaced with a "Nil" value in the execution graph, instead of
diffing the value and returning "nil" as the new value, we were using
the result of a failed cast `newVal, ok :=
val.(map[string]interface{})`.
The `newVal` would be an empty map (instead of the "nil" value we really
wanted).  This would lead to weird behavior, where, on an update,
instead of returning a nil value for something like current driver, we'd
return an empty object instead.

### DIFF
--- a/diff/diff.go
+++ b/diff/diff.go
@@ -138,7 +138,7 @@ func diffMap(old map[string]interface{}, newAny interface{}) interface{} {
 	// Verify the type of new.
 	new, ok := newAny.(map[string]interface{})
 	if !ok {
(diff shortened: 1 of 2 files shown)


willhug:
oh boy, a travis-ci outage: https://www.traviscistatus.com/incidents/g9xm3l02dknv

PR #281, opened 2019-07-05
go graphql/schemabuilder: add Expensive field func option

wedaly:
Add a field func option to mark a field func as "expensive".

Previously, Thunder would assume that any field func accepting a "context" parameter must be expensive and so should be executed concurrently.

The new behavior is to assume that a field func is inexpensive unless the Expensive option is set explicitly.

### DIFF
--- a/graphql/schemabuilder/types.go
+++ b/graphql/schemabuilder/types.go
@@ -44,6 +44,12 @@ var Paginated fieldFuncOptionFunc = func(m *method) {
 	m.Paginated = true
 }
 
+// Expensive is an option that can be passed to a FieldFunc to indicate that
+// the function is expensive to execute, so it should be parallelized.
+var Expensive fieldFuncOptionFunc = func(m *method) {
+	m.Expensive = true
+}
+
 type Filter struct {
 	Name            string
 	FilterFunc      interface{}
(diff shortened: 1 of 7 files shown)


graphql/schemabuilder/types.go
```diff
@@ -44,6 +44,14 @@ var Paginated fieldFuncOptionFunc = func(m *method) {
 	m.Paginated = true
 }
 
+// Inexpensive is an option that can be passed to a FieldFunc to indicate that
+// the function is inexpensive to execute, so it does not need to be parallelized.
+// Note that field funcs that do not accept a `context.Context` argument
+// are always considered inexpensive.
+var Inexpensive fieldFuncOptionFunc = func(m *method) {
```
wedaly: Perhaps there's a clearer name for this?  I used "inexpensive" for consistency with the existing Thunder code, but I can imagine other options (maybe "force serial execution" to more clearly connect the option with the change in runtime behavior?)

jellevandenhooff:
Could we flip the default? I’d rather mark every existing field func
expensive with some code mod if that means no one will get surprise
goroytines in the future.

wedaly:
@jellevandenhooff Yeah, that makes sense -- I'll update the PR.  Is there a way you'd suggest to automate that migration?

wedaly:
Updated!

PR #283, opened 2019-07-11
FilterTextOptions to Select Filters

CodeBrew28:
FilterText allows us to query specific fields when we make paginated requests. However, we run the filterText against all the filters. FilterTextOptions allow us to specify which filters to run it against. The api now would look something like

```
filterByCan: innerConnectionWithFilter(filterText: "can", filterTextField:["foo","wug"],first: 5, after: "") 
```

### DIFF
--- a/graphql/connection_test.go
+++ b/graphql/connection_test.go
@@ -280,7 +280,7 @@ func TestConnection(t *testing.T) {
 
 	snap.SnapshotQuery("Pagination, filter", `{
 		inner {
-			filterByCan: innerConnectionWithFilter(filterText: "can", first: 5, after: "") {
+			filterByCan: innerConnectionWithFilter(filterText: "can", filterTextFields: ["foo","wug"], first: 5, after: "") {
 				totalCount
 				edges {
 					node {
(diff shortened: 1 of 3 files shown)


graphql/connection_test.go
```diff
@@ -280,7 +280,7 @@ func TestConnection(t *testing.T) {
 
 	snap.SnapshotQuery("Pagination, filter", `{
 		inner {
-			filterByCan: innerConnectionWithFilter(filterText: "can", first: 5, after: "") {
+			filterByCan: innerConnectionWithFilter(filterText: "can", filterTextOption:["foo","wug"],first: 5, after: "") {
```
ehzhang: how about `filterTextFields` to match the `TextFilterFields` naming we use to declare the text filter fields

graphql/schemabuilder/pagination.go
```diff
@@ -442,8 +454,10 @@ func (c *connectionContext) applyTextFilter(ctx context.Context, nodes []interfa
 
 	filterTextFieldsNotBatched := make(map[string]*graphql.Field)
 	filterTextFieldsBatched := make(map[string]*graphql.Field)
-
 	for name, filterField := range c.FilterTextFields {
+		if args.FilterTextFields != nil && Index(*args.FilterTextFields, name) < 0 {
```
jellevandenhooff: instead of this index call, make a map[string]bool with fields to filter by

PR #285, opened 2019-07-15
Non-expensive Filter Execution

CodeBrew28:
Paginated filters that aren't batched should only spawn go routines per node if every node is doing an expensive operation. If a paginated filter is not doing an expensive operation, the overhead of spawning that many go routines takes more time. The benchmarks below show this.

```
BenchmarkFiltersBatched10Items                     30000             39562 ns/op           15851 B/op        247 allocs/op
BenchmarkFiltersNotBatched10Items                  50000             28960 ns/op           12886 B/op        234 allocs/op
BenchmarkFiltersNotBatched10ItemsExpensive         30000             60457 ns/op           14703 B/op        245 allocs/op
BenchmarkFiltersBatched100Items                    10000            114775 ns/op           69400 B/op        930 allocs/op
BenchmarkFiltersNotBatched100Items                 10000            105256 ns/op           47515 B/op        963 allocs/op
BenchmarkFiltersNotBatched100ItemsExpensive         3000            384842 ns/op           63854 B/op       1064 allocs/op
BenchmarkFiltersBatched1000Items                    2000            845294 ns/op          753633 B/op       7598 allocs/op
BenchmarkFiltersNotBatched1000Items                 2000            806417 ns/op          434279 B/op       8181 allocs/op
BenchmarkFiltersNotBatched1000ItemsExpensive         300           3991979 ns/op          594366 B/op       9161 allocs/op
```

### DIFF
--- a/graphql/benchmark_test.go
+++ b/graphql/benchmark_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func pagintedQueryWithFilterBenchmark(b *testing.B, n int, batchFunc bool) {
+func pagintedQueryWithFilterBenchmark(b *testing.B, n int, batchFunc bool, expensive bool) {
 	schema := schemabuilder.NewSchema()
 	type Inner struct {
 	}
@@ -27,20 +27,38 @@ func pagintedQueryWithFilterBenchmark(b *testing.B, n int, batchFunc bool) {
 		items[i] = Item{Id: int64(i), FilterText: filterTexts[i%5], String: "a"}
 	}
 	if !batchFunc {
-		inner.FieldFunc("innerConnectionWithFilter", func() []Item {
-			return items
-		}, schemabuilder.Paginated,
-			schemabuilder.FilterField("foo",
-				func(ctx context.Context, i Item) string {
-					return i.FilterText
-				},
-			),
-			schemabuilder.FilterField("bar",
-				func(ctx context.Context, i Item) string {
-					return i.String
-				},
-			),
-		)
+		if expensive {
+			inner.FieldFunc("innerConnectionWithFilter", func() []Item {
+				return items
+			}, schemabuilder.Paginated,
+				schemabuilder.FilterField("foo",
+					func(ctx context.Context, i Item) string {
+						return i.FilterText
+					},
+				),
+				schemabuilder.FilterField("bar",
+					func(ctx context.Context, i Item) string {
+						return i.String
+					},
+				),
+			)
+		} else {
+			inner.FieldFunc("innerConnectionWithFilter", func() []Item {
+				return items
+			}, schemabuilder.Paginated,
+				schemabuilder.FilterField("foo",
+					func(i Item) string {
+						return i.FilterText
+					},
+				),
+				schemabuilder.FilterField("bar",
+					func(i Item) string {
+						return i.String
+					},
+				),
+			)
+		}
+
 	} else {
 		inner.FieldFunc("innerConnectionWithFilter", func() []Item {
 			return items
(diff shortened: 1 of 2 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -433,65 +433,103 @@ func (c *connectionContext) applyBatchTextFilter(ctx context.Context, nodes []in
...
+	}
+	return nodesToKeep, nil
+}
+
+func (c *connectionContext) applyTextFilterNotBatched(ctx context.Context, nodes []interface{}, matchStrings []string, filterFields map[string]*graphql.Field) ([]bool, error) {
+	nodesToKeep := make([]bool, len(nodes))
+	if len(filterFields) > 0 {
+		for unscopedI, unscopedNode := range nodes {
+			i, node := unscopedI, unscopedNode
+			keep, _ := c.checkFilters(ctx, node, matchStrings, filterFields)
```
wedaly: should we propagate any errors to the caller here?

wedaly:
I'm also a little confused by the benchmark numbers.  If I'm reading those right, it looks like the expensive version is always slower than the non-expensive version?  Is that what you and Jelle found on Friday?  (I recall him saying that parallelizing the field funcs was actually a little bit faster, but not worth the memory overhead)

More generally, is there ever a case where we'd want developers to use the "expensive" version rather than the batched version?  I'd think we'd want to enforce *some* limit on the number of goroutines to avoid CloudServer out-of-memory crashes?

graphql/schemabuilder/pagination.go
```diff
@@ -433,65 +433,103 @@ func (c *connectionContext) applyBatchTextFilter(ctx context.Context, nodes []in
...
+			keep = true
+			break
+		}
+	}
+	return keep, nil
+}
+
+func (c *connectionContext) applyTextFilterNotBatchedExpensive(ctx context.Context, nodes []interface{}, matchStrings []string, filterFields map[string]*graphql.Field) ([]bool, error) {
+	g, ctx := errgroup.WithContext(ctx)
+	nodesToKeep := make([]bool, len(nodes))
```
wedaly: I noticed that we're using a mutex in `applyBatchTextFilter`, but not in this function.  I think maybe the mutex is leftover from when we were using a map rather than a slice?

CodeBrew28:
The expensive version has the overhead of spawning many goroutines. For filters that make db calls/grpc requests, it makes more sense to use the expensive version. It also currently uses the logic that will mark it as expensive if the context is passed through, but the PR https://github.com/samsarahq/thunder/pull/281 should make it more explicit. 

The benchmarks currently pass in the context, so it is expensive, but don't actually do anything it expensive. It shows the overhead cost of spawning that many goroutines.

graphql/schemabuilder/pagination.go
CodeBrew28: for batching you need the mutex since we make multiple changes to the same index in nodesToKeep, but for both applyTextFilterNotBatchedExpensive and applyTextFilterNotBatched we only touch each index once, so we don't need the mutex lock

wedaly:
> More generally, is there ever a case where we'd want developers to use the "expensive" version rather than the batched version? I'd think we'd want to enforce _some_ limit on the number of goroutines to avoid CloudServer out-of-memory crashes?

Discussed offline: we'll keep the expensive option for now for backwards compatibility with existing (non-batched) field funcs.  Eventually we'll want to convert everything to either non-expensive (serial) or batched.

graphql/schemabuilder/pagination.go
wedaly: Ah, that makes sense.  Thanks!

graphql/benchmark_test.go
```diff
@@ -27,20 +27,38 @@ func pagintedQueryWithFilterBenchmark(b *testing.B, n int, batchFunc bool) {
...
-				func(ctx context.Context, i Item) string {
-					return i.String
-				},
-			),
-		)
+		if expensive {
+			inner.FieldFunc("innerConnectionWithFilter", func() []Item {
+				return items
+			}, schemabuilder.Paginated,
+				schemabuilder.FilterField("foo",
```
jellevandenhooff: I like this but instead of inferring from context can we make this explicit?

CodeBrew28: Yup, will's PR #281 should do it automatically

CodeBrew28: So looking back I think #281  will enable the option to allow us to make it explicitly, and I can piggy back off the logic. I think the `FilterField` function in `graphql/schemabuilder/types.go` would just need to have an argument to pass the `expensive` flag in, and it should work

PR #286, opened 2019-07-15
Batching Paginated Sorts

CodeBrew28:
Batches the paginated sorts by 
(1) Run the paginated sorts batched
(2) Run the paginated sorts not batched and executed serially, which is useful if we aren't making expensive operations
(3) Run the  paginated sorts not batched but with an expensive flag to spawn a goroutine per node. It detects if it is expensive if it has a ctx passed in. 

 
```
BenchmarkSortsBatched10Items                       50000             39933 ns/op           15834 B/op        246 allocs/op
BenchmarkSortsNotBatched10Items                    50000             29063 ns/op           12065 B/op        213 allocs/op
BenchmarkSortsNotBatchedExpensive10Items           50000             34620 ns/op           12385 B/op        213 allocs/op
BenchmarkSortsBatched100Items                      10000            116631 ns/op           54667 B/op        852 allocs/op
BenchmarkSortsNotBatched100Items                   10000            114231 ns/op           47156 B/op        939 allocs/op
BenchmarkSortsNotBatchedExpensive100Items          10000            161803 ns/op           50356 B/op        939 allocs/op
BenchmarkSortsBatched1000Items                      2000           1064689 ns/op          505393 B/op       7156 allocs/op
BenchmarkSortsNotBatched1000Items                   2000           1119087 ns/op          379250 B/op       8145 allocs/op
BenchmarkSortsNotBatchedExpensive1000Items          1000           1555780 ns/op          411252 B/op       8145 allocs/op
```

For reference when looking at the stats, the BenchmarkSortsNotBatchedExpensive numbers show the how long it use to take. Ideally everything will move to BenchmarkSortsBatched or BenchmarkSortsNotBatched and BenchmarkSortsNotBatchedExpensive will be deprecated

### DIFF
--- a/graphql/connection_test.go
+++ b/graphql/connection_test.go
@@ -90,17 +90,24 @@ func TestConnection(t *testing.T) {
 			{Id: 4, Number: 2, String: "2", Float: 2.0},
 			{Id: 5, Number: 4, String: "4", Float: 4.0},
 		}
-	}, schemabuilder.Paginated, schemabuilder.SortFields{
-		"numbers": func(ctx context.Context, i Item) int64 {
-			return i.Number
-		},
-		"strings": func(ctx context.Context, i *Item) string {
-			return i.String
-		},
-		"floats": func(ctx context.Context, i Item) float64 {
-			return i.Float
-		},
-	})
+	},
+		schemabuilder.Paginated,
+		schemabuilder.BatchSortField(
+			"numbers", func(ctx context.Context, items map[batch.Index]Item) (map[batch.Index]int64, error) {
+				myMap := make(map[batch.Index]int64, len(items))
+				for i, item := range items {
+					myMap[i] = item.Number
+				}
+				return myMap, nil
+			}),
+		schemabuilder.SortField(
+			"strings", func(ctx context.Context, i Item) string {
+				return i.String
+			}),
+		schemabuilder.SortField(
+			"floats", func(ctx context.Context, i Item) float64 {
+				return i.Float
+			}))
 	inner.FieldFunc("innerConnectionNilArg", func() []Item {
 		return []Item{{Id: 1}, {Id: 2}, {Id: 3}, {Id: 4}, {Id: 5}}
 	}, schemabuilder.Paginated)
@@ -581,9 +588,12 @@ func TestPaginateBuildFailure(t *testing.T) {
 		object.Key("id")
 		inner.FieldFunc("connection", func(ctx context.Context, args Args) ([]StructWithKey, error) {
 			return nil, nil
-		}, schemabuilder.Paginated, schemabuilder.SortFields{
-			"badReturn": func() struct{} { return struct{}{} },
-		})
+		},
+			schemabuilder.Paginated,
+			schemabuilder.SortField(
+				"badReturn",
+				func() struct{} { return struct{}{} },
+			))
 
 		_, err := schema.Build()
 		require.Error(t, err)
@@ -603,11 +613,11 @@ func TestPaginateBuildFailure(t *testing.T) {
 		object.Key("id")
 		inner.FieldFunc("connection", func(ctx context.Context, i *Inner, args Args) ([]StructWithKey, error) {
 			return nil, nil
-		}, schemabuilder.Paginated, schemabuilder.SortFields{
-			"someArgs": func(ctx context.Context, i *StructWithKey, args Args) string {
+		}, schemabuilder.Paginated, schemabuilder.SortField(
+			"someArgs", func(ctx context.Context, i *StructWithKey, args Args) string {
 				return ""
 			},
-		})
+		))
 
 		_, err := schema.Build()
 		require.Error(t, err)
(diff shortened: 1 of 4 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -559,23 +559,53 @@ func (c *connectionContext) applySort(ctx context.Context, nodes []interface{},
 	// node order.
 	sortValues := make([]sortReference, len(nodes))
 	g, ctx := errgroup.WithContext(ctx)
-
-	for unscopedI, unscopedNode := range nodes {
-		i, node := unscopedI, unscopedNode
+	if sortField.BatchResolver != nil {
```
willhug: do we want the feature flag here as well?

graphql/schemabuilder/pagination.go
```diff
@@ -676,6 +706,14 @@ func getFuncReturnType(fn interface{}) reflect.Type {
 	return typ.Out(0)
 }
 
+func getFuncReturnTypeBatched(fn interface{}) reflect.Type {
```
wedaly: is this called?  I can't find any other references to this function in the commit

graphql/schemabuilder/pagination.go
```diff
@@ -767,18 +807,47 @@ func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ refle
...
+			c.SortFunctions[name] = getSort(sortableTyp.Elem())
 		}
-		c.SortFunctions[name] = getSort(sortableTyp)
 
 		// Build a GraphQL field for the function.
-		field, err := sb.buildFunction(typ, &method{Fn: fn})
+		var field *graphql.Field = nil
+		var err error = nil
+		if fn.BatchSortFunc != nil && fn.SortFunc != nil {
+			field, err = sb.buildBatchFunction(typ,
```
wedaly: should we be using `buildBatchFunctionWithFallback` somewhere?

graphql/schemabuilder/types.go
```diff
@@ -101,17 +101,71 @@ func BatchFilterFieldWithFallback(name string, batchFilter interface{}, filter i
...
+	var sortStruct = Sort{
+		Name:          name,
+		BatchSortFunc: batchSort,
+	}
+	var fieldFuncSortFields fieldFuncOptionFunc = func(m *method) {
+		if m.SortFuncs == nil {
+			m.SortFuncs = map[string]Sort{}
+		}
+		if filter, ok := m.SortFuncs[name]; ok {
+			panic("Sorts Filters have the same name: " + filter.Name)
```
wedaly: this is a sort, not a filter?

graphql/schemabuilder/types.go
```diff
@@ -213,7 +267,7 @@ type method struct {
 	TextFilterFuncs map[string]Filter
 
 	// Sort methods
-	SortFuncs map[string]interface{}
+	SortFuncs map[string]Sort
```
wedaly: Maybe consider renaming this: it's a little misleading to call this `SortFuncs` when it actually points to a `Sort` structure rather than the function.

graphql/schemabuilder/pagination.go
```diff
@@ -709,6 +747,8 @@ func getSort(typ reflect.Type) func([]sortReference, SortOrder) {
 		return sorts[reflect.Float64]
 	case reflect.String:
 		return sorts[reflect.String]
+	case reflect.Map:
```
willhug: why is map supported for sorts?

graphql/schemabuilder/pagination.go
```diff
@@ -767,18 +807,47 @@ func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ refle
...
+					sortableTyp,
+				)
+			}
+			c.SortFunctions[name] = getSort(sortableTyp.Elem())
 		}
-		c.SortFunctions[name] = getSort(sortableTyp)
 
 		// Build a GraphQL field for the function.
-		field, err := sb.buildFunction(typ, &method{Fn: fn})
+		var field *graphql.Field = nil
```
willhug: we can just define
```
var field *graphql.Field
var err error
```

graphql/schemabuilder/pagination.go
CodeBrew28: good catch!

graphql/schemabuilder/pagination.go
```diff
@@ -555,27 +569,43 @@ func (c *connectionContext) applySort(ctx context.Context, nodes []interface{},
...
-	// sortValues is the slice we'll be sorting (with the sorted values) in order to figure out
-	// node order.
+	// sortValues is the slice we'll be sorting (with the sorted values) in order to figure out node order.
 	sortValues := make([]sortReference, len(nodes))
 	g, ctx := errgroup.WithContext(ctx)
-
-	for unscopedI, unscopedNode := range nodes {
-		i, node := unscopedI, unscopedNode
+	if sortField.Batch && sortField.UseBatchFunc(ctx) {
 		g.Go(func() error {
```
willhug: don't need a goroutine for the batched version (there's only one field).

graphql/schemabuilder/pagination.go
```diff
@@ -766,19 +796,57 @@ func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ refle
...
-	for name, fn := range m.SortFuncs {
-		sortableTyp := getFuncReturnType(fn)
-		if !supportedSort(sortableTyp) {
-			return fmt.Errorf(
-				"invalid sort field %s: unsupported return type %v, must be of kind int, uint, float or string",
-				name,
-				sortableTyp,
-			)
+	for name, fn := range m.SortFuncsStruct {
+		if fn.SortFunc != nil {
```
willhug: can we put all of this into a `getFieldForSortStruct` function?  This for loop is pretty complicated. (Possible for follow-up)

graphql/schemabuilder/types.go
```diff
@@ -97,21 +97,75 @@ func BatchFilterFieldWithFallback(name string, batchFilter interface{}, filter i
...
+		if filter, ok := m.SortFuncsStruct[name]; ok {
+			panic("Batch sorts fields have the same name: " + filter.Name)
+		}
+		m.SortFuncsStruct[name] = sortStruct
+	}
+	return fieldFuncSortFields
+}
+
+func BatchSortFieldWithFallback(name string, batchSort interface{}, sort interface{}, flag func(context.Context) bool) FieldFuncOption {
+	var sortStruct = Sort{
```
willhug: wait, should these just return `&method{` structs?  Is there any reason for us to hold onto a specific `Sort{` struct? (maybe follow-up)

graphql/connection_test.go
```diff
@@ -966,3 +976,151 @@ func TestPaginatedFilters(t *testing.T) {
...
+				return item.Number, nil
+			},
+			func(context.Context) bool {
+				return false
+			},
+		),
+	)
+
+	builtSchema := schema.MustBuild()
+	querries := []*graphql.Query{
```
willhug: arrrrrrrrrrrrr

willhug:
Oh, can we add a test for the Flagging logic, (e.g. batch fails, regular succeeds, flag on/off)

CodeBrew28:
@willhug they're already tests I added for this in connection_test.go, I added a one more for fallback functions, but lmk if there are any edge cases you want me to test

graphql/schemabuilder/pagination.go
```diff
@@ -776,12 +812,45 @@ func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ refle
 			)
 		}
 		c.SortFunctions[name] = getSort(sortableTyp)
+	}
+
+	if sortMethod.Batch == true {
```
willhug: if/else?

PR #287, opened 2019-07-16
Paginated Filters Optimization

CodeBrew28:
Parallelizes the processing of batched vs nonbatched vs expensive. The use case is if you have multiple batch filters and multiple nonbatched filters, it shouldn't be waiting for one to finish before starting the other. They should be executed in parallel

I created a field func with 2 batched filters and 2 nonbatched filters to see the impact
Before
```
BenchmarkFilters10Items            30000             67672 ns/op           19236 B/op        300 allocs/op
BenchmarkFilters100Items            5000            277301 ns/op           97238 B/op       1396 allocs/op
BenchmarkFilters1000Items           1000           2410648 ns/op         1026338 B/op      12207 allocs/op
```

After 
```
BenchmarkFilters10Items            30000             52655 ns/op           18434 B/op        291 allocs/op
BenchmarkFilters100Items            5000            246871 ns/op           96384 B/op       1386 allocs/op
BenchmarkFilters1000Items           1000           2124011 ns/op         1026218 B/op      12210 allocs/op
```

### DIFF
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -460,15 +460,13 @@ func (c *connectionContext) checkFilters(ctx context.Context, node interface{}, 
 func (c *connectionContext) applyTextFilterNotBatchedExpensive(ctx context.Context, nodes []interface{}, matchStrings []string, filterFields map[string]*graphql.Field) ([]bool, error) {
 	g, ctx := errgroup.WithContext(ctx)
 	nodesToKeep := make([]bool, len(nodes))
-	if len(filterFields) > 0 {
-		for unscopedI, unscopedNode := range nodes {
-			i, node := unscopedI, unscopedNode
-			g.Go(func() error {
-				keep, err := c.checkFilters(ctx, node, matchStrings, filterFields)
-				nodesToKeep[i] = keep
-				return err
-			})
-		}
+	for unscopedI, unscopedNode := range nodes {
+		i, node := unscopedI, unscopedNode
+		g.Go(func() error {
+			keep, err := c.checkFilters(ctx, node, matchStrings, filterFields)
+			nodesToKeep[i] = keep
+			return err
+		})
 	}
 	if err := g.Wait(); err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -514,22 +514,22 @@ func (c *connectionContext) applyTextFilter(ctx context.Context, nodes []interfa
...
-	}
-
-	batchedNodesToKeep, err := c.applyBatchTextFilter(ctx, nodes, matchStrings, filterTextFieldsBatched)
-	if err != nil {
-		return nil, err
-	}
-
+	nodesToKeep := make([]bool, len(nodes))
+	expensiveNodesToKeep := make([]bool, len(nodes))
+	batchedNodesToKeep := make([]bool, len(nodes))
```
wedaly: It looks like we're allocating memory for these slices twice, once here and once in the helper functions?  Probably we can just declare these:
```
var nodesToKeep []bool
var expensiveNodesToKeep []bool
var batchedNodesToKeep []bool
```

graphql/schemabuilder/pagination.go
```diff
@@ -514,22 +514,22 @@ func (c *connectionContext) applyTextFilter(ctx context.Context, nodes []interfa
...
-
-	batchedNodesToKeep, err := c.applyBatchTextFilter(ctx, nodes, matchStrings, filterTextFieldsBatched)
-	if err != nil {
-		return nil, err
-	}
-
+	var nodesToKeep []bool
+	var expensiveNodesToKeep []bool
+	var batchedNodesToKeep []bool
+	var err error = nil
```
wedaly: I think it's a race condition to have three goroutines updating the same error variable?  (Also you don't need `= nil` here because the zero-value of `error` is `nil`)

graphql/schemabuilder/pagination.go
CodeBrew28: hm thought this would work, but it expects that array to be a certain length and gets out of index errors otherwise. However, the maps are populated since the filtering tests work correctly.

wedaly: maybe I'm misunderstanding how Go slice assignment works... I would have thought that line 519 would copy the slice data to the `nodesToKeep` value, overwriting the pointer.  So the old memory would be garbage-collected, and the `nodesToKeep` value would point to the memory allocated on line 462.  Is that not the case?

wedaly: discussed and resolved offline

PR #288, opened 2019-07-16
Error Handling for Batched Pagination Filters

CodeBrew28:
(1) Builds the batch function with fallbacks correctly for paginated filters. This also means setting it to be nonullable in all cases
(2) Uses the SafeExecuteBatchResolver and SafeExecuteResolver for pagination to handle panics better 
(3) Allows you to pass in optional arguments for options when working on a filter. This is useful for the BatchFilterFieldWithFallback where you can set it to Expensive

### DIFF
--- a/graphql/schemabuilder/batch.go
+++ b/graphql/schemabuilder/batch.go
@@ -20,7 +20,6 @@ func (sb *schemaBuilder) buildBatchFunctionWithFallback(typ reflect.Type, m *met
 	if err != nil {
 		return nil, err
 	}
-
 	if fallbackFuncCtx.hasContext != batchFuncCtx.hasContext ||
 		!fallbackFuncCtx.hasSource || // Batch func always has a source.
 		fallbackFuncCtx.hasArgs != batchFuncCtx.hasArgs ||
@@ -29,6 +28,7 @@ func (sb *schemaBuilder) buildBatchFunctionWithFallback(typ reflect.Type, m *met
 		fallbackFuncCtx.hasRet != batchFuncCtx.hasRet {
 		return nil, fmt.Errorf("batch and fallback function signatures did not match")
 	}
+
 	if fallbackField.Type.String() != batchField.Type.String() {
 		return nil, fmt.Errorf("batch and fallback graphql return types did not match: Batch(%v) Fallback(%v)", batchField.Type, fallbackField.Type)
 	}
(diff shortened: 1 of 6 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -810,19 +810,35 @@ func (c *connectionContext) consumeTextFilters(sb *schemaBuilder, m *method, typ
...
-					Fn: fn.BatchFilterFunc,
-					BatchArgs: batchArgs{
-						FallbackFunc:          fn.FilterFunc,
-						ShouldUseFallbackFunc: fn.FallbackFlag,
-					}, Batch: true})
+		// Apply all options to the method that are passed in when the filter in instantiated.
+		for _, opt := range fn.Options {
+			opt.apply(m)
+		}
+		m.MarkedNonNullable = true
```
wedaly: nit: maybe a comment explaining why nullable fields on a filter don't make sense

graphql/schemabuilder/batch.go
```diff
@@ -29,7 +29,8 @@ func (sb *schemaBuilder) buildBatchFunctionWithFallback(typ reflect.Type, m *met
 		fallbackFuncCtx.hasRet != batchFuncCtx.hasRet {
 		return nil, fmt.Errorf("batch and fallback function signatures did not match")
 	}
-	if fallbackField.Type.String() != batchField.Type.String() {
+
+	if strings.TrimSuffix(fallbackField.Type.String(), "!") != batchField.Type.String() {
```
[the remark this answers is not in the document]
CodeBrew28: It's from the graphql types which means the field is nonullable

graphql/schemabuilder/types.go
```diff
@@ -85,12 +88,13 @@ func BatchFilterField(name string, batchFilter interface{}) FieldFuncOption {
 	return fieldFuncTextFilterFields
 }
 
-func BatchFilterFieldWithFallback(name string, batchFilter interface{}, filter interface{}, flag func(context.Context) bool) FieldFuncOption {
+func BatchFilterFieldWithFallback(name string, batchFilter interface{}, filter interface{}, flag func(context.Context) bool, options ...FieldFuncOption) FieldFuncOption {
```
willhug: Do we still need the fieldFuncOptions?  Can we get rid of them?

CodeBrew28: @wedaly was working on a PR to add a tag to filter funcs that were expensive instead of inferring that from if the context is passed in. Since we also want the ability to add the expensive tag to filter fields, I was going to leave it in. 

(This is the PR for reference https://github.com/samsarahq/thunder/pull/281, it will get re-merged after this PR lands)

PR #291, opened 2019-07-19
Refactoring Paginated Filters

CodeBrew28:
Previously we made a filter struct to pass information, but realized we didn't actually need it. Same functionality, just makes the code easier to read.

### DIFF
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -861,54 +861,47 @@ func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ refle
 	return nil
 }
 
+func (c *connectionContext) checkFilterTextFunctionTypes(name string, filterMethod *method) error {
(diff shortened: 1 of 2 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -861,54 +861,48 @@ func (c *connectionContext) consumeSorts(sb *schemaBuilder, m *method, typ refle
...
-
-		// Build the function from the method.
-		if fn.BatchFilterFunc != nil && fn.FilterFunc != nil && fn.FallbackFlag != nil {
-			field, err = sb.buildBatchFunctionWithFallback(typ, m)
-		} else if fn.FilterFunc != nil {
-			field, err = sb.buildFunction(typ, m)
-		} else if fn.BatchFilterFunc != nil {
-			field, err = sb.buildBatchFunction(typ, m)
+
+		var field *graphql.Field = nil
```
wedaly: nit: the zero-value of a pointer is always `nil`, so you can remove the `= nil`

PR #294, opened 2019-07-25
Expensive field func option take two

wedaly:
Already reviewed in #281, but worth another review to make sure it works correctly with the recent pagination improvements.


CodeBrew28:
lgtm! There's one test failing in testLimit, but its a bit of a flaky test that works sometimes (might be worth considering rewritting or removing it, but for now you can just rerun the tests). 

just a heads up @willhug has a PR to pull thunder into backend for some of the work he's doing in https://github.com/samsara-dev/backend/pull/30872

PR #295, opened 2019-07-26
Work Scheduler for pagination

CodeBrew28:
All goroutines spawned by filters and sorts now use the work scheduler, which allows us to track the goroutines spawned. This also exposes the context, which can be used for xray traces.

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -10,6 +10,14 @@ import (
 	"github.com/samsarahq/thunder/reactive"
 )
 
+type executeFunction func(ctx context.Context) error
+
+type ExecutionUnit struct {
+	Context         context.Context
+	ExeucteFunction executeFunction
+	Fields          map[string]*Field
+}
+
 // WorkUnit is a set of execution work that will be done when running
 // a graphql query.  For every source there is an equivalent destination
 // OutputNode that is used to record the result of running a section of the
@@ -52,6 +60,7 @@ type UnitResolver func(*WorkUnit) []*WorkUnit
 // work unit.
 type WorkScheduler interface {
 	Run(resolver UnitResolver, startingUnits ...*WorkUnit)
+	RunAll(ctx context.Context, executionUnits []*ExecutionUnit)
 }
 
 func NewExecutor(scheduler WorkScheduler) ExecutorRunner {
@@ -80,6 +89,7 @@ func (e *Executor) Execute(ctx context.Context, typ Type, source interface{}, qu
 	topLevelRespWriter := newTopLevelOutputNode(query.Name)
 	initialSelectionWorkUnits := make([]*WorkUnit, 0, len(topLevelSelections))
 	writers := make(map[string]*outputNode)
+	ctx = context.WithValue(ctx, "pagiantionScheduler", e.scheduler)
 	for _, selection := range topLevelSelections {
 		field, ok := queryObject.Fields[selection.Name]
 		if !ok {
@@ -423,7 +433,6 @@ func resolveObjectBatch(ctx context.Context, sources []interface{}, typ *Object,
 			destinations: destForSelection,
 			selection:    selection,
 		}
-
 		switch {
 		case shouldUseBatch(ctx, field):
 			unit.useBatch = true
--- a/graphql/batch_executor_test.go
+++ b/graphql/batch_executor_test.go
@@ -275,15 +275,16 @@ func TestNonExpensiveExecution(t *testing.T) {
 }
 
 type counterGoroutineScheduler struct {
-	wg sync.WaitGroup
-
+	wg    sync.WaitGroup
 	count int64
 }
 
 func (q *counterGoroutineScheduler) Run(resolver graphql.UnitResolver, initialUnits ...*graphql.WorkUnit) {
 	q.runEnqueue(resolver, initialUnits...)
-
 	q.wg.Wait()
+}
+
+func (q *counterGoroutineScheduler) RunAll(ctx context.Context, executionUnits []*graphql.ExecutionUnit) {
 }
 
 func (q *counterGoroutineScheduler) runEnqueue(resolver graphql.UnitResolver, units ...*graphql.WorkUnit) {
--- a/graphql/batch_scheduler.go
+++ b/graphql/batch_scheduler.go
@@ -1,7 +1,9 @@
 package graphql
 
 import (
+	"context"
 	"sync"
+	"sync/atomic"
 )
 
 // NewImmediateGoroutineScheduler creates a new batch execution scheduler that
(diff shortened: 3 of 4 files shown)


graphql/batch_executor.go
```diff
@@ -80,6 +89,7 @@ func (e *Executor) Execute(ctx context.Context, typ Type, source interface{}, qu
 	topLevelRespWriter := newTopLevelOutputNode(query.Name)
 	initialSelectionWorkUnits := make([]*WorkUnit, 0, len(topLevelSelections))
 	writers := make(map[string]*outputNode)
+	ctx = context.WithValue(ctx, "pagiantionScheduler", e.scheduler)
```
wedaly: nit: small typo here "pagiantionScheduler" -> "paginationScheduler"
Also, consider making this string a `const` to ensure that all callers are using the exact same string.

graphql/batch_executor_test.go
```diff
@@ -275,17 +275,18 @@ func TestNonExpensiveExecution(t *testing.T) {
...
 }
 
 func (q *counterGoroutineScheduler) Run(resolver graphql.UnitResolver, initialUnits ...*graphql.WorkUnit) {
 	q.runEnqueue(resolver, initialUnits...)
-
 	q.wg.Wait()
 }
 
+func (q *counterGoroutineScheduler) RunAll(ctx context.Context, executionUnits []*graphql.ExecutionUnit) {
+}
```
wedaly: Maybe a comment here explaining that this does nothing because the test doesn't need it?

graphql/batch_scheduler.go
```diff
@@ -10,15 +12,33 @@ func NewImmediateGoroutineScheduler() WorkScheduler {
...
-
 	r.wg.Wait()
 }
 
+func (q *immediateGoroutineScheduler) RunAll(ctx context.Context, executionUnits []*ExecutionUnit) {
+	q.runEnqueue(executionUnits)
+	q.wg.Wait()
+}
+
+func (q *immediateGoroutineScheduler) runEnqueue(executionUnits []*ExecutionUnit) {
```
wedaly: This is a little confusing: there are two `runEnqueue` methods, one on the scheduler and one on the runner.  It seems like maybe the responsibilities between these two structs are blurring a little bit?

graphql/batch_executor.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/samsarahq/thunder/reactive"
 )
 
+type executeFunction func(ctx context.Context) error
+
+type ExecutionUnit struct {
+	Context         context.Context
+	ExeucteFunction executeFunction
```
wedaly: nit: small typo here "ExeucteFunction" -> "ExecuteFunction"

graphql/batch_scheduler.go
wedaly: Maybe we should rename the methods to `runEnqueueWorkUnits` and `runEnqueueExecutionUnits`, then move the latter to `immediateGoroutineSchedulerRunner`.

graphql/batch_executor.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/samsarahq/thunder/reactive"
 )
 
+type executeFunction func(ctx context.Context) error
+
+type ExecutionUnit struct {
```
wedaly: I think a comment would be helpful here.  What is the difference between a work unit and an execution unit?  When would I need to use the former vs the latter?

graphql/schemabuilder/pagination.go
```diff
@@ -401,35 +402,38 @@ type SafeBatchNodesToKeep struct {
 	mux         sync.Mutex
 }
 
-func (c *connectionContext) applyBatchTextFilter(ctx context.Context, nodes []interface{}, matchStrings []string, batchedFields map[string]*graphql.Field, nodesToKeep []bool) error {
-	g, ctx := errgroup.WithContext(ctx)
+func (c *connectionContext) applyBatchTextFilter(ctx context.Context, nodes []interface{}, matchStrings []string, batchedFields map[string]*graphql.Field, nodesToKeep []bool) []*graphql.ExecutionUnit {
 	m := &sync.Mutex{}
+	executors := []*graphql.ExecutionUnit{}
```
wedaly: nit: maybe call this `executionUnits`?  There's another struct called an executor, so naming this executors is a little confusing.

graphql/schemabuilder/pagination.go
```diff
@@ -511,21 +530,18 @@ func (c *connectionContext) applyTextFilter(ctx context.Context, nodes []interfa
...
-		})
+		executors = append(executors, c.applyTextFilterNotBatchedExpensive(ctx, nodes, matchStrings, filterTextFieldsNotBatchedExpensive, expensiveNodesToKeep)...)
 	}
 	if len(filterTextFieldsBatched) > 0 {
-		g.Go(func() error {
-			return c.applyBatchTextFilter(ctx, nodes, matchStrings, filterTextFieldsBatched, batchedNodesToKeep)
-		})
+		executors = append(executors, c.applyBatchTextFilter(ctx, nodes, matchStrings, filterTextFieldsBatched, batchedNodesToKeep)...)
 	}
+	ctx.Value("pagiantionScheduler").(graphql.WorkScheduler).RunAll(ctx, executors)
```
wedaly: This will panic if the value isn't set or the type conversion fails.  Could you maybe add a comment explaining that this is always set, so it's safe?

graphql/schemabuilder/pagination.go
```diff
@@ -603,6 +627,7 @@ func (c *connectionContext) applySort(ctx context.Context, nodes []interface{},
 				sortValues[i] = *sortValue
 			}
 		}
+		ctx.Value("pagiantionScheduler").(graphql.WorkScheduler).RunAll(ctx, executors)
```
wedaly: Maybe pull this into a helper function to avoid duplicating code (line 543)

graphql/schemabuilder/pagination.go
```diff
@@ -621,14 +646,46 @@ func (c *connectionContext) applySort(ctx context.Context, nodes []interface{},
...
+	for _, unit := range executionUnits {
+		q.wg.Add(1)
+		go func(u *graphql.ExecutionUnit) {
+			defer q.wg.Done()
+			u.ExeucteFunction(u.Context)
+		}(unit)
+	}
+}
+
+func getScheduler(ctx context.Context) graphql.WorkScheduler {
```
wedaly: Can't seem to find where this is used?  Could we remove it?

graphql/batch_executor.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/samsarahq/thunder/reactive"
 )
 
+type executeFunction func(ctx context.Context) error
```
willhug: Need to make this a public function I think.  We currently have a public field that has a private type.

graphql/batch_executor.go
```diff
@@ -10,6 +10,14 @@ import (
...
+
+type ExecutionUnit struct {
+	Context         context.Context
+	ExeucteFunction executeFunction
+	Fields          map[string]*Field
```
willhug: What are the fields in this case?  Shouldn't it only be one field?

graphql/batch_executor.go
```diff
@@ -52,6 +60,7 @@ type UnitResolver func(*WorkUnit) []*WorkUnit
 // work unit.
 type WorkScheduler interface {
 	Run(resolver UnitResolver, startingUnits ...*WorkUnit)
+	RunAll(ctx context.Context, executionUnits []*ExecutionUnit)
```
willhug: Why does this function take a context, we don't use it.  I think the contexts on the execution Units should be enough.

graphql/schemabuilder/pagination.go
```diff
@@ -401,35 +402,38 @@ type SafeBatchNodesToKeep struct {
...
-			return nil
-		})
-	}
 
-	if err := g.Wait(); err != nil {
-		return err
+		executors = append(executors,
+			&graphql.ExecutionUnit{
+				Context: ctx,
+				Fields:  batchedFields,
```
willhug: In this context how do I know which of the fields I'm running?

PR #297, opened 2019-08-02
go reactive: Cancel context after each reactive Run

willhug:
Summary: Some pieces of code will hold onto the context waiting for it
to be done (usually in a goroutine).  This means that all the objects
that are stored on the context are held until the original context is
marked as "done".  This would not happen until the websocket/http
request finishes.  So for continous requests we have a leak where each
intermediary context is holding onto the entire cache stack (and
anything else on the context) until we reset the websocket connection.

This commit adds a cancel to context for each reactive.Run.  This should
allow all code that depends on the context.Done() to be able to clear/reset
between runs.

### DIFF
--- a/reactive/rerunner.go
+++ b/reactive/rerunner.go
@@ -369,7 +369,13 @@ func (r *Rerunner) run() {
 		time.Sleep(WriteThenReadDelay)
 	}
 	r.cache.cleanInvalidated()
-	ctx := context.WithValue(r.ctx, cacheKey{}, r.cache)
+
+	// Cancel the context passed to "run". Canceling the context ensures that
+	// libraries let go of the context when they might otherwise hold onto it long
+	// past this run.
+	ctx, cancel := context.WithCancel(r.ctx)
+	defer cancel()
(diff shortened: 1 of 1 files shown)


willhug:
Done

reactive/rerunner.go
```diff
@@ -369,7 +369,9 @@ func (r *Rerunner) run() {
 		time.Sleep(WriteThenReadDelay)
 	}
 	r.cache.cleanInvalidated()
-	ctx := context.WithValue(r.ctx, cacheKey{}, r.cache)
+	ctx, cancel := context.WithCancel(r.ctx)
```
ccfrost: Please add a brief comment that describes why this is here.

willhug: Done

ccfrost: Thanks!
Grammar: I don't know how to read the first sentence. Between "context we" I am unable to follow the connection.
How about, "Cancel the context passed to "run". Canceling the context ensures that libraries let go of the context when they might otherwise hold onto it long past this run."

PR #301, opened 2019-08-15
go graphql/schemabuilder: fix expensive option for BFF fallback

wedaly:
BatchFieldFuncWithFallback was dropping the Expensive flag
when building the fallback field.  Pass that flag down
so the fallback field respects it.

### DIFF
--- a/graphql/schemabuilder/batch.go
+++ b/graphql/schemabuilder/batch.go
@@ -11,7 +11,11 @@ import (
 
 // buildBatchFunction corresponds to buildFunction for a batchFieldFunc
 func (sb *schemaBuilder) buildBatchFunctionWithFallback(typ reflect.Type, m *method) (*graphql.Field, error) {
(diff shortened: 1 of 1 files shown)


graphql/schemabuilder/batch.go
```diff
@@ -11,7 +11,11 @@ import (
 
 // buildBatchFunction corresponds to buildFunction for a batchFieldFunc
 func (sb *schemaBuilder) buildBatchFunctionWithFallback(typ reflect.Type, m *method) (*graphql.Field, error) {
-	fallbackField, fallbackFuncCtx, err := sb.buildFunctionAndFuncCtx(typ, &method{Fn: m.BatchArgs.FallbackFunc, MarkedNonNullable: m.MarkedNonNullable})
+	fallbackField, fallbackFuncCtx, err := sb.buildFunctionAndFuncCtx(typ, &method{
```
jellevandenhooff: is there a way to make this a blacklist instead of a whitelist for options to pass through?

wedaly: hmm, that sounds do-able.  Can look into that after getting the expensive stuff working again :)

PR #313, opened 2019-09-03
Passing ObjectName correctly to subworkunits

CodeBrew28:
Object name now gets passed down. Also "nonexpensive" doesn't get added to contexts that are queries

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -44,6 +44,7 @@ func splitWorkUnit(unit *WorkUnit) []*WorkUnit {
 			sources:      []interface{}{source},
 			destinations: []*outputNode{unit.destinations[idx]},
 			selection:    unit.selection,
+			objectName:   unit.objectName,
 		})
 	}
 	return workUnits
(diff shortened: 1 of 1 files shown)


graphql/batch_executor.go
```diff
@@ -157,10 +158,11 @@ func executeBatchWorkUnit(unit *WorkUnit) []*WorkUnit {
 func executeNonExpensiveWorkUnit(unit *WorkUnit) []*WorkUnit {
 	results := make([]interface{}, 0, len(unit.sources))
 	for idx, src := range unit.sources {
+		ctx := unit.Ctx
 		if unit.objectName == "Query" {
```
willhug: !Mutation

graphql/batch_executor.go
```diff
@@ -157,10 +157,11 @@ func executeBatchWorkUnit(unit *WorkUnit) []*WorkUnit {
 func executeNonExpensiveWorkUnit(unit *WorkUnit) []*WorkUnit {
 	results := make([]interface{}, 0, len(unit.sources))
 	for idx, src := range unit.sources {
-		if unit.objectName == "Query" {
-			unit.Ctx = context.WithValue(unit.Ctx, nonExpensive{}, struct{}{})
+		ctx := unit.Ctx
+		if unit.objectName != "Mutation" {
```
willhug: did we still want to set the objectName "properly" everywhere.  I think right now it only gets set for the top-level "Mutation" and "Query" types.

CodeBrew28: we can, I just wasn't sure if we needed to pass that information down

willhug: we should add it back since right now objectName is not "really" what it says it is.  I'd rather get it done now instead of wondering why it wasn't set properly later.

Additionally can we add a comment here explaining what we're doing?

graphql/batch_executor.go
```diff
@@ -157,10 +158,14 @@ func executeBatchWorkUnit(unit *WorkUnit) []*WorkUnit {
 func executeNonExpensiveWorkUnit(unit *WorkUnit) []*WorkUnit {
 	results := make([]interface{}, 0, len(unit.sources))
 	for idx, src := range unit.sources {
-		if unit.objectName == "Query" {
-			unit.Ctx = context.WithValue(unit.Ctx, nonExpensive{}, struct{}{})
+		ctx := unit.Ctx
+
+		// ObjectName is initially either "query" or "mutation" for the initial work unit, and for later work units
+		// is the parents object name. We only want to exclude mutations on the initial work unit level.
```
willhug:
```
Fields on the Mutation object should not be marked as "non-Expensive" because they are guaranteed to only execute once.  The only fields we want to validate "expensiveness" on are non-Mutation Fields.
```

PR #314, opened 2019-09-10
Manual Pagination With Fallback

CodeBrew28:
For the Fast Pagination Indices project, we want to use manual pagination such that the index reader service can handle the pagination logic and only return the data going to be displayed. As endpoints get moved over to the indexing solution, we want a nice way to feature flag them. This means that we want to register both the field func (that uses indices and is manually paginated) and the fallback, which requires a thunder change. 

RFC reference
https://paper.dropbox.com/doc/RFC-Pagination-Indices--Ajk_EnByDy_fw8S6SrVjgxhAAg-E948ZaMsM7Ls6ZAWtpT5e

### DIFF
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -959,7 +959,6 @@ func (c *connectionContext) consumePaginatedArgs(sb *schemaBuilder, in []reflect
 		}
 
 	}
-
 	return argParser, argType, in, nil
 }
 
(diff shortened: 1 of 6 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -1022,50 +1021,116 @@ func (c *connectionContext) parsePaginatedReturnSignature(m *method) (err error)
...
+	if err != nil {
+		return nil, err
+	}
+
+	manualPaginationField, manualPaginationFuncCtx, err := sb.buildPaginatedFunctionAndFuncCtx(typ, m)
+	if err != nil {
+		return nil, err
+	}
+
+	// Validate that function signatures batch between the manually paginated version and fallback version
```
wedaly: s/batch/match

graphql/schemabuilder/pagination.go
```diff
@@ -1022,50 +1021,116 @@ func (c *connectionContext) parsePaginatedReturnSignature(m *method) (err error)
...
+		fallbackFuncCtx.hasError != manualPaginationFuncCtx.hasError ||
+		fallbackFuncCtx.hasRet != manualPaginationFuncCtx.hasRet {
+		return nil, fmt.Errorf("manual pagination and fallback function signatures did not match")
+	}
+
+	if fallbackField.Type.String() != manualPaginationField.Type.String() {
+		return nil, fmt.Errorf("manual pagination and fallback graphql return types did not match: ManualPagination(%v) Fallback(%v)", manualPaginationField.Type, fallbackField.Type)
+	}
+
+	if len(fallbackField.Args) != (len(manualPaginationField.Args)) {
```
wedaly: nit: can remove parens from `(len(manualPaginationField.Args))`?

PR #319, opened 2019-09-26
 go sqlgen: add dynamiclimit option to db object

alanefl:
Adds `WithDynamicLimit`, which functions the same as `WithShardLimit` but allows for user-provided callbacks to get what filter to apply and what to do on failure.

I broke up the `WithShardLimit` test into smaller chunks and moved them to `db_test.go` in order to test `WithDynamicLimit` and `WithShardLimit` (which should function the same beneath the hood) with the same test logic. I also added some unit tests specific to `WithDynamicLimit`.


sqlgen/db.go
```diff
@@ -100,35 +110,82 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
...
+// both user-provided callbacks.
+//
+// Before making an SQL query, a Thunder DB handle with a dynamic limit
+// will call `getLimitFilter` with the request context and the name of
+// the table being queried. It will use the sqlgen filter returned by
+// the function as the limit to check for against the current query.
+// The underlying limit check is the same as that used in `WithShardLimit`.
+//
+// If the limit check fails, Thunder will call `onError` with the resulting
+// error object.  If `onError` itself returns an error, the query will fail.
```
stephen: it took me a couple reads to understand what `onError` is for. consider a name like `shouldContinue` or similar?

sqlgen/db_test.go
```diff
@@ -0,0 +1,538 @@
...
+			"name": "Alice",
+		}
+	}
+	dynamicLimitErrorFunc := func(err error) error {
+		return err
+	}
+
+	defer tdb.Close()
+	aliceDb, err := db.WithDynamicLimit(dynamicLimitFunc, dynamicLimitErrorFunc)
+	if err != nil {
```
stephen: nit: `assert.NoError(t, err)`

sqlgen/db.go
```diff
@@ -18,6 +18,16 @@ type DB struct {
 
 	batchFetch *batch.Func
 	shardLimit Filter
+
+	dynamicLimit dynamicLimit
+}
+
+type dynamicLimitFilterCallback func(context.Context, string) Filter
```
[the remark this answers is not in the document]
alanefl: Yes, that should be the case.  Good eye.

sqlgen/db.go
alanefl: Going with `shouldContinueOnError`.  As a refresher, this can't just be a global bool setting because we'd like to be able to notify user code that their dynamic limit check failed.

sqlgen/db_test.go
alanefl: Updated the entire unit test to have better `assert` hygiene.

sqlgen/db.go
```diff
@@ -100,35 +110,82 @@ func NewDB(conn *sql.DB, schema *Schema) *DB {
...
+// will call `getLimitFilter` with the request context and the name of
+// the table being queried. It will use the sqlgen filter returned by
+// the function as the limit to check for against the current query.
+// The underlying limit check is the same as that used in `WithShardLimit`.
+//
+// If the limit check fails, Thunder will call `shouldContinueOnError` with
+// the resulting error object.  If `shouldContinueOnError` itself returns
+// an error, the query will fail. Otherwise, if the return value is nil Thunder
+// will still carry out the query.
+func (db *DB) WithDynamicLimit(getLimitFilter DynamicLimitFilterCallback, shouldContinueOnError DynamicLimitErrorCallback) (*DB, error) {
```
Xeroday: Might be cleaner to export the DynamicLimit struct and taking a DynamicLimit here instead of the 2 callbacks. 

The documentation for the callback functionality can then go in the DynamicLimit doc rather than the WithDynamicLimit doc.

alanefl: Yes, I agree this would be cleaner.

stephen: i like that passing the functions directly is unambiguous and you cannot mess up.

(i.e. harder to accidentally not pass one of the functions)

Xeroday: We could add that check in WithDynamicLimit, that way we could also remove the duplicate checks in checkFilterAgainstLimit(s)

stephen: sorry i thought about it for another second, i think i'd propose an api that looks like:
```
func ... WithDyanamicLimit(filterCreator DynamicLimitFilterCreator) {
  // ...
}

type DynamicLimitFilterCreator func(ctx context.Context, tableName string) (Filter, bool) # or second return value of type error
```

where a single callback sends both the filter and what to do if it fails. I think that makes the intent clear to the implementer how they should think about the inputs and outputs of the decision they need to write.

alanefl: Ahh, but then how would we notify user code of a filter check failure? We'd like this for people to be able to log failures, etc.

alanefl: We could have `DynamicLimitFilterCreator` return the filter and the error callback instead of a pure error or bool. All the logic is in one place, and we could have internal checks for a nil error callback.

alanefl: We're sticking with 

```
Might be cleaner to export the DynamicLimit struct and taking a DynamicLimit here instead of the 2 callbacks.

The documentation for the callback functionality can then go in the DynamicLimit doc rather than the WithDynamicLimit doc.
```

But I am adding the table name as an argument to the error callback.

PR #322, opened 2019-10-04
go graphql: Add ConcurrencyController option for batch and non-expensive execution

willhug:
Summary: Adds the ability to control the number of concurrent work units
we'll create for batch and non-expensive fields.  This allows us to control
how big batches get and allows us the ability to split these batches into
smaller sub-executions (sometimes we get diminishing returns from batches
after they get to a certain size).

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -41,12 +41,44 @@ func splitWorkUnit(unit *WorkUnit) []*WorkUnit {
 		workUnits = append(workUnits, &WorkUnit{
 			Ctx:          unit.Ctx,
 			field:        unit.field,
+			selection:    unit.selection,
 			sources:      []interface{}{source},
 			destinations: []*outputNode{unit.destinations[idx]},
-			selection:    unit.selection,
+			useBatch:     unit.useBatch,
 			objectName:   unit.objectName,
 		})
 	}
+	return workUnits
+}
+
+// Splits the work unit to N work units (based on configuration).
+func splitToNWorkUnits(unit *WorkUnit, numUnits int) []*WorkUnit {
+	if numUnits > len(unit.sources) {
+		numUnits = len(unit.sources)
+	}
+	if numUnits <= 0 {
+		numUnits = 1
+	}
+
+	avgUnitSize := (len(unit.sources) / numUnits) + 1
+	workUnits := make([]*WorkUnit, 0, numUnits)
+	for i := 0; i < numUnits; i++ {
+		workUnits = append(workUnits, &WorkUnit{
+			Ctx:          unit.Ctx,
+			field:        unit.field,
+			selection:    unit.selection,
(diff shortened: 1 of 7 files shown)


graphql/schemabuilder/types.go
```diff
@@ -299,6 +301,18 @@ type method struct {
 	ManualPaginationArgs manualPaginationArgs
 }
 
+type concurrencyArgs struct {
+	concurrencyFunc ConcurrencyController
+}
+
+// ConcurrencyController is a configuration option for non-expensive and batch
```
jellevandenhooff: Let's use a more self-describing name, eg. NumGoroutines() or NumParallelInvocations()?

Thinking about this some more, I think the API I would want to use is a MaxBatchSize()? I suspect if we default that to 1,000 it will do the right thing for pretty much most use cases.

willhug: Yeah, can change the name, NBD.

For the MaxBatchSize option.  It should be pretty trivial to write a internal function that uses the `ConcurrencyController` under the hood, but, can be embedded directly as `internalpackage.MaxBatchSize`.

jellevandenhooff:
In what case do we need something more fine grained than max batch size?
I’d rather not expose the knob if not necessary.

willhug:
The gps.Formatted first case is a good example.  We aren't bound by RPCs, we're bound by CPUs, if we make too many goroutines it actually decreases our performance a bit

jellevandenhooff:
Maybe! But in a multitenant situation the exact same problem applies, and
I’d rather rely on either the go or our work item scheduled to do something
sane.

Have you benchmarked this? If you have compelling data showing that a fixed
batch size performs worse than specifying a number of cores, I’m sold.

Maybe the way we figure that out is landing this code and experimenting. I
hope that whatever we learn as the best way to control concurrency we can
express in a easy to use and simple api and hide this complex knob.

willhug:
I had a parallelism test that compared 8 goroutines vs 100 goroutines for the gps.Formatted field, the 8 goroutines outperformed the 100 goroutines (600ms vs 700ms).

Also I think regardless we want to have a function instead of a static `MaxBucketSize` (so we can get some feature flagging goodness).

graphql/batch_executor.go
```diff
@@ -41,12 +41,44 @@ func splitWorkUnit(unit *WorkUnit) []*WorkUnit {
...
+	}
+	return workUnits
+}
+
+// Splits the work unit to N work units (based on configuration).
+func splitToNWorkUnits(unit *WorkUnit, numUnits int) []*WorkUnit {
+	if numUnits > len(unit.sources) {
+		numUnits = len(unit.sources)
+	}
+	if numUnits <= 0 {
```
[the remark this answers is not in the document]
willhug: yeah, users can provide negative or zero values in the provided function (e.g. a default integer feature flag value would be `0`)

PR #338, opened 2019-11-22
graphql: expose UnparsedArgs and ParentType for traceability

stephen:
Expose these two types on selections so that middleware can read
them for debugging/tracing.

### DIFF
--- a/graphql/executor.go
+++ b/graphql/executor.go
@@ -135,7 +135,7 @@ func PrepareQuery(ctx context.Context, typ Type, selectionSet *SelectionSet) err
 		}
 		for _, selection := range selectionSet.Selections {
 			if selection.Name == "__typename" {
-				if !isNilArgs(selection.Args) {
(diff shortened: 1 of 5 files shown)


graphql/executor.go
```diff
@@ -170,15 +170,17 @@ func PrepareQuery(ctx context.Context, typ Type, selectionSet *SelectionSet) err
 			}
 
 			// Only parse args once for a given selection.
-			if !selection.parsed {
+			if selection.UnparsedArgs == nil {
```
willhug: Is it possible for the args to be nil? Might make sense to keep the parsed flag.

In which case we could set the "unparsedArgs" wherever we're setting the "args" right now.

So we'd do
```
parsed, err := field.ParseArguments(selection.UnparsedArgs)
```
(which I think is a bit cleaner anyway)

stephen: thanks, i took your suggestion here.

PR #341, opened 2019-12-12
Pagination: add PostProcessOptions

wit221:
This adds a PostProcessOptions that externally managed paginated endpoints need to return and optionally fill options of. This lets Thunder perform additional operations on the output based on the turned-on flags.

The first flags, added in this PR, are `ApplyTextFilter`, which will apply a text filter on the output, and `SetPageInfo`, which will set the page info based on the output.

### DIFF
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -149,6 +149,20 @@ type PaginationInfo struct {
 	Pages          []string
 }
 
+// PostProcessOptions is used to instruct Thunder to perform additional operations on the output
+// in the case of manually paginated resolvers.
+// This struct should be returned as the second result from a manually-paginated field func.
+type PostProcessOptions struct {
+	// Whether or not Thunder should apply filtering on the output. This flag can
+	// be used by externally managed resolvers when they decide to fall back to Thunder's
+	// filtering based on the query arguments.
+	ApplyTextFilter bool
+	// Whether or not Thunder should set the Page Info's fields based on the output.
+	// An exception to this are start and end cursors, which will be set by Thunder
+	// regardless of the SetPageInfo flag.
+	SetPageInfo bool
+}
+
 func (i PaginationInfo) TotalCount() (int64, error) {
 	if i.TotalCountFunc == nil {
 		return 0, errors.New("must set TotalCountFunc on PaginationInfo")
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -1009,6 +1009,11 @@ func (c *connectionContext) parsePaginatedReturnSignature(m *method) (err error)
 	if len(out) > 0 && out[0] == reflect.TypeOf(PaginationInfo{}) {
 		c.ReturnsPageInfo = true
 		out = out[1:]
+		if len(out) == 0 || out[0] != reflect.TypeOf(PostProcessOptions{}) {
+			err = fmt.Errorf("%s returns a PaginationInfo not followed by a PaginationPostProcessOptions", c.funcType)
+			return
+		}
+		out = out[1:]
 	}
 
 	if len(out) > 0 && out[0] == errType {
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -183,6 +183,8 @@ type connectionContext struct {
 	Key string
 	// Whether or not the FieldFunc returns PageInfo (overrides thunder's auto-populated PageInfo).
 	ReturnsPageInfo bool
+	// The post process options returned by the resolver.
+	PostProcessOptions PostProcessOptions
 	// The index of PaginationArgs in the arguments provided to the FieldFunc.
 	PaginationArgsIndex int
 	// The GraphQL fields for filtered text to be resolved.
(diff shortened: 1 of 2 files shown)


graphql/schemabuilder/pagination.go
```diff
@@ -149,6 +149,14 @@ type PaginationInfo struct {
 	Pages          []string
 }
 
+// PostProcessOptions can be returned in a PaginateFieldFunc as an argument that succeeds
+// PaginationInfo. When a PaginationInfo is set, it must be succeeded by a PostProcessOptions.
+// The struct is used to instruct Thunder to perform additional operations on the output
+// in the case of manually paginated resolvers.
```
wedaly: nit: I feel that the last sentence here is the most important, since it explains what this struct means, so maybe that should be the first sentence in the comment.

graphql/schemabuilder/pagination.go
```diff
@@ -149,6 +149,14 @@ type PaginationInfo struct {
 	Pages          []string
 }
 
+// PostProcessOptions can be returned in a PaginateFieldFunc as an argument that succeeds
+// PaginationInfo. When a PaginationInfo is set, it must be succeeded by a PostProcessOptions.
```
wedaly: nit: I found this description a little confusing ("succeeds" has multiple meanings).   Maybe something like: "This struct should be returned as the second result from a manually-paginated field func"?

graphql/schemabuilder/pagination.go
```diff
@@ -664,7 +678,7 @@ func (c *connectionContext) getConnection(ctx context.Context, out []reflect.Val
 
 	// If the pagination is externally managed, thunder isn't going to handle setting page
 	// information or reducing the edges.
-	if c.IsExternallyManaged() {
+	if c.IsExternallyManaged() && !c.PostProcessOptions.ApplyTextFilter {
```
wedaly: I imagine if we add more postprocessing options that could affect pagination info in the future we'd have to remember to update this `if` statement.  Maybe it would be more maintainable to have this be calculated in a method on `PostProcessOptions`?  Imagining something like:
```
func (p *PostProcessOptions) SetPaginationInfo() bool {
    return p.ApplyTextFilter
}
```

graphql/schemabuilder/pagination.go
```diff
@@ -995,6 +1009,11 @@ func (c *connectionContext) parsePaginatedReturnSignature(m *method) (err error)
 	if len(out) > 0 && out[0] == reflect.TypeOf(PaginationInfo{}) {
 		c.ReturnsPageInfo = true
 		out = out[1:]
+		if len(out) == 0 || out[0] != reflect.TypeOf(PostProcessOptions{}) {
+			err = fmt.Errorf("%s returns a PaginationInfo not followed by a PaginationPostProcessOptions", c.funcType)
+			return
```
wedaly: just to confirm my understanding: if we accidentally forget to add this to a manually paginated field, we'd expect the service to fail on startup?  (that's what I'd like to have happen -- it's way scarier if stuff might fail silently)

graphql/schemabuilder/pagination.go
wit221: Agreed. Going further, however, I would also decouple the ApplyTextFilter with SetPaginationInfo logic. With the above approach, resolvers that will want to set a paginationInfo without having to apply a text filter will be forced to also apply the text filter.

graphql/schemabuilder/pagination.go
wit221: That's correct. This PR also includes a test that verifies just that.

wit221:
@wedaly changelog:
- Improved comments;
- Added a `SetPaginationInfo` option.

Would you care to take another look?

graphql/schemabuilder/pagination.go
```diff
@@ -149,6 +149,14 @@ type PaginationInfo struct {
 	Pages          []string
 }
 
+// PostProcessOptions is used to instruct Thunder to perform additional operations on the output
+// in the case of manually paginated resolvers.
+// This struct should be returned as the second result from a manually-paginated field func.
+type PostProcessOptions struct {
+	ApplyTextFilter   bool
+	SetPaginationInfo bool
```
wedaly: I'd suggest adding a comment explaining what each of these does.

wit221:
FYI changelog:
- renamed SetPaginationInfo -> SetPageInfo;
- Added comments explaining ApplyTextFilter and SetPageInfo;
- Added tests that test different combinations of the above options: with ApplyTextFilter only, with SetPageInfo only and with both.

graphql/schemabuilder/pagination.go
```diff
@@ -149,6 +149,18 @@ type PaginationInfo struct {
 	Pages          []string
 }
 
+// PostProcessOptions is used to instruct Thunder to perform additional operations on the output
+// in the case of manually paginated resolvers.
+// This struct should be returned as the second result from a manually-paginated field func.
+type PostProcessOptions struct {
+	// Whether or not Thunder should apply filtering on the output.
+	ApplyTextFilter bool
```
CodeBrew28: What happens if thunder applies the text filter and filters out too many items?

wit221: Resolvers should set this option with the possibility in mind that an arbitrary number of items will get filtered out.

A use case is when externally-managed resolvers determine they would like to fall back  to Thunder's filtering and fetch the entire data range (disregarding the `First` parameter). Then, Thunder's potential filtering of data such that the resulting set's length is less than the `First` parameter is just an inherent characteristic of the query and data store - there are just not enough items that fit the provided filter.

CodeBrew28: Ah, that makes a lot more sense!

CodeBrew28: Can we maybe add a comment explaining when to use it

wit221: 👍

PR #345, opened 2020-01-07
Federation Schema Merging 

CodeBrew28:
Federation Part 1: Schema Merging

Merging between versions (intersection)
During an update to a new version, we want to have a version of the schema that is compatible with both schemas. In this case, we want to get the intersection between the two schemas so that we know what we are running is compatible with both schemas. 

Merging between schemas (union)
The gateway needs to be able to take all the schemas from separate servers and generate a merged version that is compatible with all schemas and annotated with federation info (ie which server has which field func) so that it can generate a query plan

### DIFF
--- a/federation/schema/merge_schemas.go
+++ b/federation/schema/merge_schemas.go
@@ -1,0 +1,419 @@
+package federation
+
+import (
+	"errors"
+	"fmt"
+	"sort"
+
+)
+
+// MergeMode controls how to combine two different schemas. Union is used for
+// two independent services, Intersection for two different versions of the same
+// service.
+type MergeMode string
+
+const (
+	// Union computes a schema that is supported by the two services combined.
+	//
+	// A Union is used to to combine the schema of two independent services.
+	// The proxy will split a GraphQL query to ask each service the fields
+	// it knows about.
+	//
+	// Two schemas must be compatible: Any overlapping types (eg. a field that
+	// is implemented by both services, or two input types) must be compatible.
+	// In practice, this means types must be identical except for non-nil
+	// modifiers.
+	//
+	// XXX: take intersection on ENUM values to not confuse a service with a
+	// type it doesn't support?
+	Union MergeMode = "union"
+
+	// Intersection computes a schema that is supported by both services.
+	//
+	// An Intersection is used to combine two schemas of different versions
+	// of the same service. During a deploy, only of two versions might be
+	// available, and so queries must be compatible with both schemas.
+	//
+	// Intersection computes a schema that can be executed by both services.
+	// It only includes types and fields (etc.) exported by both services.
+	// Overlapping types must be compatible as in a Union merge.
+	//
+	// One surprise might be that newly added ENUM values or UNION types might
+	// be returned by the merged schema.
+	Intersection MergeMode = "intersection"
+)
+
+// introspectionTypeRef is a type reference from the GraphQL introspection
+// query.
+type introspectionTypeRef struct {
+	Kind   string                `json:"kind"`
+	Name   string                `json:"name"`
+	OfType *introspectionTypeRef `json:"ofType"`
+}
+
+func (t *introspectionTypeRef) String() string {
+	if t == nil {
+		return "<nil>"
+	}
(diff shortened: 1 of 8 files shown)


federation/schema/merge_schemas.go
```diff
@@ -0,0 +1,419 @@
...
+	for name := range types {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	merged := make([]introspectionType, 0, len(names))
+	for _, name := range names {
+		p := types[name]
+		if len(p) == 1 {
+			// When interesction only include types that appear in both schemas
```
jellevandenhooff: intersection

federation/schema/normalize.go
```diff
@@ -0,0 +1,47 @@
...
+
+
+import (
+	"fmt"
+
+	"github.com/samsarahq/thunder/graphql"
+)
+
+
+func collectTypes(typ graphql.Type, types map[graphql.Type]string) error {
```
jellevandenhooff: // collectTypes finds all types reachable from typ and stores them in types as a map from type to name.

graphql/schemabuilder/schema.go
```diff
@@ -137,6 +137,9 @@ func (s *Schema) Build() (*graphql.Schema, error) {
 		typeCache:    make(map[reflect.Type]cachedType, 0),
 	}
 
+	s.Object("Query", query{})
```
jellevandenhooff: can you put this change in its own commit

graphql/introspection/introspection.go
```diff
@@ -344,7 +344,7 @@ func (s *introspection) schema() *graphql.Schema {
 	return schema.MustBuild()
 }
 
-func AddIntrospectionToSchema(schema *graphql.Schema) {
+func BareIntrospectionSchema(schema *graphql.Schema) *graphql.Schema {
```
jellevandenhooff: can you put this change in its own commit?

federation/schema/schema.go
```diff
@@ -0,0 +1,337 @@
...
+		return nil, fmt.Errorf("unknown type kind %s", t.Kind)
+	}
+}
+
+// parseInputFields maps a list of input types to a list of graphql types
+func parseInputFields(source []introspectionInputField, all map[string]graphql.Type) (map[string]graphql.Type, error) {
+	fields := make(map[string]graphql.Type)
+
+	for _, field := range source {
+
```
jellevandenhooff: remove this newline

federation/schema/schema.go
```diff
@@ -0,0 +1,337 @@
...
+		rawType, err := lookupType(field.Type, all)
+		if err != nil {
+			return nil, fmt.Errorf("type %s not found", rawType.Name)
+		}
+		switch rawType.Kind {
+		case "INPUT_OBJECT", "SCALAR":
+		default:
+			return nil, fmt.Errorf("input field %s has bad typ: %s",	field.Name, rawType.Kind)
+		}
+
```
jellevandenhooff: remove this newline

federation/schema/schema.go
```diff
@@ -0,0 +1,337 @@
...
+	case "LIST":
+		return lookupType(t.OfType, all)
+	case "NON_NULL":
+		return lookupType(t.OfType, all)
+	default:
+		return nil, fmt.Errorf("unknown type kind %s", t.Kind)
+	}
+}
+
+
```
jellevandenhooff: actually i think this files just needs gofmt

graphql/introspection/introspection.go
```diff
@@ -381,7 +390,7 @@ func ComputeSchemaJSON(schemaBuilderSchema schemabuilder.Schema) ([]byte, error)
 	}
 
 	executor := graphql.NewExecutor(graphql.NewImmediateGoroutineScheduler())
-	value, err := executor.Execute(context.Background(), schema.Query, nil, query)
+	value, err := executor.Execute(context.Background(), schema.Query,nil, query)
```
jellevandenhooff: gofmt

PR #346, opened 2020-01-14
federation: add normalize implementation

jellevandenhooff:
Normalize flattens queries into easy-to-wrangle shapes.

TODO: Use normalize for query execution in the non-federation codepath.

### DIFF
--- a/federation/normalize.go
+++ b/federation/normalize.go
@@ -1,0 +1,282 @@
+package federation
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+
+	"github.com/samsarahq/thunder/graphql"
+)
+
+// CollectTypes finds all types reachable from typ and stores them in types as a
+// map from type to name.
+//
+// TODO: Stick this in an internal package.
+func CollectTypes(typ graphql.Type, types map[graphql.Type]string) error {
+	if _, ok := types[typ]; ok {
+		return nil
+	}
+
+	switch typ := typ.(type) {
+	case *graphql.NonNull:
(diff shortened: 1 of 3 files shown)


federation/normalize.go
```diff
@@ -0,0 +1,290 @@
...
+	default:
+		return false, fmt.Errorf("unknown fragment type %s", fragment.On)
+	}
+}
+
+// flattenFragments flattens all fragments at the current level. It inlines the
+// selections of each fragment, but does not descend down recursively into those
+// selections.
+func (f *flattener) flattenFragments(selectionSet *graphql.SelectionSet, typ *graphql.Object, c *complexityLimiter, target *[]*graphql.Selection) error {
+	// TODO: is this a good place to limit query complexity?
```
CodeBrew28: How do we want to limit query complexity? Is it saying that there's a limit on how much we can recurse?

jellevandenhooff: Yes, something like that, I'm not quite sure. I figured we can return to that later and mostly don't worry about it with query whitelisting.

federation/normalize.go
```diff
@@ -0,0 +1,278 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+
+	"github.com/samsarahq/thunder/graphql"
+)
+
+func collectTypes(typ graphql.Type, types map[graphql.Type]string) error {
```
CodeBrew28: This function is also in `schema/normalize.go`, can we remove it from there?

jellevandenhooff: done

PR #348, opened 2020-01-14
introspection empty type name should be nil

lulucas:


jellevandenhooff:
Thanks. Rewrote snapshots and merged in #349.

PR #350, opened 2020-01-15
go sql/fields: Add support for all protobuf type sql fields

willhug:
Summary: Currently we have support for protobuf sql fields on structs if
they implement the Marshal/Unmarshal methods.  This works for most protobuf
types generated by gogoproto.  However, since the Marshal/Unmarshal methods 
generated by gogoproto can be quite massive (and significantly increase the 
size of a binary), we want to have the option to _not_ generate Marshal/Unmarshal
methods (we're already doing this in one of our repos). But, we still want to have
support for using protobuf fields in our structs that interact with the database.

This PR adds support for protobuf generated structs that don't have generated 
Marshal/Unmarshal methods.  We do this by calling into the proto.Marshal/Unmarshal
methods with a type that implements the `proto.Message` interface.  We also
add a series of tests to validate that the Read/Write operations still work.


internal/fields/sql.go
```diff
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/go-sql-driver/mysql"
+	"github.com/gogo/protobuf/proto"
```
jellevandenhooff: I think we already might have a gogo proto dependency, but if not, it'd be nice to not have this edge. Can we inline the important parts of the interface?

internal/fields/sql.go
```diff
@@ -82,6 +94,12 @@ func (f Valuer) Value() (driver.Value, error) {
 		if iface, ok := i.(encoding.BinaryMarshaler); ok {
 			return iface.MarshalBinary()
 		}
+		if v, ok := nonPointerProtoMessage(f.Descriptor, f.value); ok {
```
jellevandenhooff: when do we end up with a non-pointer? i'd expect proto structs to be big and probably don't want to copy them

internal/fields/sql_test.go
```diff
@@ -273,3 +275,33 @@ func TestField_SupportsProtobuf(t *testing.T) {
 		assert.Equal(t, event, got.Elem().Interface())
 	})
 }
+
+func TestField_SupportsProtobufWithoutMarshalMethods(t *testing.T) {
+	t.Run("pointer", func(t *testing.T) {
```
jellevandenhooff: if there's no code shared between the tests, i'd prefer multiple Test_ methods to prevent the extra indentation from t.Run

internal/fields/sql.go
willhug: It's the non-pointer case:
```
type MyType struct {
  MyProto proto.SomeProtoType
}
```
As opposed to:
```
type MyType struct {
  MyProto *proto.SomeProtoType
}
```

internal/fields/sql.go
willhug: We do already have the dependency (it's checked into the vendor folder).  Unfortunately I don't think there's a good way to avoid adding it since we're depending on the `proto.Marshal`/`proto.Unmarshal` functions.

internal/fields/sql_test.go
willhug: fair, I just copied the style from above to be consistent.  Can change.

PR #351, opened 2020-01-15
Federation Planner

CodeBrew28:
Planner: Given a query and a merged schema, it creates a plan that tells it what subqueries to execute on what service.


federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+type PathStep struct {
+	Kind StepKind
+	Name string
+}
+
+type Plan struct {
+	Path    []PathStep
+	Service string
+	Kind    string
+	// XXX: What are we using Type for here again? -- oh, it's for the __federation field...
```
jellevandenhooff: could you get rid of this comment, and document the top level types? also happy to do this in a follow-up pr.

federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+	Type         string
+	SelectionSet *graphql.SelectionSet
+	After        []*Plan
+}
+
+type Executor struct {
```
jellevandenhooff: how about we rename this struct to Planner?

federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+func (e *Executor) planObject(typ *graphql.Object, selectionSet *graphql.SelectionSet, service string) (*Plan, error) {
+	p := &Plan{
+		Type:         typ.Name,
+		Service:      service,
+		SelectionSet: &graphql.SelectionSet{},
+		After:        nil,
+		Kind:         "query",
+	}
+
+	// XXX: pass in sub-path (and sub-plan slice?) to make sub-plan munging simpler?
```
jellevandenhooff: clean these comments also please

federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+			// XXX: assert existence of types elsewhere?
+			var err error
+			// XXX type assertoin
+			childPlan, err = e.plan(field.Type, selection.SelectionSet, service)
+			if err != nil {
+				return nil, fmt.Errorf("planning for %s: %v", selection.Name, err)
+			}
+		}
+
+		// newSelection := &graphql.RawSelection{
```
jellevandenhooff: can we make a copy here still? i don't want to modify the incoming query if we can help it.

federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+		fieldInfo := e.schema.Fields[field]
+
+		// if we can stick to the current service, stay there
+		if fieldInfo.Services[service] {
+			localSelections = append(localSelections, selection)
+		} else {
+			serviceWithField := ""
+
+			for service, hasField := range fieldInfo.Services {
+					if hasField {
```
jellevandenhooff: we'll want to expose knobs in the future to tune where to route specific field funcs

federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	if needKey {
+		hasKey := false
+
+		for _, selection := range p.SelectionSet.Selections {
+			if selection.Name == "__federation" && selection.Alias == "__federation" {
+				hasKey = true
+			} else if selection.Alias == "__federation" {
+				// error, conflict, can't do this.
```
jellevandenhooff: let's return an error here?

federation/planner_test.go
```diff
@@ -0,0 +1,407 @@
...
+
+	f, err := newFlattener(merged.Schema)
+	return &Executor{
+		flattener: f,
+		schema: merged,
+	}, nil
+}
+
+func TestPlanner(t *testing.T) {
+	e,err := setupExecutor(t)
```
jellevandenhooff: go fmt this file

federation/planner_test.go
```diff
@@ -0,0 +1,407 @@
...
+					},
+				},
+			},
+		},
+		{
+			Name: "query with fields from nested levels between two services",
+			Input: `query Test {
+				s1fff {
+					s2ok
+
```
jellevandenhooff: delete empty line

federation/planner_test.go
```diff
@@ -0,0 +1,407 @@
...
+							SelectionSet: mustParse(`{
+								s2ok
+							}`),
+						},
+					},
+				},
+			},
+		},
+		{
+			Name: "query with fields from nested levels between two services",
```
jellevandenhooff: this test looks identical to the one above. get rid of it? i think the one below covers it

federation/planner_test.go
```diff
@@ -0,0 +1,407 @@
...
+			require.NoError(t, err)
+			assert.Equal(t, testCase.Output, plan.After)
+			// prettyPrintSelectionSets(plan) // uncomment to print the plan and debug
+		})
+	}
+
+}
+
+
+func prettyPrintSelectionSets(rootPlan *Plan){
```
jellevandenhooff: maybe we move this to the graphql types package later?

federation/planner.go
wedaly: :+1: for documenting the public types, I think that would make this code much easier to follow.  Could I ask that we also document the fields in `Plan`?  It would be super helpful to know what kinds of values we expect for "Kind", "Type", etc. without having to refer to the test cases.

federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+	var localSelections []*graphql.Selection
+	selectionsByService := make(map[string][]*graphql.Selection)
+
+	// Assert that we are working with a flattened query.
+	if len(selectionSet.Fragments) > 0 {
+		return nil, errors.New("selectionSet has fragments, expected flattened query")
+	}
+
+	for _, selection := range selectionSet.Selections {
+		if selection.Name == "__typename" {
```
wedaly: I'm not sure I understand how "__typename" is being used (what does it mean, who sets it, etc.).  Maybe this would be a good thing to document somewhere?  (top-level comment in this file?)

federation/planner.go
```diff
@@ -0,0 +1,300 @@
...
+
+		// p.AfterNode.Services[other] = subPlan
+		p.After = append(p.After, subPlan)
+	}
+
+	if needKey {
+		hasKey := false
+
+		for _, selection := range p.SelectionSet.Selections {
+			if selection.Name == "__federation" && selection.Alias == "__federation" {
```
wedaly: similar comment for "__federation", it would be great to document these magic strings somewhere

federation/planner.go
```diff
@@ -0,0 +1,318 @@
...
+type StepKind int
+
+const (
+	KindType StepKind = iota
+	KindField
+)
+
+// PathStep defines where in the plan this part of the query came from
+type PathStep struct {
+	Kind StepKind // KindType indicates a selection type and KindField indicates a union type
```
CodeBrew28: @jellevandenhooff not sure if this correct or what it should be

federation/planner.go
CodeBrew28: +1 Was thinking about doing this in the PR right after

federation/planner.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	p, err := e.plan(schema, flattened, "gateway-coordinator-service")
+	if err != nil {
+		return nil, err
+	}
+
+	if query.Kind == "mutation" {
+		if len(p.After) > 1 {
+			return nil, errors.New("only support 1 mutation step to maintain ordering")
```
wedaly: I'm not sure what "maintain ordering" means in this context.  Maybe add a comment explaining what problem this avoids?

federation/planner.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	return p, nil
+
+}
+
+func (e *Planner) planUnion(typ *graphql.Union, selectionSet *graphql.SelectionSet, service string) (*Plan, error) {
+	plan := &Plan{
+		// TODO: only include __typename if needed for dispatching? ie. len(types) > 1 and len(fragments) > 0?
+		// TODO: ensure __typename doesn't conflict with another field?
```
wedaly: Are these TODOs for future PRs?  Or should be resolved before merging this PR?

CodeBrew28: yup, was planning to do it in a future PR

federation/planner.go
CodeBrew28: Added a comment, but it makes sure that mutations run on separate graphql servers won't be run out of order

CodeBrew28:
@wedaly  I added a section called "Executing a subquery" to explain how some of the fields (eg "__federation" and "__typename") are used. It walks through an example of what the subquery will look like when it gets resolved by the executor.

federation/planner.go
```diff
@@ -0,0 +1,368 @@
...
+// and describes what sub-queries need to be resolved first.
+type Planner struct {
+	schema    *SchemaWithFederationInfo //schema describes what fields the graphql servers know about along with the services that know how to execute each field
+	flattener *flattener                //flattener knows how to combine all the fragments on a query into a singel query
+}
+
+// Executing a subquery
+//
+// When a subquery is run on a seperate graphql server, we want the subquery to be nested
+// on the "__federation" type so that the graphQL server
```
wedaly: this sentence seems incomplete

federation/planner.go
```diff
@@ -0,0 +1,368 @@
...
+// {
+// 	allFoos {
+// 		name
+// 	}
+// }
+// When we send the subquery to service2 it should look like the example below
+// {
+// 	__federation {
+// 		fooById(id: $id) {
+//      __typename
```
wedaly: nit: indendentation a little off here?

federation/planner.go
```diff
@@ -0,0 +1,368 @@
...
+// 			name
+// 		}
+// 	}
+// }
+// "__federation" becomes the root query that the subquery is nested under,
+// "fooById" is the federated object type that we need to refetch,
+// and "__typename" lets us know what type the object is.
```
wedaly: does "us" mean the gateway or the schema server?

CodeBrew28: the gateway, just changed it!

PR #356, opened 2020-02-11
Federated executor!

CodeBrew28:
Executes a query on seperate servers and stitches the responses together


thunderpb/federation.proto
```diff
@@ -0,0 +1,39 @@
...
+package thunderpb;
+option go_package = "github.com/samsarahq/thunder/thunderpb";
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+message Selection {
+  string name = 1;
+  string alias = 2;
+  SelectionSet selectionSet = 3;
+  bytes arguments = 4;
```
wedaly: maybe comment that this is JSON-encoded?

graphql/schemabuilder/input.go
```diff
@@ -316,7 +316,7 @@ func (sb *schemaBuilder) makeSliceParser(typ reflect.Type) (*argParser, graphql.
 	return &argParser{
 		FromJSON: func(value interface{}, dest reflect.Value) error {
 			asSlice, ok := value.([]interface{})
-			if !ok {
+			if asSlice != nil && !ok {
```
wedaly: I would have thought that if the type conversion fails, we wouldn't care about the value of `asSlice`.  Is this really necessary?

federation/executor.go
```diff
@@ -0,0 +1,323 @@
...
+
+	types, err := convertSchema(schemas)
+	if err != nil {
+		return nil, fmt.Errorf("converting schema error: %v", err)
+	}
+
+	introspectionSchema := introspection.BareIntrospectionSchema(types.Schema)
+	introspectionServer := &Server{schema: introspectionSchema}
+
+	executors["introspection"] = &DirectExecutorClient{Client: introspectionServer}
```
wedaly: consider making the key string a const and maybe adding a check that the `executors` map doesn't already contain this key?

federation/executor.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+
+func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Executor, error) {
+	// Fetches the schemas from the executors clients
+	schemas := make(map[string]*introspectionQueryResult)
+	for server, client := range executors {
+		schema, err := fetchSchema(ctx, client)
+		if err != nil {
+			return nil, fmt.Errorf("fetching schema %s: %v", server, err)
```
wedaly: I'm wondering how you're thinking about partial failures here.  What happens if one of the introspection queries fails (maybe a temporary network error)?  Would the caller of `NewExecutor` be responsible for retrying?

federation/executor.go
```diff
@@ -0,0 +1,323 @@
...
+								SelectionSet: selectionSet,
+							},
+						},
+					},
+				},
+			},
+		}
+	}
+
+	// TODO: make sure that if this hangs we're still good?
```
wedaly: I'd imagine the caller could set a timeout on `ctx` so this code doesn't wait indefinitely on the RPC?

federation/executor.go
```diff
@@ -0,0 +1,323 @@
...
+func (e *Executor) execute(ctx context.Context, p *Plan, keys []interface{}) ([]interface{}, error) {
+	var res []interface{}
+
+	if p.Service == gatewayCoordinatorServiceName {
+		res = []interface{}{
+			map[string]interface{}{},
+		}
+	} else {
+		var err error
+		// Eexecutes that part of the plan (the subquery) on one of the federated gqlservers
```
wedaly: small typo: Eexecutes -> Executes

federation/executor.go
```diff
@@ -0,0 +1,323 @@
...
+		subPlan := subPlan
+		var pf pathFollower
+		if p.Service != gatewayCoordinatorServiceName {
+			if err := pf.extractTargets(res, subPlan.Path); err != nil {
+				return nil, fmt.Errorf("failed to follow path %v: %v", subPlan.Path, err)
+			}
+		} else {
+			pf.keys = nil
+			pf.targets = []map[string]interface{}{
+				res[0].(map[string]interface{}),
```
wedaly: who guarantees that `res` has at least one element and is a map?  Maybe a comment explaining why this is okay would be helpful

federation/executor.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+}
+
+func (e *Executor) Execute(ctx context.Context, q *graphql.Query) (interface{}, error) {
+	p, err := e.planner.planRoot(q)
+	if err != nil {
+		return nil, err
+	}
+
+	printPlan(p)
```
wedaly: leftover from debugging?

PR #359, opened 2020-02-26
go graphql/server: Close Mutation subscription after execution

willhug:
Summary: Noticed this as a memory leak when we would upload a lot of
images through mutation endpoints.  The subscriptions for mutations are
never cleaned up, and the memory from the request (an image) would
remain in-memory until we'd kill the websocket connection.

Tested this by uploading a bunch of images and monitoring the heap usage.

Before change:
Screen Shot 2020-02-25 at 5 14 36 PM
After change:
Screen Shot 2020-02-25 at 5 17 14 PM
(you can't even see the spike, we miss it)

### DIFF
--- a/graphql/server.go
+++ b/graphql/server.go
@@ -349,6 +349,7 @@ func (c *conn) handleMutate(in *inEnvelope) error {
 		go c.rerunSubscriptionsImmediately()
 
 		initial = false
+		go c.closeSubscription(id)
 		return nil, errors.New("stop")
 	}, c.minRerunIntervalFunc(c.ctx, query), c.alwaysSpawnGoroutineFunc(c.ctx, query))
 


willhug:
Or, are mutations supposed to stick around forever?  A lot of the code seems to have been built for subscription invalidation, but I don't think we do that right?

graphql/server.go
```diff
@@ -349,6 +349,7 @@ func (c *conn) handleMutate(in *inEnvelope) error {
 		go c.rerunSubscriptionsImmediately()
 
 		initial = false
+		go c.closeSubscription(id)
```
CodeBrew28: Once we run a mutation, it makes a change to the data, and then returns the new data back to the client. By not closing the subscription, does that mean we were rerunning that mutation every time the data changed?

willhug: We weren't rerunning the query because we return an error below.  The Rerunner exits immediately as soon as it gets an error that is not of type "RetrySentinelError".  The subscription would stick around though.

CodeBrew28: Ahh that makes sense, nice catch!

anubhav94:
🙏 🙌  nice find

PR #368, opened 2020-05-12
Federated Executor Without Nested service calls

CodeBrew28:
This supports us in parsing the plan and executing the query on multiple services. We can split up the query and run it on different schemas or multiple schemas, but not support nested queries that are on separate schemas. 

This is a stacked PR, so only the last two commits are new.

### DIFF
--- a/federation/executor.go
+++ b/federation/executor.go
@@ -3,8 +3,10 @@ package federation
 import (
 	"context"
 	"encoding/json"
+	"sync"
 
 	"github.com/samsarahq/go/oops"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/samsarahq/thunder/graphql"
 	"github.com/samsarahq/thunder/graphql/introspection"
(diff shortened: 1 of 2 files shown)


federation/executor.go
```diff
@@ -88,3 +91,100 @@ func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Exe
...
+				}
+				for k, v := range result {
+					subPlanResults[k] = v
+				}
+			}
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
```
KevinXing, line 165: Should we combine `subPlanResults` to `res` somewhere?

federation/executor.go
```diff
@@ -88,3 +91,100 @@ func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Exe
...
+
+	g, ctx := errgroup.WithContext(ctx)
+	var resMu sync.Mutex
+
+	// For every nested query in the plan, execute it on the specified service and stitch
+	// the results into a response
+	for _, currentSubPlan := range p.After {
+		subPlan := currentSubPlan
+		var subPlanMetaData pathSubqueryMetadata
+		if p.Service == gatewayCoordinatorServiceName {
```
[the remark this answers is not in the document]
CodeBrew28: The first call to execute from any query is always from the gateway coordinator service. Since its a root query, there are no federated keys

federation/executor.go
CodeBrew28: Since `subPlanMetaData.results` points to res, we are directly updating it already. We also acquire the mutex lock right before this section so that its thread-safe.

federation/executor.go
```diff
@@ -88,3 +91,100 @@ func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Exe
...
+	if p.Service != gatewayCoordinatorServiceName {
+		var err error
+		res, err = e.runOnService(ctx, p.Service, p.Type, keys, p.Kind, p.SelectionSet)
+		if err != nil {
+			return nil, oops.Wrapf(err, "run on service")
+		}
+	}
+
+	g, ctx := errgroup.WithContext(ctx)
+	var resMu sync.Mutex
```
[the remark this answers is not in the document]
CodeBrew28: Added this comment amove the muxtec lock
	// resMu protects the results (res) as we stitch the results together from separate goroutines
	// executing different parts of the plan on different services

federation/executor.go
```diff
@@ -88,3 +91,104 @@ func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Exe
...
+			result, ok := results.(map[string]interface{})
+			if !ok {
+				return fmt.Errorf("result is not an object: %v", result)
+			}
+
+			// Acquire mutex lock before modifying results
+			resMu.Lock()
+			defer resMu.Unlock()
+			for k, v := range result {
+				subPlanMetaData.results[k] = v
```
KevinXing, line 163: What if the key is already existing?

CodeBrew28: I updated it to error if the key exists. In the flatten stage in the planner it checks for duplicates and removes them or gives a corresponding error, so if this happens we should have the query error out.

PR #369, opened 2020-05-19
Federated Executor: Support Args

CodeBrew28:
The unparsed args are sent in the request, and parsed in the corresponding service before the subquery is run. If any args are missing, the error is correctly bubbled up to the origional requ


CodeBrew28:
There weren't unit tests for testing federation queries with args before (not sure if that answers your question?)

PR #372, opened 2020-05-26
Federation Executor: nested services union types

CodeBrew28:
Federated executor where we have a union type in which we need to parse the keys and send a subquery to a different gql server

### DIFF
--- a/federation/executor.go
+++ b/federation/executor.go
@@ -15,6 +15,7 @@ import (
 
 const keyField = "__key"
 const federationField = "__federation"
+const typeNameField = "__typeName"
 
 type ExecutorClient interface {
 	Execute(ctx context.Context, req *graphql.Query) ([]byte, error)
@@ -182,7 +183,7 @@ func (pathTargets *pathSubqueryMetadata) extractKeys(node interface{}, path []Pa
 	if slice, ok := node.([]interface{}); ok {
 		for i, elem := range slice {
 			if err := pathTargets.extractKeys(elem, path); err != nil {
-				return fmt.Errorf("idx %d: %v", i, err)
+				return oops.Errorf("idx %d: %v", i, err)
 			}
 		}
 		return nil
@@ -221,6 +222,16 @@ func (pathTargets *pathSubqueryMetadata) extractKeys(node interface{}, path []Pa
 		}
 		if err := pathTargets.extractKeys(next, path[1:]); err != nil {
 			return fmt.Errorf("elem %s: %v", next, err)
+		}
+	case KindType:
+		typ, ok := obj["__typename"].(string)
+		if !ok {
+			return fmt.Errorf("does not have string key __typename")
+		}
+		if typ == step.Name {
+			if err := pathTargets.extractKeys(obj, path[1:]); err != nil {
+				return fmt.Errorf("typ %s: %v", typ, err)
+			}
 		}
 	default:
 		return fmt.Errorf("unsupported step type name: %s kind: %v", step.Name, step.Kind)
@@ -277,7 +288,6 @@ func (e *Executor) execute(ctx context.Context, p *Plan, keys []interface{}) (in
 			defer resMu.Unlock()
 			for k, v := range result {
 				if _, ok := subPlanMetaData.results[k]; !ok {
-					deleteKey(v, federationField)
 					subPlanMetaData.results[k] = v
 				} else {
 					if k != keyField || v != subPlanMetaData.results[k] {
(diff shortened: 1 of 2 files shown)


federation/executor.go
```diff
@@ -280,7 +292,7 @@ func (e *Executor) execute(ctx context.Context, p *Plan, keys []interface{}) (in
 					deleteKey(v, federationField)
 					subPlanMetaData.results[k] = v
 				} else {
-					if k != keyField || v != subPlanMetaData.results[k] {
+					if k != keyField && v == subPlanMetaData.results[k] {
```
KevinXing, line 295: Why to make this change?

federation/executor.go
```diff
@@ -222,6 +223,17 @@ func (pathTargets *pathSubqueryMetadata) extractKeys(node interface{}, path []Pa
 		if err := pathTargets.extractKeys(next, path[1:]); err != nil {
 			return fmt.Errorf("elem %s: %v", next, err)
 		}
+	case KindType:
+		typ, ok := obj["__typename"].(string)
+		if !ok {
+			return fmt.Errorf("does not have string key __typename")
+		}
+
+		if typ == step.Name {
```
KevinXing, line 232: What if typ != step.Name?

CodeBrew28: In that case we just want to not extract any keys, but we don't want to error. I think this is easier to illustrate with an example

Let's say our query is 
```
query Foo {
				everyone {
					... on Admin {
						id
						hiding
					}
					... on User {
						id
					}
				}
			}
```
We first call extract keys on the results from "everyone", and when we figure out what the next object is https://github.com/samsarahq/thunder/blob/7fb9b9a865aece0ffc5b73f9398fdf8310e84886/federation/executor.go#L219 it has both a union of both the user and admin type. 

We then look at every element in the slice https://github.com/samsarahq/thunder/blob/7fb9b9a865aece0ffc5b73f9398fdf8310e84886/federation/executor.go#L183 , and recursively call it for the user and admin type. 

When we get to the line https://github.com/samsarahq/thunder/blob/7fb9b9a865aece0ffc5b73f9398fdf8310e84886/federation/executor.go#L232 , we want to check which type it is before extracting keys because we can call it with the wrong type. 


I can absolutely see how this is confusing, so I moved this check to when we are parsing the slice such that we only call extract keys if the type name is correct.

federation/executor.go
CodeBrew28: Ah, that was a mistake due to tests, nice catch!

federation/executor.go
KevinXing: Ah, I see, thanks for the explanation! 

Could you please squash the commit into the previous one?

CodeBrew28: updated!

federation/executor.go
```diff
@@ -181,8 +182,25 @@ func (pathTargets *pathSubqueryMetadata) extractKeys(node interface{}, path []Pa
 	// Extract key for every element in the slice
 	if slice, ok := node.([]interface{}); ok {
 		for i, elem := range slice {
-			if err := pathTargets.extractKeys(elem, path); err != nil {
-				return fmt.Errorf("idx %d: %v", i, err)
+			obj, ok := elem.(map[string]interface{})
+			if !ok {
+				return fmt.Errorf("does not have string key __typename")
```
KevinXing, line 187: The error message does not reflect the check.

federation/executor.go
```diff
@@ -287,6 +287,7 @@ func (e *Executor) execute(ctx context.Context, p *Plan, keys []interface{}) (in
 			subPlanMetaData.keys = nil // On the root query there are no specified keys
 			subPlanMetaData.results = res
 		} else {
+			fmt.Println("keys", subPlan.Path)
```
KevinXing, line 290: Remove this?

federation/executor.go
```diff
@@ -324,5 +337,6 @@ func (e *Executor) Execute(ctx context.Context, query *graphql.Query) (interface
 	if err != nil {
 		return nil, err
 	}
+	deleteKey(r, federationField)
```
CodeBrew28, line 340: I moved delete key back to the end since it would minimize the number of times we would have to parse through all the entries in the query to delete the federation field. 

Before this PR it was nested in the execute function, but it wasn't recursively looking through the entire query. The union tests have a use case where the federtionField is in a nested field, so we would need to change the implementation which makes deleting the federation field in the execute function slightly more expensive

PR #375, opened 2020-06-01
Federation: Custom federated keys

CodeBrew28:
A federated server can have multiple keys. We pass only the necessary keys to each service by checking what keys that federated schema knows about.


graphql/schemabuilder/types.go
```diff
@@ -263,6 +267,29 @@ func (s *Object) ManualPaginationWithFallback(name string, manualPaginatedFunc i
 	s.Methods[name] = m
 }
 
+func (s *Object) FederatedFieldFunc(name string, f interface{}, options ...FieldFuncOption) {
+
```
KevinXing, line 271: Remove the empty line.

graphql/schemabuilder/types.go
```diff
@@ -263,6 +267,29 @@ func (s *Object) ManualPaginationWithFallback(name string, manualPaginatedFunc i
...
+	m := &method{Fn: f}
+	for _, opt := range options {
+		opt.apply(m)
+	}
+
+	if _, ok := s.Methods[name]; ok {
+		panic("duplicate method")
+	}
+	federatedMethodName := fmt.Sprintf("%s-%s", name, s.ServiceName)
```
KevinXing, line 284: How do you feel to switch name and s.ServiceName? 
It is more natural to me to have the serviceName first.

CodeBrew28: I think I slightly prefered the object name in front so that when I look at all the federated methods in the format of "<object>-<serviceName>", I can think off all the services that know about that object together. Its also an internal format that isn't surface, so I can switch it if you strongly prefer the other way.

federation/schema.go
```diff
@@ -83,6 +91,47 @@ func convertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
...
+											f := obj.Fields[field.Name]
+											if f.FederatedKey == nil {
+												f.FederatedKey = make(map[string]bool, len(serviceNames))
+											}
+											f.FederatedKey[service] = true
+										}
+									}
+								}
+							}
+						}
```
KevinXing, line 127: Can we replace this logic with 
```

						obj, ok := types[objName].(*graphql.Object)
						if ok {
							for fName, f := range obj.Fields {
								if _, ok := inputType.InputFields[fName]; !ok {
									continue
								}
								if f.FederatedKey == nil {
									f.FederatedKey = make(map[string]bool, len(serviceNames))
								}

								f.FederatedKey[service] = true 
							}
						}
```

federation/schema.go
```diff
@@ -83,6 +91,47 @@ func convertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
...
+			// fields are the federated keys. They annotate that information 
+			// on the field object.
+			if typ.Name == "Federation" {
+				for _, field := range typ.Fields {
+					for _, arg := range field.Args {
+						
+						// Extract the type name from the formatting <object>-<service>
+						rootType := getRootType(arg.Type)
+						names := strings.SplitN(field.Name, "-", 2)
+						objName := names[0]
```
KevinXing, line 105: Why we never use names[1]?

CodeBrew28: Since we know the service we are on, we don't need to use names[1]. We do however need the field to be named "objectName-serviceName" so that when we merge the schemas it gets treated as an independent field func instead of a field collision. This also allows us to query the correct field in the executor.

federation/schema.go
CodeBrew28: Yes! Really excited about how this simplifies the logic!

graphql/types.go
```diff
@@ -134,6 +134,7 @@ type Field struct {
 	Batch        bool
 	External     bool
 	Expensive    bool
+	FederatedKey map[string]bool
```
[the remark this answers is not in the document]
KevinXing: Maybe more verbose like:
"FederatedKey tells us which services need this field as federated key".

federation/schema.go
```diff
@@ -83,6 +92,53 @@ func convertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
...
+						}
+						objName := names[0]
+						inputType, ok := types[rootType.Name].(*graphql.InputObject)
+						if !ok {
+							return nil, oops.Errorf("Object %s is not an input object, but it is an argument to the field %s", rootType.Name, field.Name)
+						}
+
+						// For the object type on the current service
+						for _, t := range serviceSchemasByName[service].Schema.Types {
+							if objName == t.Name {
```
KevinXing, line 117: Do you think we can remove this loop and the check?

CodeBrew28: yup! removed!

federation/executor.go
```diff
@@ -291,7 +323,7 @@ func (e *Executor) execute(ctx context.Context, p *Plan, keys []interface{}, opt
 				if _, ok := subPlanMetaData.results[k]; !ok {
 					subPlanMetaData.results[k] = v
 				} else {
-					if k != keyField || v != subPlanMetaData.results[k] {
+					if k != keyField && v == subPlanMetaData.results[k] {
```
KevinXing, line 326: Why make this change?

federation/executor.go
```diff
@@ -105,28 +105,59 @@ func (e *Executor) runOnService(ctx context.Context, service string, typName str
...
+		var ok bool
+		for f, _ := range e.planner.schema.Fields {
+			if f.Type.String() == typName {
+				rootObject, ok = f.Type.(*graphql.Object)
+				if !ok {
+					return nil, oops.Errorf("root object isn't a graphql object")
+				}
+			}
+		}
+
```
KevinXing, line 131: What if rootObject is nil?

federation/executor.go
```diff
@@ -105,28 +105,59 @@ func (e *Executor) runOnService(ctx context.Context, service string, typName str
...
 		selectionSet = &graphql.SelectionSet{
 			Selections: []*graphql.Selection{
 				{
-					Name:  federationField,
-					Alias: federationField,
+					Name:  "__federation",
```
KevinXing, line 151: why don't use the alias `federationField`?

federation/executor.go
```diff
@@ -105,28 +105,59 @@ func (e *Executor) runOnService(ctx context.Context, service string, typName str
...
 					SelectionSet: &graphql.SelectionSet{
 						Selections: []*graphql.Selection{
 							{
-								Name:  typName,
-								Alias: typName,
+								Name:  federatedName,
+								Alias: federatedName,
 								UnparsedArgs: map[string]interface{}{
-									"keys": keys,
+									"keys": []interface{}{newKeys},
```
KevinXing, line 160: Why this is a slice?

federation/executor.go
```diff
@@ -105,28 +105,59 @@ func (e *Executor) runOnService(ctx context.Context, service string, typName str
...
+				rootObject, ok = f.Type.(*graphql.Object)
+				if !ok {
+					return nil, oops.Errorf("root object isn't a graphql object")
+				}
+			}
+		}
+
+		// If it is a federated key on that service, add it to the input args
+		// passed in to the federated field func as one of the federatd keys 
+		for name, key := range keys[0].(map[string]interface{}) {
```
KevinXing, line 134: why only use keys[0]?

federation/executor.go
CodeBrew28: it was from the rebase :( changing it back

graphql/schemabuilder/types.go
```diff
@@ -263,6 +266,27 @@ func (s *Object) ManualPaginationWithFallback(name string, manualPaginatedFunc i
...
+	if s.Methods == nil {
+		s.Methods = make(Methods)
+	}
+
+	m := &method{Fn: f}
+	for _, opt := range options {
+		opt.apply(m)
+	}
+
+	if _, ok := s.Methods[name]; ok {
```
KevinXing, line 279: Do we still need this?

federation/planner.go
```diff
@@ -233,11 +233,41 @@ func (e *Planner) planObject(typ *graphql.Object, selectionSet *graphql.Selectio
...
+				for service := range field.FederatedKey {
+					if len(selectionsByService[service]) > 0 {
+						selections = append(selections, &graphql.Selection{
+							Name:         name,
+							Alias:        name,
+							UnparsedArgs: map[string]interface{}{},
+						})
+						break
+					}
+
```
KevinXing, line 252: Remove the empty line

federation/planner.go
```diff
@@ -104,7 +104,7 @@ func printSelections(selectionSet *graphql.SelectionSet) {
 		fmt.Println(" selections")
 		for _, subSelection := range selectionSet.Selections {
 			fmt.Println(" ", subSelection.Name)
-			if (subSelection.Args != nil) {
+			if subSelection.Args != nil {
```
KevinXing, line 107: Do we want to error out at line 161 if `serviceWithField` is empty?

federation/executor.go
KevinXing: If so, can we change the key to be an inferface{} instead of a slice of interface{}?

federation/executor.go
```diff
@@ -105,15 +105,51 @@ func (e *Executor) runOnService(ctx context.Context, service string, typName str
...
+		// passed in to the federated field func as one of the federatd keys
+		for name, key := range keys[0].(map[string]interface{}) {
+			if name == "__key" {
+				continue
+			}
+			for fieldName, field := range rootObject.Fields {
+				if fieldName == name {
+					_, ok := field.FederatedKey[service]
+					if ok {
+						newKeys[name] = key
```
KevinXing, line 147: why we need to create the `newKeys` map, why cannot just use `keys[0]`?

federation/planner.go
CodeBrew28: This is already verified when we parse the query, but we can definitely also add an error check on line 161 too

federation/schema.go
```diff
@@ -83,6 +92,50 @@ func convertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
...
+							return nil, oops.Errorf("Object %s is not an input object, but it is an argument to the field %s", rootType.Name, field.Name)
+						}
+
+						// Get the object type from the merged schema and loop through all it's fields
+						obj, ok := types[objName].(*graphql.Object)
+						if ok {
+							for fName, f := range obj.Fields {
+								// If the field is one of the input fields to the federatedfieldfunc,
+								// add the service name to the list of federated keys
+								if _, ok := inputType.InputFields[fName]; !ok {
```
KevinXing, line 121: Should we also check if all `inputType.InputFields` have the corresponding `obj.Field`?

CodeBrew28: We're looking through the obj.Fields in the line above, so we've already confirmed it is an field on the object

federation/executor.go
CodeBrew28: I misspoke earlier, the federated field returns a list of objects, so we actually want to parse all the keys. I updated this along with fixing the executor to support this.

CodeBrew28:
@KevinXing I updated this to now allow you to federated an already federated object, could you take another look

KevinXing:
@amyliu345 , FYI, this is the PR for the federation key.

federation/executor.go
```diff
@@ -105,15 +105,58 @@ func (e *Executor) runOnService(ctx context.Context, service string, typName str
...
+		}
+	
+		// If it is a federated key on that service, add it to the input args
+		// passed in to the federated field func as one of the federatd keys
+		newKeys := make([]interface{}, len(keys))
+
+		for i, key := range keys {
+			keyFields, ok := key.(map[string]interface{})
+			if !ok {
+				return nil, oops.Errorf("baddd %s", typName)
```
KevinXing, line 141: more detailed error message?

federation/executor.go
```diff
@@ -274,28 +319,27 @@ func (e *Executor) execute(ctx context.Context, p *Plan, keys []interface{}, opt
...
 			}
 
 			// Acquire mutex lock before modifying results
 			resMu.Lock()
 			defer resMu.Unlock()
-			for k, v := range result {
-				if _, ok := subPlanMetaData.results[k]; !ok {
-					subPlanMetaData.results[k] = v
-				} else {
-					if k != keyField || v != subPlanMetaData.results[k] {
```
KevinXing, line 294: Why remove this check?

graphql/schemabuilder/build.go
```diff
@@ -67,6 +67,7 @@ func (sb *schemaBuilder) getType(nodeType reflect.Type) (graphql.Type, error) {
 		return &graphql.NonNull{Type: sb.types[nodeType]}, nil
 	}
 	if nodeType.Kind() == reflect.Ptr && nodeType.Elem().Kind() == reflect.Struct {
+
```
KevinXing, line 70: remove this ?

federation/executor.go
CodeBrew28: ah oops, added back!

federation/schema.go
```diff
@@ -83,6 +92,44 @@ func convertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
...
+					if len(names) != 2 {
+						return nil, oops.Errorf("Field %s doesnt have an object name and service name", field.Name)
+					}
+					objName := names[0]
+					obj, ok := types[objName].(*graphql.Object)
+					if !ok {
+						return nil, oops.Errorf("Expected objectName %s on merged schema", objName)
+					}
+					for _, arg := range field.Args {
+						rootType := getRootType(arg.Type)
```
KevinXing, line 113: Do we think we could force the rootType to be a subset of obj here?

CodeBrew28: I added a check to ensure this!

KevinXing:
could you please resolve the conflict?

PR #376, opened 2020-06-04
Schema Syncer 

CodeBrew28:
We want to create a schemasyncer in the gateway that listens for changes to the seperate gqlservers. It then can then update its planner to know about the changes whenever any of the gqlservers change without restarting


federation/executor.go
```diff
@@ -39,65 +43,41 @@ func fetchSchema(ctx context.Context, e ExecutorClient, optionalArgs interface{}
...
-		schemas[server] = &iq
+func NewCustomExecutor(ctx context.Context, executors map[string]ExecutorClient, schemaSyncer SchemaSyncer, optionalArgs interface{}, schemaSyncInterval int64) (*Executor, error) {
+	if schemaSyncer == nil {
+		schemaSyncer = newIntrospectionSchemaSyncer(ctx, executors, optionalArgs)
 	}
-
-	types, err := convertSchema(schemas)
-	if err != nil {
-		return nil, oops.Wrapf(err, "converting schemas error")
+	if schemaSyncInterval < 1 {
```
CodeBrew28, line 53: What would be a good number for this? It is something configurable, but it needs a default number

federation/executor.go
```diff
@@ -25,8 +26,11 @@ type ExecutorClient interface {
 // subquery on any of the federated servers.
 // The planner allows it to coordinate the subqueries being sent to the federated servers
 type Executor struct {
-	Executors map[string]ExecutorClient
-	planner   *Planner
+	Executors    map[string]ExecutorClient
+	planner      *Planner
```
KevinXing, line 30: Rename ticker to `schemaSyncerTicker`, or group these three fields into another struct `syncer? `

federation/executor.go
```diff
@@ -39,65 +43,41 @@ func fetchSchema(ctx context.Context, e ExecutorClient, optionalArgs interface{}
...
-			return nil, oops.Wrapf(err, "fetching schema %s", server)
-		}
-
-		var iq introspectionQueryResult
-		if err := json.Unmarshal(schema, &iq); err != nil {
-			return nil, oops.Wrapf(err, "unmarshaling schema %s", server)
-		}
-
-		schemas[server] = &iq
+func NewCustomExecutor(ctx context.Context, executors map[string]ExecutorClient, schemaSyncer SchemaSyncer, optionalArgs interface{}, schemaSyncInterval int64) (*Executor, error) {
```
KevinXing, line 49: What is the unit of the interval, maybe rename it to `schemaSyncInternvalSecond`

federation/executor.go
```diff
@@ -39,65 +43,41 @@ func fetchSchema(ctx context.Context, e ExecutorClient, optionalArgs interface{}
...
+	if schemaSyncer == nil {
```
KevinXing, line 50: hmm, why we need the input parameter?

federation/executor.go
KevinXing: 30 secs? 
I feel the default value maybe better longer than shorter.

federation/executor.go
```diff
@@ -39,65 +43,41 @@ func fetchSchema(ctx context.Context, e ExecutorClient, optionalArgs interface{}
...
+	return executor, nil
+}
 
-	// The planner is aware of the merged schema and what executors
-	// know about what fields
-	planner := &Planner{
-		schema:    types,
-		flattener: flattener,
+func (e *Executor) poll(ctx context.Context) {
+	for {
```
KevinXing, line 71: maybe update it to respect ctx.cancel?

federation/executor.go
```diff
@@ -39,65 +43,41 @@ func fetchSchema(ctx context.Context, e ExecutorClient, optionalArgs interface{}
...
+		select {
+		case <-e.ticker.C:
+			schema, newPlanner, err := e.schemaSyncer.UpdatePlanner(ctx, e.schema)
+			if err == nil && schema != nil && newPlanner != nil {
+				e.schema = schema
+				e.planner = newPlanner
```
KevinXing, line 77: what will happen to the inflight queries?

KevinXing: should we add a lock around the planner?

federation/executor.go
KevinXing: Or put the ticker inside the syncer?

federation/executor.go
```diff
@@ -25,8 +26,11 @@ type ExecutorClient interface {
 // subquery on any of the federated servers.
 // The planner allows it to coordinate the subqueries being sent to the federated servers
 type Executor struct {
-	Executors map[string]ExecutorClient
-	planner   *Planner
+	Executors    map[string]ExecutorClient
+	planner      *Planner
+	ticker       *time.Ticker
+	schemaSyncer SchemaSyncer
+	schema       []byte
```
KevinXing, line 33: what is the purpose of this field?

federation/executor.go
CodeBrew28, line 49: Going to change this to a struct instead of a rigid list of arguments

federation/executor.go
CodeBrew28: To tell if the schema has changed, but I think we can add it to the schema syncer instead and remove it from here

federation/executor.go
CodeBrew28: discussed offline - adding a mutex lock around the planner

CodeBrew28:
@KevinXing updated!

federation/executor.go
```diff
@@ -17,6 +18,7 @@ import (
 const keyField = "__key"
 const federationField = "__federation"
 const typeNameField = "__typeName"
+const minSchemaSyncInterval = 30
```
KevinXing, line 21: Add an unit in the variable name like `minSchemaSyncIntervalSeconds`?

federation/executor.go
```diff
@@ -44,7 +46,26 @@ type ExecutorClient interface {
 // The planner allows it to coordinate the subqueries being sent to the federated servers
 type Executor struct {
 	Executors map[string]ExecutorClient
-	planner   *Planner
+	syncer    *Syncer
+}
+
+type Syncer struct {
```
KevinXing, line 52: Add a comment to describe the Syncer.

federation/schema_syncer.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"encoding/json"
+
+	"time"
+
+	"github.com/samsarahq/go/oops"
+	"github.com/samsarahq/thunder/graphql/introspection"
+)
+
+type SchemaSyncer interface {
```
KevinXing, line 13: Add a comment for the interface. 
Add names to the input argument to make the interface easier to use.

federation/executor.go
```diff
@@ -59,67 +80,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
-		return nil, oops.Wrapf(err, "converting schemas error")
-	}
-
-	introspectionSchema := introspection.BareIntrospectionSchema(types.Schema)
-	introspectionServer := &Server{schema: introspectionSchema}
-
-	executors["introspection"] = &DirectExecutorClient{Client: introspectionServer}
-	schema, err := introspection.RunIntrospectionQuery(introspection.BareIntrospectionSchema(introspectionServer.schema))
+func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Executor, error) {
+	return NewCustomExecutor(ctx, executors, &CustomExecutorArgs{})
```
KevinXing, line 90: nit: change &CustomExecutorArgs{} to nil? And change the NewCustomExecutor to check the pointer?

federation/executor.go
```diff
@@ -59,67 +80,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
 	})
 }
 
-func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Executor, error) {
-	return NewExecutorWithOptionalArgs(ctx, executors, nil)
+type CustomExecutorArgs struct {
+	schemaSyncer              SchemaSyncer
+	optionalArgs              interface{}
+	schemaSyncInternvalSecond int64
```
KevinXing, line 86: Maybe move this field to be a function of the SchemaSyncer interface?

federation/executor.go
```diff
@@ -59,67 +80,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
+	go executor.poll(ctx, c.optionalArgs)
+	return executor, nil
+}
 
-	return &Executor{
-		Executors: executors,
-		planner:   planner,
-	}, nil
+func (e *Executor) poll(ctx context.Context, optionalArgs interface{}) {
+	ctx, cancel := context.WithCancel(context.Background())
```
KevinXing, line 118: Use the input context.

federation/executor.go
```diff
@@ -59,67 +80,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
+	defer cancel()
 
+	for {
+		select {
```
KevinXing, line 122: Check ctx.Done.

federation/executor.go
```diff
@@ -59,67 +80,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
+		case <-e.syncer.ticker.C:
+			newPlanner, err := e.syncer.schemaSyncer.UpdatePlanner(ctx, optionalArgs)
+			if err == nil && newPlanner != nil {
```
KevinXing, line 125: Do we need to check if newPlanner is nil?

federation/executor.go
```diff
@@ -145,7 +152,8 @@ func (e *Executor) runOnService(ctx context.Context, service string, typName str
 
 		var rootObject *graphql.Object
 		var ok bool
-		for f, _ := range e.planner.schema.Fields {
+		p := e.getPlanner()
```
KevinXing, line 155: nit: change the variable to the word `planner`?

federation/executor.go
```diff
@@ -59,67 +80,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
+		case <-e.syncer.ticker.C:
+			newPlanner, err := e.syncer.schemaSyncer.UpdatePlanner(ctx, optionalArgs)
+			if err == nil && newPlanner != nil {
+				e.setPlanner(newPlanner)
+			}
+		}
+	}
 }
 
 func (e *Executor) runOnService(ctx context.Context, service string, typName string, keys []interface{}, kind string, selectionSet *graphql.SelectionSet, optionalArgs interface{}) ([]interface{}, interface{}, error) {
```
KevinXing, line 132: Do you think it is better to change this function to take a planner passed by the Execute function? 
My thought is we want to make sure the planner is consistent during the lifetime of a query.

federation/executor.go
CodeBrew28: Hm, since the ticker is in the executor, would it make sense to keep the sync interval here too? It might get confusing if in a custom syncer there's another ticker for when it fetches the schema file from some external source?

federation/executor.go
```diff
@@ -59,70 +81,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
+	go executor.poll(ctx, c.optionalArgs)
+	return executor, nil
+}
 
-	return &Executor{
-		Executors: executors,
-		planner:   planner,
-	}, nil
+func (e *Executor) poll(ctx context.Context, optionalArgs interface{}) {
+	ctx, cancel := context.WithCancel(ctx)
```
KevinXing, line 115: Ah, I don't think we need this?

federation/executor.go
KevinXing: ping this? 
why we need to check newPlanner != nil?

federation/executor.go
```diff
@@ -59,70 +81,53 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
+			}
+		case <-ctx.Done():
+		}
+	}
 }
 
-func (e *Executor) runOnService(ctx context.Context, service string, typName string, keys []interface{}, kind string, selectionSet *graphql.SelectionSet, optionalArgs interface{}) ([]interface{}, interface{}, error) {
+func (e *Executor) runOnService(ctx context.Context, service string, typName string, keys []interface{}, kind string, selectionSet *graphql.SelectionSet, optionalArgs interface{}, planner *Planner) ([]interface{}, interface{}, error) {
 	// Execute query on specified service
 	schema, ok := e.Executors[service]
```
KevinXing, line 132: Not in this PR, how about we change the variable name to executorClient?

federation/schema_syncer.go
```diff
@@ -0,0 +1,79 @@
...
+	"encoding/json"
+
+	"github.com/samsarahq/go/oops"
+	"github.com/samsarahq/thunder/graphql/introspection"
+)
+
+// SchemaSyncer has a function that checks if the schema has changed,
+// and if so updates the planner in the federated executor
+type SchemaSyncer interface {
+	UpdatePlanner(ctx context.Context, optionalArgs interface{}) (*Planner, error)
```
KevinXing, line 14: I think we need to create an external NewPlanner function?

KevinXing:
:sheep:

KevinXing:
> 🐑

hmm, wrong PR.

federation/executor.go
CodeBrew28: Which variable do you want to rename? planner?

KevinXing: schema.

federation/schema_syncer.go
CodeBrew28: I renamed it to `FetchPlanner` since we also use this function to create a new planner. Let me know if this sounds okay!

federation/executor.go
```diff
@@ -44,7 +46,41 @@ type ExecutorClient interface {
 // The planner allows it to coordinate the subqueries being sent to the federated servers
 type Executor struct {
 	Executors map[string]ExecutorClient
-	planner   *Planner
+	syncer    *Syncer
+}
+
+// Syncer checks if there is a new schema availible and then updates the planner as needed
```
amyliu345, line 52: nit: availible -> available

federation/executor.go
```diff
@@ -59,72 +95,52 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
 	})
 }
 
-func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Executor, error) {
-	return NewExecutorWithOptionalArgs(ctx, executors, nil)
+type CustomExecutorArgs struct {
+	schemaSyncer              SchemaSyncer
+	optionalArgs              interface{}
+	schemaSyncInternvalSecond int64
```
amyliu345, line 101: nit: schemaSyncIntervalSecond instead of schemaSyncInternvalSecond

federation/schema_syncer.go
```diff
@@ -0,0 +1,68 @@
...
+
+	types, err := convertSchema(schemas)
+	if err != nil {
+		return nil, oops.Wrapf(err, "converting schemas error")
+	}
+
+	introspectionSchema := introspection.BareIntrospectionSchema(types.Schema)
+	schema, err := introspection.RunIntrospectionQuery(introspection.BareIntrospectionSchema(introspectionSchema))
+	if err != nil || schema == nil {
+		return nil, oops.Wrapf(err, "error running intrspectin query")
```
amyliu345, line 53: nit: intrspectin -> introspection

federation/executor.go
```diff
@@ -44,7 +46,41 @@ type ExecutorClient interface {
...
+	return e.syncer.planner
+}
+
+func (e *Executor) setPlanner(p *Planner) {
+	e.syncer.plannerMu.Lock()
+	defer e.syncer.plannerMu.Unlock()
+	e.syncer.planner = p
+}
+
+func newPlanner(types *SchemaWithFederationInfo) (*Planner, error) {
```
KevinXing, line 72: Make this public? I think we need to call this in outside of thunder?

federation/executor.go
```diff
@@ -59,72 +95,52 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
-		flattener: flattener,
+func (e *Executor) poll(ctx context.Context, optionalArgs interface{}) {
+	for {
+		select {
+		case <-e.syncer.ticker.C:
+			newPlanner, err := e.syncer.schemaSyncer.FetchPlanner(ctx, optionalArgs)
+			if err == nil && newPlanner != nil {
+				e.setPlanner(newPlanner)
+			}
+		case <-ctx.Done():
```
KevinXing, line 136: Should we return ctx.Err here?

federation/executor.go
```diff
@@ -59,72 +95,52 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
 	})
 }
 
-func NewExecutor(ctx context.Context, executors map[string]ExecutorClient) (*Executor, error) {
-	return NewExecutorWithOptionalArgs(ctx, executors, nil)
+type CustomExecutorArgs struct {
+	schemaSyncer             SchemaSyncer
+	optionalArgs             interface{}
+	schemaSyncIntervalSecond int64
```
KevinXing, line 101: Should we expose these fields? Since we need to use them outside of thunder. 
Also, do you think we can move the `schemaSyncIntervalSecond` and the poller function to the syncer?

federation/executor.go
CodeBrew28: This function doesn't return anything, so I'm not really sure where the error will get reported? Its also run as a separate goroutine? Could you explain more about what we can do to catch the error?

federation/executor.go
CodeBrew28: The executor is the object that has the ticker and is polling the schema syncer so it would be misleading to put it into the schema syncer. The schema syncer is also just an interface with a method to fetch the planner - so I would prefer to leave schemaSyncIntervalSecond in the executor.

federation/executor.go
KevinXing: Then maybe just return for now?

federation/executor.go
KevinXing: Then how about change `schemaSyncIntervalSecond` to a function that returns an int64 which gives the user more flexibility to control the interval realtime?

CodeBrew28: Ah thats a really good point! I added it as a function to schema syncer!

CodeBrew28:
@KevinXing could you take another look?

federation/executor.go
```diff
@@ -59,72 +94,56 @@ func fetchSchema(ctx context.Context, e ExecutorClient, metadata interface{}) (*
...
-	schema, err := introspection.RunIntrospectionQuery(introspection.BareIntrospectionSchema(introspectionServer.schema))
-
-	var iq introspectionQueryResult
-	if err := json.Unmarshal(schema, &iq); err != nil {
-		return nil, oops.Wrapf(err, "unmarshaling introspection schema")
-	}
-
-	schemas["introspection"] = &iq
-	types, err = convertSchema(schemas)
+	interval, err := c.SchemaSyncer.GetRefetchInterval(ctx)
```
KevinXing, line 111: This feels a little weird to me since we have the GetInterval inside the syncer while the poller is outside of the syncer.
I am okay with either to put all the interval and poller stuff to the syncer or the other way.

federation/schema_syncer.go
```diff
@@ -0,0 +1,73 @@
...
+type SchemaSyncer interface {
+	FetchPlanner(ctx context.Context, optionalArgs interface{}) (*Planner, error)
+	GetRefetchInterval(ctx context.Context) (int64, error)
+}
+type IntrospectionSchemaSyncer struct {
+	executors    map[string]ExecutorClient
+	optionalArgs interface{}
+}
+
+func newIntrospectionSchemaSyncer(ctx context.Context, executors map[string]ExecutorClient, optionalArgs interface{}) *IntrospectionSchemaSyncer {
```
KevinXing, line 22: Do you think we want to make it public?

KevinXing: And maybe add a comment to describe the syncer?

CodeBrew28: We can, although I suspect it will never be used since its always set as the default syncer and a developer can choose to write their own syncer if needed

PR #381, opened 2020-06-16
Federation: Support multiple federated keys

CodeBrew28:
A federated server can have multiple keys. We pass only the necessary keys to each service by checking what keys that federated schema knows about.

(This is the same PR as https://github.com/samsarahq/thunder/pull/375)

### DIFF
--- a/federation/schema.go
+++ b/federation/schema.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
+	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/graphql"
 )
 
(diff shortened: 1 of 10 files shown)


KevinXing:
The link is incorrect?

federation/schema.go
```diff
@@ -115,6 +115,14 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
 						if !ok {
 							return nil, oops.Errorf("Object %s is not an input object, but it is an argument to the field %s", rootType.Name, field.Name)
 						}
+
+						// Check that all the input fields are on the federated object
+						for fName := range inputType.InputFields {
+							if _, ok := inputType.InputFields[fName]; !ok {
```
KevinXing, line 121: `if _, ok != obj.Fields[fName]`? 
Do you think we should add an unit test for it?

CodeBrew28: updated and added a unit test

PR #382, opened 2020-06-16
Prototype: Federation API changes

CodeBrew28:
When created a federated field func we use to have to do 
```
schema.Federation().FederatedFieldFunc("Device", func(args struct{ Keys []SafetyDevice }) []*SafetyDevice {
        return args.Keys 
})
```
but since this is always on a federated object we can simplify the api
```
s2.FederatedFieldFunc("Device", func(args struct{ Keys []SafetyDevice }) []*SafetyDevice {
        return args.Keys 
})
```
Also fixed any linting issues in the files changed.

### DIFF
--- a/federation/customexecutor_test.go
+++ b/federation/customexecutor_test.go
@@ -88,11 +88,9 @@ func createFederatedSchema(t *testing.T) *schemabuilder.Schema {
 		Name  string
 	}
 	s1 := schemabuilder.NewSchema()
-	user := s1.Object("User", User{})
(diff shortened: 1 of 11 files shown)


KevinXing:
The code before and after is the same in the description?

CodeBrew28:
They're not the same, the before is `schema.Federation().FederatedFieldFunc` and the after is just `schema.FederatedFieldFunc`

KevinXing:
Ahh i see.

KevinXing:
I think we can close this since we will get rid of this API?

PR #385, opened 2020-06-19
Simplify api for federating a object 

CodeBrew28:
When we create a an object that can be federated, we use to have to specify a function to expose certain fields 
```
user := s1.Object("User", User{})
user.Federation(func(u *User) *User {
        return u
})
```

but now we can pass in an option to the object to autogenerate this function in the schema

```
user := s1.Object("User", User{}, schemabuilder.RootObject)
```

We want developers to add this so we are aware of what objects are federated and what the root types are instead of making everything a federated object.

### DIFF
--- a/graphql/schemabuilder/function.go
+++ b/graphql/schemabuilder/function.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/graphql"
 )
 
@@ -12,6 +13,12 @@ import (
 // that object to build a GraphQL Field that can be resolved in the GraphQL
 // graph.
 func (sb *schemaBuilder) buildFunction(typ reflect.Type, m *method) (*graphql.Field, error) {
+	// If the method is federated, we want to built a graphql field that returns all
+	// fields on that object. This allows them to be sent as federated keys to other servers.
+	if m.FederationType != nil {
+		return sb.buildFederatedFunction(typ, m)
+	}
+
 	field, _, err := sb.buildFunctionAndFuncCtx(typ, m)
 	return field, err
 }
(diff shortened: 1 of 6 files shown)


graphql/schemabuilder/function.go
```diff
@@ -5,13 +5,37 @@ import (
...
 )
 
 // buildFunction takes the reflect type of an object and a method attached to
 // that object to build a GraphQL Field that can be resolved in the GraphQL
 // graph.
 func (sb *schemaBuilder) buildFunction(typ reflect.Type, m *method) (*graphql.Field, error) {
+	// If the method is federated, we want to built a graphql field that returns all
+	// fields on that object. This allows them to be sent as federated keys to other servers.
+	if m.FederationType != nil {
+		var argParser *argParser
```
KevinXing, line 19: Wrap this into a helper function?

graphql/schemabuilder/types.go
```diff
@@ -317,6 +316,8 @@ type method struct {
 	BatchArgs batchArgs
 
 	ManualPaginationArgs manualPaginationArgs
+
+	FederationType interface{}
```
KevinXing, line 320: Could you please add a comment to describe this field?

graphql/schemabuilder/schema.go
```diff
@@ -104,24 +105,60 @@ func getEnumMap(enumMap interface{}, typ reflect.Type) (map[string]interface{},
...
+	apply(*Object)
+}
+
+// objectOptionFunc is a helper to define ObjectOptions when creating an object
+type objectOptionFunc func(*Object)
+
+func (f objectOptionFunc) apply(m *Object) { f(m) }
+
+// RootObject is an option that can be passed to a Object to indicate that
+// its return value is a root object tha
```
KevinXing, line 120: The comment is incomplete?

CodeBrew28:
Yes! 
1. This should be safe since in our current world we aren't federating anything. We only want to add the `schemabuilder.RootObject` annotation when we move a fieldfunc to another server.
2. This is an annotation we add to the root object since this exposes the object struct's fields as federated fields. All the federated services request these fields using the `FedeatedFieldFunc`.

So in the example below, it defines the field to resolve `__federation`
```

{
        users {
            __federation {
               id
               orgId
            }
        }
}
```

PR #386, opened 2020-06-19
Federated Field Func Api Simplification

CodeBrew28:
Federated field funcs are always nested on a root query, so we can simplify the devleoper-facing API. Instead of `schema.Federation().FederatedFieldFunc` we can just do `schema.FederatedFieldFunc`. We want this to be exposed such that users can override it as needed with a custom federated field func. 

However, if you pass in `schemabuilder.ShadowObject` as an option to the field func, it will autogenerate the `FederatedFieldFunc` method for us


graphql/schemabuilder/types.go
```diff
@@ -265,19 +265,30 @@ func (s *Object) ManualPaginationWithFallback(name string, manualPaginatedFunc i
 	s.Methods[name] = m
 }
 
-func (s *Object) FederatedFieldFunc(name string, f interface{}, options ...FieldFuncOption) {
-	if s.Methods == nil {
-		s.Methods = make(Methods)
+type federation struct{}
+
+func (s *Schema) FederatedFieldFunc(name string, f interface{}, options ...FieldFuncOption) {
+	// Create a field func called "__federation" on the root query object
```
KevinXing, line 272: nit: move this before the function?

graphql/schemabuilder/schema.go
```diff
@@ -150,15 +154,38 @@ func (s *Schema) Object(name string, typ interface{}, options ...ObjectOption) *
 		if object.Methods == nil {
 			object.Methods = make(Methods)
 		}
-
-		federatedMethod := &method{}
+		federatedMethod := &method{
```
KevinXing, line 157: How do you feel rename `isFederated` to `isRoot`?

graphql/schemabuilder/schema.go
```diff
@@ -150,15 +154,38 @@ func (s *Schema) Object(name string, typ interface{}, options ...ObjectOption) *
...
 	}
 
+	if object.IsShadow {
+		// Create federation object on the root query
+		q := s.Query()
+		if _, ok := q.Methods["__federation"]; !ok {
+			q.FieldFunc("__federation", func() federation { return federation{} })
+		}
+		federationObject := s.Object("Federation", federation{})
+		federatedObjectType := reflect.New(reflect.TypeOf(typ)).Interface()
```
KevinXing, line 173: so this can be rootObjectType which seems easier to understand for me.

graphql/schemabuilder/schema.go
```diff
@@ -150,15 +154,38 @@ func (s *Schema) Object(name string, typ interface{}, options ...ObjectOption) *
 		if object.Methods == nil {
 			object.Methods = make(Methods)
 		}
-
-		federatedMethod := &method{}
+		federatedMethod := &method{
+			FederationType: federatedObjectType,
```
KevinXing, line 158: and this to `RootObjectType`, then we also add a `ShadowObjectType` field? 
I think currently `IsRoot` and `IsShadow` shares this field which seems a bit confusion.

graphql/schemabuilder/function.go
```diff
@@ -16,7 +16,11 @@ func (sb *schemaBuilder) buildFunction(typ reflect.Type, m *method) (*graphql.Fi
 	// If the method is federated, we want to built a graphql field that returns all
 	// fields on that object. This allows them to be sent as federated keys to other servers.
 	if m.FederationType != nil {
-		return sb.buildFederatedFunction(typ, m)
+		if m.IsShadow {
```
KevinXing, line 19: Do you think we need to check if `typ` is `federation` where to make sure we only build the function to `federation` object?

graphql/schemabuilder/function.go
```diff
@@ -111,6 +115,48 @@ func (sb *schemaBuilder) buildFederatedFunction(typ reflect.Type, m *method) (*g
 	return field, nil
 }
 
+func (sb *schemaBuilder) buildShadowObjectFederationFunction(typ reflect.Type, m *method) (*graphql.Field, error) {
```
KevinXing, line 118: Add a comment to say this build a fieldfunc to `federation` object and the signature of the functions is 
blahblah?

graphql/schemabuilder/function.go
```diff
@@ -111,6 +115,48 @@ func (sb *schemaBuilder) buildFederatedFunction(typ reflect.Type, m *method) (*g
...
+			Name: "Keys",
+			Type: reflect.SliceOf(shadowType),
+		},
+	})
+	i := make([]reflect.Type, 1)
+	i[0] = input
+
+	argParser, argType, _, err := funcCtx.getArgParserAndTyp(sb, i)
+	if err != nil {
+		return nil, err
```
KevinXing, line 134: wrap the error and below?

graphql/schemabuilder/function.go
```diff
@@ -111,6 +115,48 @@ func (sb *schemaBuilder) buildFederatedFunction(typ reflect.Type, m *method) (*g
...
+	}
+	funcCtx.hasArgs = argParser != nil
+	args, err := funcCtx.argsTypeMap(argType)
+	if err != nil {
+		return nil, err
+	}
+	returnType, err := sb.getType(reflect.TypeOf(m.FederationType))
```
KevinXing, line 141: I think this should be something like `rType := &graphql.NonNull{Type: &graphql.List{Type: retType}}`
unless thunder will think the return type is `device{}` rather than `[]*device` and then pass the entire slice rather than individual element to the downstream field func.

federation/schema.go
```diff
@@ -118,6 +120,36 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
 
 						// Check that all the input fields are on the federated object
 						for fName := range inputType.InputFields {
+
+							validFederatedKey := false
```
KevinXing, line 124: wrap this into a helper?

graphql/schemabuilder/schema.go
KevinXing: ping this.

graphql/schemabuilder/function.go
```diff
@@ -13,12 +13,19 @@ import (
...
+
 	// If the method is federated, we want to built a graphql field that returns all
 	// fields on that object. This allows them to be sent as federated keys to other servers.
 	if m.FederationType != nil {
 		return sb.buildFederatedFunction(typ, m)
 	}
 
+	// If the method is a shadow object, we want to built a graphql field that parses the federation
+	// keys and reconstructs the object to run a subquery on a federated servers
+	if m.ShadowObjectType != nil && typ.Name() == "federation" {
```
KevinXing, line 25: should we error out if typ.Name() != "federation"?

graphql/schemabuilder/function.go
KevinXing: Maybe augment the comment to include the shape of the generated function like 
```
This generates a function 
federation.fieldfunc(blahblah)

File: graphql/schemabuilder/function.go [hunk 11]
```diff
@@ -111,6 +118,62 @@ func (sb *schemaBuilder) buildFederatedFunction(typ reflect.Type, m *method) (*g
...
+	returnType, err := sb.getType(reflect.TypeOf(m.ShadowObjectType))
+	if err != nil {
+		return nil, oops.Wrapf(err, "Invalid return type")
+	}
+	rType := &graphql.NonNull{Type: &graphql.List{Type: returnType}}
+
+	field := &graphql.Field{
+		Resolve: func(ctx context.Context, source, funcRawArgs interface{}, selectionSet *graphql.SelectionSet) (interface{}, error) {
+			args := funcRawArgs.(map[string]graphql.Type)
+			return args["keys"], nil
```
KevinXing, line 165: Just curious, the funcRawRags has a non pointer type and our return type is a pointer, is that okay?

federation/schema.go
```diff
@@ -118,6 +152,37 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
 
 						// Check that all the input fields are on the federated object
 						for fName := range inputType.InputFields {
+
+							err := validateFederationKeys(serviceNames, serviceSchemasByName, obj, fName)
+							// validFederatedKey := false
```
KevinXing, line 157: Remove the comments?

federation/schema.go
```diff
@@ -39,6 +39,38 @@ func getRootType(typ *introspectionTypeRef) *introspectionTypeRef {
 	return getRootType(typ.OfType)
 }
 
+func validateFederationKeys(serviceNames []string, serviceSchemasByName map[string]*introspectionQueryResult, obj *graphql.Object, fName string) error {
```
KevinXing, line 42: Could you describe why we need this check and how this check could fail?

federation/executor_test.go
```diff
@@ -45,9 +45,11 @@ func createExecutorWithFederatedUser() (*Executor, *schemabuilder.Schema, *schem
 		}
 	*/
 	type User struct {
-		Id    int64
-		OrgId int64
-		Name  string
+		Id          int64
```
KevinXing, line 48: Maybe also update the comment message above?

graphql/schemabuilder/types.go
```diff
@@ -15,7 +15,7 @@ type Object struct {
 	Methods     Methods // Deprecated, use FieldFunc instead.
 	key         string
 	ServiceName string
-	IsFederated bool
+	IsRoot      bool
```
KevinXing, line 18: Should we move this change to one of the previous commit?

graphql/schemabuilder/function.go
CodeBrew28: Yup I think so since the funcRawRags are nonnullable

graphql/schemabuilder/types.go
CodeBrew28: yes!

graphql/schemabuilder/function.go
KevinXing: Ah sorry what i meant is our arg looks like this 
struct {
    keys: []Device
}
whereas the `rType` is `[]*Device`.
Is that okay or we need to do some further translation?

CodeBrew28: Ah, its working right now since the return type isn't expecting a pointer. We can change it so that they both use a pointer

KevinXing: Does that matter? 
I think executor is smart enough to hand both cases?

KevinXing: But it should be more efficient if we can return a pointer?

KevinXing: How about we change both input and output to be a pointer?

KevinXing: Is there any rule that the input keys could not be pointers?

CodeBrew28: Ah, actually I misread the code, both the input type and return type have a pointer to the device!

graphql/schemabuilder/schema.go
```diff
@@ -145,20 +149,42 @@ func (s *Schema) Object(name string, typ interface{}, options ...ObjectOption) *
...
 	}
 
+	if object.IsShadow {
+		// Create federation object on the root query
+		q := s.Query()
+		if _, ok := q.Methods["__federation"]; !ok {
+			q.FieldFunc("__federation", func() federation { return federation{} })
+		}
+		federationObject := s.Object("Federation", federation{})
+		rootObjectType := reflect.New(reflect.TypeOf(typ)).Interface()
```
KevinXing, line 173: s/rootObjectType/shadowObjectType/ ?

graphql/schemabuilder/types.go
```diff
@@ -331,6 +336,10 @@ type method struct {
 	// FederationType is an object where all the fields are keys
 	// that can be exposed over federation
 	FederationType interface{}
+
+	// ShadowObjectType is an object where all the fields are keys
+	// that can be fetched in a federated subquery
+	ShadowObjectType interface{}
```
KevinXing, line 342: Maybe change the comment to say 
"ShadowObjectType is the reflect type of parent object if it is a shadow object" ?

graphql/schemabuilder/types.go
```diff
@@ -331,6 +336,10 @@ type method struct {
 	// FederationType is an object where all the fields are keys
 	// that can be exposed over federation
 	FederationType interface{}
```
KevinXing, line 338: How about change this to RootObjectType and add a comment to say
"RootObjectType is the reflect type of parent object if it is a root object" ?

KevinXing: The motivation is the word federation is kind of confusion to me.

graphql/schemabuilder/types.go
```diff
@@ -388,7 +397,7 @@ func (s *Schema) FederatedObject(name string, typ interface{}) *Object {
 		Name:        name,
 		Type:        typ,
 		ServiceName: s.Name,
-		IsFederated: true,
+		IsRoot:      true,
```
KevinXing, line 400: I feel we don't need this function anymore?

graphql/schemabuilder/function.go
KevinXing: hmm, ping this?

graphql/schemabuilder/types.go
KevinXing: Also, do you think we can change the type of this field and the field above to reflect.Type? why we need it to be an interface here?

graphql/schemabuilder/function.go
```diff
@@ -111,6 +118,62 @@ func (sb *schemaBuilder) buildFederatedFunction(typ reflect.Type, m *method) (*g
...
+	in[0] = input
+
+	argParser, argType, _, err := funcCtx.getArgParserAndTyp(sb, in)
+	if err != nil {
+		return nil, oops.Wrapf(err, "Error parsing args for shadow object field")
+	}
+	funcCtx.hasArgs = argParser != nil
+	args, err := funcCtx.argsTypeMap(argType)
+	if err != nil {
+		return nil, err
```
KevinXing, line 152: wrap the error here?

graphql/schemabuilder/function.go
```diff
@@ -111,6 +118,62 @@ func (sb *schemaBuilder) buildFederatedFunction(typ reflect.Type, m *method) (*g
...
+	}
+
+	// Return type is a nonnullable list of the shadow object type
+	returnType, err := sb.getType(reflect.TypeOf(m.ShadowObjectType))
+	if err != nil {
+		return nil, oops.Wrapf(err, "Invalid return type")
+	}
+	rType := &graphql.NonNull{Type: &graphql.List{Type: returnType}}
+	fmt.Println(rType, args)
```
KevinXing, line 161: Remove the print?

graphql/schemabuilder/function.go
```diff
@@ -111,6 +118,62 @@ func (sb *schemaBuilder) buildFederatedFunction(typ reflect.Type, m *method) (*g
 	return field, nil
 }
 
+// buildShadowObjectFederationFunction builds a federation object and a field func that takes the
+// federation keys as args and constructs the shadow object. This is used for federated subqueries.
+// {
+//    __federation {
+//     [ObjectName]-[Service] (keys: Keys) {
+//       subQuery
+// 		}
```
KevinXing, line 127: Fix the intention?

federation/schema.go
```diff
@@ -118,6 +155,12 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
 
 						// Check that all the input fields are on the federated object
 						for fName := range inputType.InputFields {
+
+							err := validateFederationKeys(serviceNames, serviceSchemasByName, obj, fName)
+							if err != nil {
+								return nil, err
```
KevinXing, line 161: wrap the error?
nit: 
```
if err := xxx; err != nil {}
```

federation/schema.go
```diff
@@ -39,6 +39,41 @@ func getRootType(typ *introspectionTypeRef) *introspectionTypeRef {
...
+// that have the objcet registered as a root object expose the field. This ensures that we can make 
+// the hop from the root server to any of the federated servers safely before any queries are executed. 
+func validateFederationKeys(serviceNames []string, serviceSchemasByName map[string]*introspectionQueryResult, obj *graphql.Object, fName string) error {
+	validFederatedKey := false
+	for _, service := range serviceNames {
+		for _, typ := range serviceSchemasByName[service].Schema.Types {
+			if typ.Name == obj.Name {
+				// Check that it is a root object by checking if it has a field func called
+				// "__federation" on the object
+				isFederatedFromTheService := false
```
KevinXing, line 52: How about change the variable to `isRootObject`?

federation/schema.go
```diff
@@ -39,6 +39,41 @@ func getRootType(typ *introspectionTypeRef) *introspectionTypeRef {
...
+				for _, introspectedField := range typ.Fields {
+					if introspectedField.Name == federationField {
+						isFederatedFromTheService = true
```
KevinXing, line 55: We can break here?

federation/schema.go
```diff
@@ -39,6 +39,41 @@ func getRootType(typ *introspectionTypeRef) *introspectionTypeRef {
...
+					}
+				}
+				// If it is a root object, check that it has all the fields being requested
+				// as a federated key
+				if isFederatedFromTheService {
+					for _, introspectedField := range typ.Fields {
+						if introspectedField.Name == fName {
```
KevinXing, line 62: rename fName to something else? maybe keyField?

federation/schema.go
```diff
@@ -39,6 +39,41 @@ func getRootType(typ *introspectionTypeRef) *introspectionTypeRef {
...
+							validFederatedKey = true
```
KevinXing, line 63: break here?

federation/schema.go
```diff
@@ -39,6 +39,41 @@ func getRootType(typ *introspectionTypeRef) *introspectionTypeRef {
...
+						}
+					}
+				}
+
+			}
+		}
+	}
+	if !validFederatedKey {
+		return oops.Errorf("Invalid federation key")
```
KevinXing, line 72: Include the key name?

federation/executor_test.go
```diff
@@ -22,6 +22,8 @@ func createExecutorWithFederatedUser() (*Executor, *schemabuilder.Schema, *schem
 				id: int64!
```
KevinXing, line 22: Do you think we need a test case to test the key validatoin?

graphql/schemabuilder/schema.go
```diff
@@ -145,20 +149,40 @@ func (s *Schema) Object(name string, typ interface{}, options ...ObjectOption) *
...
 		}
-		federatedMethod.FederationType = federatedObjectType
-		object.Methods[federationField] = federatedMethod
+		object.Methods[federationField] = rootMethod
 	}
 
+	if object.IsShadow {
+		// Create federation object on the root query
+		q := s.Query()
+		if _, ok := q.Methods["__federation"]; !ok {
```
KevinXing, line 169: Nit: s/__federation/federationField/

KevinXing: same below.

graphql/schemabuilder/function.go
KevinXing: ping this.

graphql/schemabuilder/function.go
KevinXing: Do you think we want to do ^?

graphql/schemabuilder/function.go
KevinXing: ping.

federation/executor_test.go
CodeBrew28: added!

graphql/schemabuilder/function.go
CodeBrew28: Hm, I fixed this is the last commit, maybe its a github UI thing? I tried reformatting again

CodeBrew28:
Yup! This only adds features to the new federation API that nothing currently uses!

PR #388, opened 2020-06-24
Kevinxing/federation migration

KevinXing:
This commit add a ServiceSelector in planner which could be used to help fieldfunc migration. 
See commit message and comments for more details.


federation/planner.go
```diff
@@ -56,8 +57,30 @@ type Plan struct {
...
+	// schema describes what fields the graphql servers know about along with the services that know how to execute each field.
+	schema *SchemaWithFederationInfo
+	// flattener knows how to combine all the fragments on a query into a singel query.
+	flattener       *flattener
+	serviceSelector ServiceSelector
+}
+
+// serviceSelector is an optional field which can be used to override the <type,field> to service mapping,
+// which is useful during the fieldfunc migration.
+type ServiceSelector func(typeName string, fieldName string) string
```
CodeBrew28, line 69: Can we allow the service selector to return a descriptive error instead of an empty string when things go wrong

federation/planner.go
```diff
@@ -116,6 +139,30 @@ func printSelections(selectionSet *graphql.SelectionSet) {
...
+func (e *Planner) maybeCustomService(
+	typeName, currentService string,
+	selection *graphql.Selection,
+	localSelections []*graphql.Selection,
+	selectionsByService map[string][]*graphql.Selection,
+) (bool, []*graphql.Selection, map[string][]*graphql.Selection) {
+	if e.serviceSelector == nil {
+		return false, localSelections, selectionsByService
+	}
+	customService := e.serviceSelector(typeName, selection.Name)
```
CodeBrew28, line 155: Should the custom service ever be an empty string? Will we just ignore the field if it is an empty string? I think I would prefer to fail the planning step of executing a query

federation/planner.go
```diff
@@ -145,22 +192,27 @@ func (e *Planner) planObject(typ *graphql.Object, selectionSet *graphql.Selectio
 			return nil, fmt.Errorf("typ %s has no field %s", typ.Name, selection.Name)
 		}
 
-		fieldInfo := e.schema.Fields[field]
+		var hasCustomService bool
```
CodeBrew28, line 195: hasCustomService seems confusing - this seems to be false only if we see an error. Could we rename to err?

federation/planner.go
```diff
@@ -145,22 +192,27 @@ func (e *Planner) planObject(typ *graphql.Object, selectionSet *graphql.Selectio
...
-			serviceWithField := ""
+		if !hasCustomService {
+			fieldInfo := e.schema.Fields[field]
 
-			for service, hasField := range fieldInfo.Services {
-				if hasField {
-					serviceWithField = service
+			// Prioritize resolving as many fields as we can in the current service
+			if fieldInfo.Services[service] {
+				localSelections = append(localSelections, selection)
```
CodeBrew28, line 203: I think we're already adding the selection to local selections in `maybeCustomService` if it is in the current service. Should we skip this step?

CodeBrew28: Although I think I might prefer if `maybeCustomService` didn't change the local selections and this was only something we do in this function. That makes it easier to think of `maybeCustomService` as deciding which service to run the field on.

federation/planner.go
KevinXing: If the customer service is an empty string, it means that we don't have a custom config to the {typeName, selection} pair, and the first return value will be false. At line 198, we will continue the normal service assignment.

federation/planner.go
KevinXing: I don't think it should be an error. 
This variable will be false for most of the time unless there is a migration going on.

federation/planner.go
KevinXing: If `hasCustomService` is true, we will skip the entire step.

federation/planner.go
KevinXing: I respond this in the comment below.

federation/planner.go
CodeBrew28: Ahh I see, I misunderstood what hasCustomService did

federation/planner.go
CodeBrew28: I think we want to enforce that if the field exists on multiple services, the service selections can't return an empty string, and instead should have an error.

KevinXing: Talked offline, I added a check for the case when there are more than one service associates with the selection field without a serviceSelecor, the planner will return an error. Also updated the unit test.

federation/planner.go
```diff
@@ -116,6 +139,30 @@ func printSelections(selectionSet *graphql.SelectionSet) {
...
+	selectionsByService map[string][]*graphql.Selection,
+) (bool, []*graphql.Selection, map[string][]*graphql.Selection) {
+	if e.serviceSelector == nil {
+		return false, localSelections, selectionsByService
+	}
+	customService := e.serviceSelector(typeName, selection.Name)
+	if customService == "" {
+		return false, localSelections, selectionsByService
+	}
+	if customService == currentService {
```
CodeBrew28, line 159: Can we make it such that `maybeCustomService` doesn't actually change the local selections or selectionsByService and returns the custom service. This makes it to understand since `maybeCustomService` isn't mutating any of the local fields and`planObject` can be responsible for all the changes to local selections or selections by service.

federation/planner.go
```diff
@@ -145,22 +192,34 @@ func (e *Planner) planObject(typ *graphql.Object, selectionSet *graphql.Selectio
...
+			localSelections,
+			selectionsByService)
+
+		if !hasCustomService {
+			fieldInfo := e.schema.Fields[field]
+			// Prioritize resolving as many fields as we can in the current service
+			if fieldInfo.Services[service] {
+				localSelections = append(localSelections, selection)
+			} else {
+				if len(fieldInfo.Services) != 1 {
```
CodeBrew28, line 209: I think we want this check outside of the else statement. Even if we're prioritizing adding it to the current service, we still want the check of whether or not the field is on multiple services and doesn't have a service selector

federation/planner.go
```diff
@@ -145,22 +192,34 @@ func (e *Planner) planObject(typ *graphql.Object, selectionSet *graphql.Selectio
...
 
-		fieldInfo := e.schema.Fields[field]
-
-		// Prioritize resolving as many fields as we can in the current service
-		if fieldInfo.Services[service] {
-			localSelections = append(localSelections, selection)
-		} else {
-			serviceWithField := ""
+		var hasCustomService bool
+		hasCustomService, localSelections, selectionsByService = e.maybeCustomService(
```
CodeBrew28, line 196: What do you think about changing`maybeCustomService` to `isMigratingField` to make it more clear that this is only true in the case of a field migration. 

Custom service seems kinda confusing to me since if you're moving it from one team's server to a smaller subteams server, you're moving to from one custom server to another custom server. I seem to think of all the federated servers as custom servers, but I want the naming to tell me that this is specifically used for migrating a field.

KevinXing: sg, changed the name.

federation/planner.go
KevinXing: Sure, changed!

federation/planner.go
KevinXing: We talked offline, move this check out of the branch.

federation/planner.go
KevinXing: Actually I changed it back, the reason is the check on line 221 will fail if we don't change the local selections return a false while we are migrating the field.

CodeBrew28: I meant changing it such that the function we call tells us what service to execute the query on and does all th evalidations, but doesn't actually change the list of selections. This makes the code a lot easier to follow in terms of understanding where the list of selections are changing. I put in an example below
```
func (e *Planner) selectService(
	typeName, currentService string,
	selection *graphql.Selection,
	field *graphql.Field,
	fieldInfo *FieldInfo,
) (error, string) {
	customService := ""
	if e.serviceSelector != nil {
		customService = e.serviceSelector(typeName, selection.Name)
	}
	if customService == "" {
		if len(field.FederatedKey) == 0 && len(fieldInfo.Services) != 1 {
			return oops.Errorf("field %s is not in serviceSelector and has %d associate services, current service %s", selection.Name, len(fieldInfo.Services), currentService), ""
		}
		if fieldInfo.Services[currentService] {
			return nil, currentService
		}
		for service, hasField := range fieldInfo.Services {
			if hasField {
				return nil, service
			}
		}
		return oops.Errorf("Field is not on multiple services"), ""
	}
	if len(field.FederatedKey) > 0 {
		return oops.Errorf("It is not allowed to select service for non field func %s for type %s", selection.Name, typeName), ""
	}
	if _, ok := fieldInfo.Services[customService]; !ok {
		return oops.Errorf("no service %s", customService), ""
	}
	return nil, customService
}

... 

err, customService := e.selectService(typ.Name, service, selection, field, fieldInfo)
		if err != nil {
			return nil, oops.Wrapf(err, "selecting service")
		}
		if customService == service {
			localSelections = append(localSelections, selection)
		} else {
			selectionsByService[customService] = append(selectionsByService[customService], selection)
		}
```

KevinXing: Ah, I see, changed according to the comment!

graphql/schemabuilder/input.go
```diff
@@ -129,7 +129,7 @@ func (sb *schemaBuilder) getStructObjectFields(typ reflect.Type) (*graphql.Input
 	}
 
 	if typ.Kind() != reflect.Struct {
-		return nil, nil, fmt.Errorf("expected struct but received type %s", typ.Name())
+		return nil, nil, fmt.Errorf("expected struct but received type %s", typ.Kind())
```
CodeBrew28, line 132: Nice catch!

KevinXing:
> lgtm! One test failing relating to mutations, it just needs the updated schema selector!

Updated!

PR #389, opened 2020-06-25
sqlgen Add batch insert API support to sqlgen

abhinavbatra:
Current sqlgen API only allows inserting one row at a time which makes running large backfills very time consuming.
The intention of the batch API is to allow running backfills at a faster rate

### DIFF
--- a/sqlgen/db.go
+++ b/sqlgen/db.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/samsarahq/go/oops"
@@ -461,6 +462,68 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
 	}
 
 	return db.execWithTrace(ctx, query, "InsertRow")
+}
+
+// InsertRows inserts multiple rows into the database, chunksize rows at a time.
+// Most SQL db enforce a limit on max size of packet, which is why we need to break
+// the rows into chunks.
+//
+// rows should be an array of pointers to a struct, for example:
+//
+//   user1 := &User{Name: "foo"}
+//   user2 := &User{Name: "fan"}
+//   if err := db.InsertRows(ctx, [](*User){user1, user2}, 100); err != nil {
+//
+func (db *DB) InsertRows(ctx context.Context, rows interface{}, chunkSize int) error {
+	val := reflect.ValueOf(rows)
+	kind := val.Kind()
+	if kind != reflect.Slice && kind != reflect.Array {
+		return fmt.Errorf("expect array/slice got %s", val.Kind().String())
+	}
+
+	rowsData := make([]interface{}, val.Len())
+	for i := 0; i < val.Len(); i++ {
+		rowsData[i] = val.Index(i).Interface()
+	}
+
+	var tx *sql.Tx
+	if !db.HasTx(ctx) {
+		var err error
+		ctx, tx, err = db.WithTx(ctx)
+		if err != nil {
+			return err
+		}
+		defer tx.Rollback()
+	}
+
+	for j := 0; j < len(rowsData); j += chunkSize {
+		sliceLength := chunkSize
+		if len(rowsData) < j+sliceLength {
+			sliceLength = len(rowsData) - j
+		}
+		slice := rowsData[j : j+sliceLength]
+
+		query, err := db.Schema.MakeBatchInsertRow(slice)
+		if err != nil {
+			return err
+		}
+
+		numColumns := len(query.Columns)
+		for i := 0; i < len(slice); i++ {
+			if err := db.checkColumnValuesAgainstLimits(ctx, query, query.Columns, query.Values[i*numColumns:(i+1)*numColumns], query.Table); err != nil {
+				return err
+			}
+		}
+		_, err = db.execWithTrace(ctx, query, "InsertRows")
+		if err != nil {
+			return err
+		}
+	}
+
+	if tx != nil {
+		return tx.Commit()
+	}
+	return nil
 }
 
 // UpsertRow inserts a single row into the database
--- a/sqlgen/db_test.go
+++ b/sqlgen/db_test.go
@@ -162,6 +162,67 @@ func TestShardLimitInsert(t *testing.T) {
 	}
 }
 
+func TestShardLimitBatchInsert(t *testing.T) {
+	testcases := getLimitTestcases(t, "BatchInsert")
+	for _, testcase := range testcases {
+		t.Run(testcase.title, func(t *testing.T) {
+			tdb, db, err := setup()
+			assert.NoError(t, err)
+			defer tdb.Close()
+			ctx := context.Background()
+
+			aliceDb := testcase.withFilterLimit(db, Filter{
+				"name": "Alice",
+			})
+
+			alice := &User{Name: "Alice"}
+			bob := &User{Name: "Bob"}
+
+			// Check aliceDb can insert alice.
+			err = aliceDb.InsertRows(ctx, [](*User){alice}, 100)
+			assert.NoError(t, err)
+
+			var user *User
+			err = aliceDb.QueryRow(ctx, &user, Filter{"name": "Alice"}, nil)
+			assert.NoError(t, err)
+
+			// Check aliceDb can't insert alice and bob.
+			err = aliceDb.InsertRows(ctx, [](*User){alice, bob}, 100)
+			assert.Contains(t, err.Error(), "db requies name = Alice, but query has name = Bob")
+
+			tom := &User{Name: "Tom"}
+			jerry := &User{Name: "Jerry"}
+			// Check db can insert tom and jerry with a batch size of 1.
+			err = db.InsertRows(ctx, [](*User){tom, jerry}, 1)
+			assert.NoError(t, err)
+
+			err = db.QueryRow(ctx, &user, Filter{"name": "Tom"}, nil)
+			assert.NoError(t, err)
+			err = db.QueryRow(ctx, &user, Filter{"name": "Jerry"}, nil)
+			assert.NoError(t, err)
+
+			// Check db cannot query bob
+			err = db.QueryRow(ctx, &user, Filter{"name": "Bob"}, nil)
+			assert.Contains(t, err.Error(), "sql: no rows in result set")
+
+			// Check db can insert bob within a transaction
+			transactionCtx, tx, err := db.WithTx(ctx)
+			assert.NoError(t, err)
+			defer tx.Rollback()
+
+			err = db.InsertRows(transactionCtx, [](*User){bob}, 100)
+			assert.NoError(t, err)
+
+			assert.NoError(t, tx.Commit())
+
+			err = db.QueryRow(ctx, &user, Filter{"name": "Alice"}, nil)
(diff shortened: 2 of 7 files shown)


sqlgen/db.go
```diff
@@ -361,6 +361,30 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
 	return db.execWithTrace(ctx, query, "InsertRow")
 }
 
+// InsertRows inserts multiple rows into the database
+//
+// rows should be an array of pointers to a struct, for example:
+//
+//   user1 := &User{Name: "foo"}
+//	 user2 := &User{Name: "fan"}
```
changpingc, line 369: this indentation look weird

sqlgen/db.go
```diff
@@ -361,6 +361,30 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
...
+//   if err := db.InsertRow(ctx, []{interface{}}{user1, user2}); err != nil {
```
changpingc, line 370: InsertRows

sqlgen/db.go
```diff
@@ -361,6 +361,30 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
...
+//
+func (db *DB) InsertRows(ctx context.Context, rows []interface{}) (sql.Result, error) {
```
changpingc, line 372: one minor concern i have is i believe bulk insert is bound by max_allowed_packet configuration which according to this [post](https://aws.amazon.com/blogs/database/best-practices-for-amazon-aurora-mysql-database-configuration/) is 4MB by default on aurora mysql. having this API makes it more likely to accidentally hit this limit and likely we won't uncover this bug until we hit production as local dataset might not be large enough to reflect large customer experience.

therefore, we probably want to always chunk the input to InsertRows to make it safer. it's still not fool-proof since rows can be arbitrarily wide. perhaps there's a way to enforce chunking? what if this API takes a chunk size parameter?

sqlgen/reflect.go
```diff
@@ -565,6 +565,52 @@ func (s *Schema) MakeInsertRow(row interface{}) (*InsertQuery, error) {
 	}, nil
 }
 
+// MakeBatchInsertRow builds a new InsertQuery to insert multiple rows
+func (s *Schema) MakeBatchInsertRow(rows []interface{}) (*BatchInsertQuery, error) {
+	if len(rows) == 0 {
```
changpingc, line 570: can we call MakeInsertRow here instead of duplicating logic?

sqlgen/mysql.go
```diff
@@ -128,6 +128,48 @@ func (q *InsertQuery) ToSQL() (string, []interface{}) {
...
+		for i, column := range q.Columns {
+			if i > 0 {
+				buffer.WriteString(", ")
+			}
+			buffer.WriteString(column)
+		}
+		buffer.WriteString(") VALUES ")
+
+		numRows := len(q.Values) / len(q.Columns)
+		for j := 0; j < numRows; j++ {
```
felixweinberger, line 155: nit: for consistency should we also use `i` at this level and `j` in the inner loop?

sqlgen/db.go
changpingc: regardless of implementation there will always be a limit on batch size though. because the limit of 4MB is so small I worry this will be a recurring source of outage.

abhinavbatra: I noticed the db migration API where we limited batch upserts to 250 as we think that's the optimal number for whatever reasons. What do you think about enforcing this hard limit of 250 on num rows. This is not foolproof either but should catch most cases @changpingc

changpingc: imposing a limit sounds great, but i'd like to make that limit very visible to callers due to the risk of outage. i don't want people to call this without knowing the limit and i claim most people don't read comment when the API looks straightforward. 

what do you think about adding a chunk size parameter and apply chunking here? i think it's great because we'd have to add chunking logic to every caller anyway so this reduces code duplication, and makes it strictly safer.

abhinavbatra: makes sense. Addressed

abhinavbatra: @changpingc

sqlgen/db.go
```diff
@@ -361,6 +361,52 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
...
+// Most SQL db enforce a limit on max size of packet, which is why we need to break
+// the rows into chunks.
+//
+// rows should be an array of pointers to a struct, for example:
+//
+//   user1 := &User{Name: "foo"}
+//   user2 := &User{Name: "fan"}
+//   if err := db.InsertRows(ctx, []{interface{}}{user1, user2}, 100); err != nil {
+//
+func (db *DB) InsertRows(ctx context.Context, rows []interface{}, chunkSize int) (sql.Result, error) {
```
changpingc, line 374: @stephen do you have a preference on taking in `[]interface{}` or `interface{}` and walking with reflect (like [this](https://golang.org/pkg/sort/#Slice))? if this takes `[]interface{}`, the caller may have to convert a strongly typed slice to an interface typed slice in a loop, because `[]*User` is not assignable to `[]interface{}`.

sqlgen/db.go
```diff
@@ -361,6 +361,52 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
...
+				return nil, err
+			}
+		}
+		result, err = db.execWithTrace(ctx, query, "InsertRows")
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	tx.Commit()
```
changpingc, line 406: ah interesting. if a caller already has allocated a transaction in context, we shouldn't call Commit as it would commit any dirty write preceding InsertRows. not using transaction is also weird because I as a user would expect this function to write none or all since it's a single function call. 

maybe we check if we are called inside a transaction first and do things differently?

sqlgen/db.go
```diff
@@ -361,6 +361,52 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
...
+			return nil, err
+		}
+
+		numColumns := len(query.Columns)
+		for i := 0; i < len(slice); i++ {
+			if err := db.checkColumnValuesAgainstLimits(ctx, query, query.Columns, query.Values[i*numColumns:(i+1)*numColumns], query.Table); err != nil {
+				return nil, err
+			}
+		}
+		result, err = db.execWithTrace(ctx, query, "InsertRows")
```
changpingc, line 400: I'm not sure what's the implication of calling execWithTrace many times using the same operation name.

changpingc: Also, it might be weird to return partial `result`. Maybe we don't return result at all?

sqlgen/mysql.go
changpingc: sg

sqlgen/reflect.go
changpingc: sg

sqlgen/db.go
abhinavbatra: good catch. changed to only start a transaction if one doesn't already exist

sqlgen/db.go
abhinavbatra: I agree that interface{} would be better. changed

sqlgen/db.go
abhinavbatra: > I'm not sure what's the implication of calling execWithTrace many times using the same operation name.

@changpingc We already do that when someone calls InsertRow in a loop, right ? So nothing different here

abhinavbatra: > Also, it might be weird to return partial result. Maybe we don't return result at all?

Makes sense

sqlgen/db.go
```diff
@@ -361,6 +362,68 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
...
+//
+func (db *DB) InsertRows(ctx context.Context, rows interface{}, chunkSize int) error {
+	val := reflect.ValueOf(rows)
+	kind := val.Kind()
+	if kind != reflect.Slice && kind != reflect.Array {
+		return fmt.Errorf("expect array/slice got %s", val.Kind().String())
+	}
+
+	rowsData := make([]interface{}, val.Len())
+	for i := 0; i < val.Len(); i++ {
```
changpingc, line 383: consider adding early return if val.Len() == 0. otherwise i'm not sure if BatchInsertQuery.ToSQL() returns a correct SQL.

changpingc:
oh just one more thing, will you add to https://github.com/samsarahq/thunder/blob/master/CHANGELOG.md

sqlgen/db.go
abhinavbatra: if val.Len == 0, then rowData will be empty as well and we end up doing nothing

abhinavbatra:
> oh just one more thing, will you add to https://github.com/samsarahq/thunder/blob/master/CHANGELOG.md

@changpingc done

abhinavbatra:
@changpingc Would you be merging this commit ?

sqlgen/db.go
```diff
@@ -361,6 +362,68 @@ func (db *DB) InsertRow(ctx context.Context, row interface{}) (sql.Result, error
...
+		}
+		defer tx.Rollback()
+	}
+
+	for j := 0; j < len(rowsData); j += chunkSize {
+		sliceLength := chunkSize
+		if len(rowsData) < j+chunkSize {
+			sliceLength = len(rowsData) - j
+		}
+		slice := rowsData[j:sliceLength]
```
changpingc, line 402: should this be `rowsData[j:(j+sliceLength)]`? if so, will you add a unit test for this?

nit: maybe this is my personal style, i find the inverted logic here more readable, like this
```
func (db *DB) InsertRows(ctx context.Context, rows interface{}, maxChunkSize int) error {
  ...
  chunkSize := len(rowsData) - j
  if chunkSize > maxChunkSize {
    chunkSize = maxChunkSize
  }
  chunk := rowsData[j:j+chunkSize]
  ...
}
```

abhinavbatra: good catch. The existing unit test already tests this path. I am surprised to see unit tests pass. Let me investigate

abhinavbatra: @changpingc fixed. Also, added a unit test with small chunk size

changpingc: where is this unit test?

sqlgen/db_test.go
```diff
@@ -123,6 +123,49 @@ func TestShardLimitInsert(t *testing.T) {
...
+
+			// Check aliceDb can't insert alice and bob.
+			err = aliceDb.InsertRows(ctx, [](*User){alice, bob}, 100)
+			assert.Contains(t, err.Error(), "db requies name = Alice, but query has name = Bob")
+
+			tom := &User{Name: "Tom"}
+			jerry := &User{Name: "Jerry"}
+			// Check db can insert tom and jerry with a batch size of 1.
+			err = db.InsertRows(ctx, [](*User){tom, jerry}, 1)
+			assert.NoError(t, err)
```
abhinavbatra, line 154: @changpingc here. Weird that github wasn't showing this previously. I forced pushed again

changpingc: can you check that these rows are inserted? this test would have passed without your fix for small batch.

abhinavbatra: @changpingc modified the test

PR #394, opened 2020-07-03
go sql: add option to panic on non-indexed queries

er9781:
In this commit we add the ability to instrument a connection
with a boolean that, when enabled, will have a query panic
when no index is present in a sql explain for that query.

We also add the ability to bypass this check on any given query
in SelectOptions. Additionally we add some thin wrappers to our
query functions to set this option.

This will permit us to enable this check in
tests and have people explicitly make full table scan queries
when they fully intend to.

Implementation detail:
We check both key and possible keys since there are times where key
will be non-null (eg PRIMARY) while possible keys is null, and also
times when key is null, but possible keys is not null (eg when
mysql chooses a full scan despite the presence of an index due to small
table size)

### DIFF
--- a/sqlgen/db.go
+++ b/sqlgen/db.go
@@ -3,9 +3,11 @@ package sqlgen
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 
+	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/batch"
 )
 
@@ -20,6 +22,8 @@ type DB struct {
 	shardLimit Filter
 
 	dynamicLimit DynamicLimit
+
+	panicOnNoIndex bool
 }
 
 type DynamicLimitFilterCallback func(context.Context, string) Filter
@@ -48,8 +52,9 @@ type DynamicLimit struct {
 
 func NewDB(conn *sql.DB, schema *Schema) *DB {
 	db := &DB{
-		Conn:   conn,
-		Schema: schema,
+		Conn:           conn,
+		Schema:         schema,
+		panicOnNoIndex: false,
 	}
 
 	db.batchFetch = &batch.Func{
@@ -132,6 +137,19 @@ func (db *DB) WithShardLimit(shardLimit Filter) (*DB, error) {
 	dbCopy := *db
 	dbCopy.shardLimit = shardLimit
 	return &dbCopy, nil
+}
+
+// WithPanicOnNoIndex will configure this db connection to run an
+// explain on every query and panic when no index is found. This setting is
+// recommended only for use in testing so that you can find non-indexed
+// queries during tests.
+func (db *DB) WithPanicOnNoIndex() (*DB, error) {
+	if db.panicOnNoIndex {
+		return nil, errors.New("already is set panic on no index")
+	}
+
+	db.panicOnNoIndex = true
+	return db, nil
 }
 
 // WithDynamicLimit is similar to WithShardLimit except that it supports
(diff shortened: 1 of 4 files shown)


sqlgen/db.go
```diff
@@ -48,8 +51,9 @@ type DynamicLimit struct {
 
 func NewDB(conn *sql.DB, schema *Schema) *DB {
 	db := &DB{
-		Conn:   conn,
-		Schema: schema,
+		Conn:           conn,
+		Schema:         schema,
+		panicOnNoIndex: false,
```
er9781, line 56: could let the `false` be implicit, but I sort of prefer being explicit on this type of stuff.

sqlgen/db_test.go
```diff
@@ -9,6 +9,45 @@ import (
...
+	}()
+
+	tdb, db, err := setup()
+	assert.NoError(t, err)
+	defer tdb.Close()
+
+	db, err = db.WithPanicOnNoIndex()
+
+	// A second set will error
+	_, err = db.WithPanicOnNoIndex()
```
EmmaOnThursday, line 26: ooc why do we want to panic if we set this a second time? why not just fail silently since there is no change?

er9781: It errors, not panics, and I tend to agree with you on silently running, but other functions similar to this in the file error out. It does make sense to some extent since it indicates an odd db setup from the caller.

sqlgen/db.go
```diff
@@ -252,6 +292,36 @@ func (db *DB) BaseQuery(ctx context.Context, query *BaseSelectQuery) ([]interfac
 
 	clause, args := selectQuery.ToSQL()
 
+	if db.panicOnNoIndex && (query.Options == nil || !query.Options.AllowNoIndex) {
+		// We run an explain first and panic if there's no index
+		explained := "EXPLAIN " + clause
+		res, err := db.QueryExecer(ctx).QueryContext(ctx, explained, args...)
+		if err != nil {
+			return nil, err
```
CodeBrew28, line 300: oops.wrap error

sqlgen/db.go
```diff
@@ -232,6 +245,33 @@ func (db *DB) checkColumnValuesAgainstLimits(ctx context.Context, query SQLQuery
...
+}
+
+func parseExplainResults(resultRows *sql.Rows) (results []ExplainResultRow, err error) {
+	for resultRows.Next() {
+		var explainType ExplainResultRow
+		err := resultRows.Scan(
+			&explainType.Id, &explainType.SelectType, &explainType.Table, &explainType.TypeColumn, &explainType.PossibleKeys, &explainType.Key,
+			&explainType.KeyLen, &explainType.Ref, &explainType.Rows, &explainType.Extra)
+		if err != nil {
+			return nil, err
```
CodeBrew28, line 268: opps.wrap error

sqlgen/db_test.go
```diff
@@ -9,6 +9,45 @@ import (
...
+	var users []*User
+	db.Query(context.Background(), &users, nil, nil)
+}
+
+func TestPanicOnNoIndexOverride(t *testing.T) {
+	tdb, db, err := setup()
+	assert.NoError(t, err)
+	defer tdb.Close()
+
+	db, err = db.WithPanicOnNoIndex()
```
CodeBrew28, line 39: Is this a safe change to merge into production. Can we safely know we haven't missed a query that doesn't use the index?

sqlgen/db.go
```diff
@@ -252,6 +292,36 @@ func (db *DB) BaseQuery(ctx context.Context, query *BaseSelectQuery) ([]interfac
...
+		res, err := db.QueryExecer(ctx).QueryContext(ctx, explained, args...)
+		if err != nil {
+			return nil, err
+		}
+
+		explainRes, err := parseExplainResults(res)
+		res.Close()
+
+		// If our explain parsing fails, we don't analyze the response or fail the query. We know
+		// this works for our happy paths and nested queries are currently out of scope.
```
CodeBrew28, line 307: Could we ever support nested querries?

EmmaOnThursday: @er9781 I don't think nested queries are an issue for parsing, actually - this turned out to be a red herring when we were testing.

sqlgen/db.go
```diff
@@ -134,6 +138,15 @@ func (db *DB) WithShardLimit(shardLimit Filter) (*DB, error) {
 	return &dbCopy, nil
 }
 
+func (db *DB) WithPanicOnNoIndex() (*DB, error) {
+	if db.panicOnNoIndex {
+		return nil, errors.New("already is set panic on no index")
```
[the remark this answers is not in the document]
KevinXing: We only want to use this in test right? Can we add a comment to call it out explicitly? I am worried that people use it in production and then cause outages.

sqlgen/db_test.go
er9781: This is not currently on in production for us. There will be followup PRs where we:
1. only turn this on in test envs (so yes it's safe)
2. in that PR, we will manually tag all full scan queries as such.

sqlgen/db.go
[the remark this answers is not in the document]
er9781: I previously had this called `IsTest` but thought to name it according to what it does. I'd happily add a comment as to what behavior this controls though!!

sqlgen/db.go
```diff
@@ -252,6 +292,36 @@ func (db *DB) BaseQuery(ctx context.Context, query *BaseSelectQuery) ([]interfac
...
+	if db.panicOnNoIndex && (query.Options == nil || !query.Options.AllowNoIndex) {
+		// We run an explain first and panic if there's no index
+		explained := "EXPLAIN " + clause
+		res, err := db.QueryExecer(ctx).QueryContext(ctx, explained, args...)
+		if err != nil {
+			return nil, err
+		}
+
+		explainRes, err := parseExplainResults(res)
+		res.Close()
```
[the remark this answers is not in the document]
er9781: We can't use defer res.Close() here (I had it that way initially). We need to close  this query before executing the main query after the explain check

sqlgen/db.go
```diff
@@ -252,6 +292,36 @@ func (db *DB) BaseQuery(ctx context.Context, query *BaseSelectQuery) ([]interfac
 
 	clause, args := selectQuery.ToSQL()
 
+	if db.panicOnNoIndex && (query.Options == nil || !query.Options.AllowNoIndex) {
```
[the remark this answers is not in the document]
er9781: sure!

sqlgen/db.go
er9781: Yeah, we  should get multiple rows in explain and we check all of them 👍

sqlgen/db.go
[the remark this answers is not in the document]
er9781: See above comment from emma re noop. I'm fine with a noop, but other similar functions in this file return an error, so I matched behavior.

sqlgen/db.go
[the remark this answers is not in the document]
er9781: I can do this now that  it's a helper!! woot

sqlgen/db.go
```diff
@@ -232,6 +250,64 @@ func (db *DB) checkColumnValuesAgainstLimits(ctx context.Context, query SQLQuery
...
+	res, err := db.QueryExecer(ctx).QueryContext(ctx, explained, args...)
+	if err != nil {
+		return oops.Wrapf(err, "Failed to run explain on the query")
+	}
+	defer res.Close()
+
+	explainRes, err := parseExplainResults(res)
+
+	// If our explain parsing fails, we don't analyze the response or fail the query.
+	if err == nil {
```
KevinXing, line 292: return if err != nil, so we can save an intent?

er9781: what do you mean "save an intent"?

KevinXing: sorry, i mean save a tab... 
Also, we should return err is err != nil

er9781: Discussed offline, we're going to fail the query if explain parsing fails.

PR #395, opened 2020-07-06
Adding New Federation Keys

CodeBrew28:
This mocks the test where we add a new federated key, and the federated executor knows about it, but the gateway doesn't. As long as the field is marked optional, it will be autofilled as an empty argument until it has the correct data. 

Similarly, if you make your federated object identical to the original object, and you add a new field, if the new field is marked optional, it will only send that field over once both servers know about it. If the root server knows about it first, the federated server won't ask for the field until it knows about it. If the federated server knows about it first, it will be autofilled as a nil value until the federated server knows about it.


federation/schema_syncer_test.go
```diff
@@ -243,3 +243,113 @@ func TestExecutorQueriesWithCustomSchemaSyncer(t *testing.T) {
...
+
+	// Run a federated query and ensure that it works
+	runAndValidateQueryResults(t, ctx, e, query, expectedOutput)
+
+	type UserWithContactInfoNew struct {
+		Id    int64
+		OrgId int64
+		Name  string
+		// When the gateway knows about email, it will fill it correctly
+		Email string `graphql:",optional"`
```
KevinXing, line 329: huh, could you point me the code where we deal with the optional field?

KevinXing:
Also, we should force people to make new field as optional.

federation/schema_syncer_test.go
CodeBrew28: Yup! https://github.com/samsarahq/thunder/blob/master/graphql/schemabuilder/input.go#L160

federation/schema_syncer_test.go
```diff
@@ -243,3 +243,113 @@ func TestExecutorQueriesWithCustomSchemaSyncer(t *testing.T) {
...
+	})
+
+	type UserWithContactInfo struct {
+		Id    int64
+		OrgId int64
+		Name  string
+	}
+
+	s2 := schemabuilder.NewSchemaWithName("schema2")
+	s2.FederatedFieldFunc("User", func(args struct{ Keys []*UserWithContactInfo }) []*UserWithContactInfo {
```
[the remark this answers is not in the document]
CodeBrew28: Yup!

CodeBrew28:
Yes we should force them, I think that logic can be built into the gateway if it can compare objects between the new and old schema? We should look into adding that enforcement

PR #398, opened 2020-07-08
Shadow FieldFunc

CodeBrew28:
In federation, we want to support shadow field funcs. This autogenerates a field to construct the shadow object from the args such that we can run a subquery. We had a bug in which it wasn't correctly reading the args, which gets fixed and tested here.


graphql/schemabuilder/function.go
```diff
@@ -171,8 +171,9 @@ func (sb *schemaBuilder) buildShadowObjectFederationFunction(typ reflect.Type, m
 	rType := &graphql.NonNull{Type: &graphql.List{Type: returnType}}
 	field := &graphql.Field{
 		Resolve: func(ctx context.Context, source, funcRawArgs interface{}, selectionSet *graphql.SelectionSet) (interface{}, error) {
-			args := funcRawArgs.(map[string]graphql.Type)
-			return args["keys"], nil
+			args := reflect.ValueOf(funcRawArgs)
+			keys := reflect.Indirect(args).FieldByName("Keys")
```
KevinXing, line 175: aha, nice catch!
Sorry I missed this when i reviewed the code.

PR #400, opened 2020-07-09
Federation Shadow Type Validation

CodeBrew28:
Enforces that a field can not return an object if it is a shadow type. This ensures that when a field func returns an object, we can query any field on that object and the client doesn't have to be aware of which ones are on federated servers. This also helps enforce that we can only federated field funcs, and that all queries for fields on the struct are responded to by the server with the root object.

### DIFF
--- a/federation/schema.go
+++ b/federation/schema.go
@@ -76,6 +76,54 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
 	return nil
 }
 
+// validateFieldsReturningFederatedObject checks that if an object is returned by a field func it can not be a shadow object type
+func validateFieldsReturningFederatedObject(serviceNames []string, serviceSchemasByName map[string]*IntrospectionQueryResult, types map[string]graphql.Type, fieldInfos map[*graphql.Field]*FieldInfo) error {
+	for service, serviceSchema := range serviceSchemasByName {
+		for _, typ := range serviceSchema.Schema.Types {
+			if typ.Kind != "OBJECT" {
+				continue
+			}
+			for _, field := range typ.Fields {
+				// Check that the field's return type is an object
+				fieldReturnType := getRootType(field.Type)
+				if fieldReturnType.Kind != "OBJECT" {
+					continue
+				}
+				// Error if it is a shadow object. To check this
+				// (1) Look through all the fields on the object to see if there is a federation field (__federation) and that
+				// the federation field is not on the current service
+				// (2) Look through all the fields on the federation object to see if it has a field for <ObjectType>-<Service>
+				returnObj, ok := types[fieldReturnType.Name].(*graphql.Object)
+				if !ok {
+					return oops.Errorf("Return type %s is not a graphql object", fieldReturnType.Name)
+				}
+				for name, f := range returnObj.Fields {
+					if name == federationField && !fieldInfos[f].Services[service] {
+						federatedFieldName := fmt.Sprintf("%s-%s", fieldReturnType, service)
+						// If the field name is <fieldType-service> on a federation object, 
+						// it is an expected function for a shadow object type
+						if field.Name == federatedFieldName {
+							continue
+						}
+						fedObj, ok := types["Federation"].(*graphql.Object)
+						// If there isn't a federation object, it isn't a shadow object
+						if !ok {
+							continue
+						}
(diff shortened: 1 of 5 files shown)


federation/schema.go
```diff
@@ -76,6 +76,43 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
...
+			if typ.Kind == "OBJECT" {
+				for _, field := range typ.Fields {
+					// Check that the field's return type is an object
+					fieldReturnType := getRootType(field.Type)
+					if fieldReturnType.Kind == "OBJECT" {
+						// Error if it is a shadow object. To check this
+						// (1) Look through all the fields on the object to see if there is a federation field (__federation) and that
+						// the federation field is not on the current service
+						// (2) Look through all the fields on the federation object to see if it has a field for <ObjectType>-<Service>
+						returnObj, _ := types[fieldReturnType.Name].(*graphql.Object)
```
KevinXing, line 92: we can remove the  `_` if we are sure the `returnObj` must be a `graphql.Object`.

federation/schema.go
```diff
@@ -76,6 +76,43 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
...
+						for name, f := range returnObj.Fields {
+							if name == federationField && !fieldInfos[f].Services[service] {
+								federatedFieldName := fmt.Sprintf("%s-%s", fieldReturnType, service)
+								if field.Name != federatedFieldName {
+									fedObj := types["Federation"].(*graphql.Object)
```
KevinXing, line 97: I think it is also possible that we don't have a "Federation" type?

federation/schema.go
```diff
@@ -76,6 +76,43 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
 	return nil
 }
 
+// validateFieldsReturningFederatedObject checks that if an object is returned by a field func it can not be a shadow object type
+func validateFieldsReturningFederatedObject(serviceNames []string, serviceSchemasByName map[string]*IntrospectionQueryResult, types map[string]graphql.Type, fieldInfos map[*graphql.Field]*FieldInfo) error {
+	for _, service := range serviceNames {
```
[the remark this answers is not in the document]
CodeBrew28: I think we still want to iterate over all the objects per service. If we remove that for loop, we might miss objects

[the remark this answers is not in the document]
KevinXing: Ah, I meant we can iterate `serviceSchemasByName` instead of `serviceName`? I am also okay with what we have now.

federation/schema.go
```diff
@@ -76,6 +76,52 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
...
+				// the federation field is not on the current service
+				// (2) Look through all the fields on the federation object to see if it has a field for <ObjectType>-<Service>
+				returnObj, ok := types[fieldReturnType.Name].(*graphql.Object)
+				if !ok {
+					return oops.Errorf("Return type %s is not a graphql object", fieldReturnType.Name)
+				}
+				for name, f := range returnObj.Fields {
+					if name == federationField && !fieldInfos[f].Services[service] {
+						federatedFieldName := fmt.Sprintf("%s-%s", fieldReturnType, service)
+						if field.Name != federatedFieldName {
```
KevinXing, line 103: Why need this check?

CodeBrew28: On the "Federation" object, we have a function that looks like  "objectName-serviceName" (eg User-s2). This is used to create the shadow object, but in this case it's allowed to return the "user" object type. We want to ignore this check then

federation/schema.go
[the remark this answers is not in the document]
CodeBrew28: Oh I see, I think I slightly prefer to keep it like this to keep it consistent with other places in the planner where we do a similar loop if that's all right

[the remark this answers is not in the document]
CodeBrew28: Ah, I actually like the way it simplifies the logic! Updated!

federation/schema.go
KevinXing: Maybe add this as a comment of this check and change it to 
`if field.Name == federatedFieldName {continue}` ?

CodeBrew28: updated!

PR #401, opened 2020-07-09
Federation: Testing adding and deleting keys

CodeBrew28:
Tests adding and deleting a federated key, and all the states in between that process. This included places where the server has an updated version, but the gateway hasn't updated yet. We do have to make sure that when deleting a field, all the shadow objects stop asking for it. Similarly, we have to make sure that when adding a field, the root object knows about it before the shadow objects know about it.

### DIFF
--- a/federation/schema_syncer_test.go
+++ b/federation/schema_syncer_test.go
@@ -324,10 +324,6 @@ func TestOnlyShadowServiceKnowsAboutNewField(t *testing.T) {
 		Id    int64
 		OrgId int64
 		Name  string
-		// When the gateway knows about email, it will fill it correctly
-		Email string `graphql:",optional"`
-		// This will never get filled correctly since the gateway doesnt know about this field
-		UnknownField string `graphql:",optional"`
 	}
 
 	s2new := schemabuilder.NewSchemaWithName("schema2")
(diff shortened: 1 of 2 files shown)


federation/schema_syncer_test.go
```diff
@@ -351,3 +347,248 @@ func TestOnlyShadowServiceKnowsAboutNewField(t *testing.T) {
...
+	s1.Query().FieldFunc("users", func(ctx context.Context) ([]*User, error) {
+		users := make([]*User, 0, 1)
+		users = append(users, &User{Id: int64(1), OrgId: int64(1)})
+		users = append(users, &User{Id: int64(2), OrgId: int64(2)})
+		return users, nil
+	})
+
+	type UserWithContactInfo struct {
+		Id    int64
+		OrgId int64
```
KevinXing, line 368: If we change the OrgId to `optional` before removing it, can we avoid making the change in the previous commit?

CodeBrew28: This is the case where we send it the org id, but that's not an arg its expected. The optional tag works the other way where we don't send it the org id, but its expecting it.

KevinXing: Ah i see, thanks for the explanation!

graphql/schemabuilder/input.go
```diff
@@ -98,11 +98,6 @@ func (sb *schemaBuilder) makeStructParser(typ reflect.Type) (*argParser, graphql
 					return fmt.Errorf("%s: %s", name, err)
 				}
 			}
-			for name := range asMap {
```
KevinXing, line 101: Do you also add a comment to say why it is safe to remove this check?

CodeBrew28: added to the commit message!

KevinXing:
Nice!

PR #405, opened 2020-07-20
batch: fix panic when context missing WithBatching

ehiggins98:
Previously line 130 panicked with the error "interface{} is nil, not *batch.batchContext" if `Invoke` was called without calling `WithBatching` on `ctx`, so the intended error message was not printed. This change ensures the correct panic message is printed, for easier debugging.

### DIFF
--- a/batch/batch.go
+++ b/batch/batch.go
@@ -127,8 +127,8 @@ func safeInvoke(
 // Invoke arranges for the Func's Many to be called with arg as one of its
 // arguments, and returns the corresponding result.
 func (f *Func) Invoke(ctx context.Context, arg interface{}) (interface{}, error) {
-	bctx := ctx.Value(batchContextKey{}).(*batchContext)
-	if bctx == nil {
+	bctx, ok := ctx.Value(batchContextKey{}).(*batchContext)
+	if !ok {
 		panic("WithBatching must be called on the context before using Func")
 	}
 
--- a/batch/batch_test.go
+++ b/batch/batch_test.go
@@ -9,6 +9,8 @@ import (
 	"time"
 
 	"github.com/samsarahq/thunder/batch"
+	"github.com/stretchr/testify/assert"
+
 )
 
 // TestBasic tests that batch.Func with default options batches calls.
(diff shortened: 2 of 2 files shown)


batch/batch.go
```diff
@@ -127,8 +127,8 @@ func safeInvoke(
 // Invoke arranges for the Func's Many to be called with arg as one of its
 // arguments, and returns the corresponding result.
 func (f *Func) Invoke(ctx context.Context, arg interface{}) (interface{}, error) {
-	bctx := ctx.Value(batchContextKey{}).(*batchContext)
-	if bctx == nil {
+	bctx, ok := ctx.Value(batchContextKey{}).(*batchContext)
+	if !ok {
 		panic("WithBatching must be called on the context before using Func")
```
stephen, line 132: consider adding a unit test

ehiggins98: @stephen added. I'm not a member of this repo, so could you merge the PR?

batch/batch_test.go
```diff
@@ -300,3 +300,28 @@ func TestError(t *testing.T) {
...
+	for i := 0; i < 20; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			defer func() {
+				if r := recover(); r != "WithBatching must be called on the context before using Func" {
+					t.Errorf("Incorrect panic message %s for %d", r, i)
+				}
+			}()
+			f(ctx, i)
```
stephen, line 323: consider `assert.PanicsWithError` instead. if this didn't panic at all (i.e. without your fix), i think your test would still pass.

batch/batch_test.go
```diff
@@ -300,3 +300,28 @@ func TestError(t *testing.T) {
...
+	f := (&batch.Func{
+		Many: func(ctx context.Context, args []interface{}) ([]interface{}, error) {
+			return nil, nil
+		},
+	}).Invoke
+
+	ctx := context.Background()
+
+	var wg sync.WaitGroup
+	for i := 0; i < 20; i++ {
```
stephen, line 314: i think you don't need this for loop wait group contraption, since the panic we're testing is immediate when f is invoked

batch/batch_test.go
ehiggins98: Ohh that's a much cleaner way to write it. Fixed

batch/batch_test.go
ehiggins98: 👍  removed it

PR #410, opened 2020-07-30
Federation API 

CodeBrew28:
* Removes shadow and root objects. We moved the generation of these functions elsewhere. The motivation for this is that we want the objects on two different servers to always be the same to allow us to hop from any server to another server. Either the keys or entire object can be sent over grpc. 
* Creates an option called `FetchObjectFromKeys` that generates the "__federation" field on an object, and takes in a function as an arg that use used for the shadow function to fetch the object from the keys
* Adds validation that if an object is federated on one server, it is federated on all the other servers with that object

https://paper.dropbox.com/doc/Federation-API-Updated--A4uTEGk52asO9hNzGXB3ZiztAg-f3mzf2dC0zjl60U7ReHAj

### DIFF
--- a/federation/schema.go
+++ b/federation/schema.go
@@ -43,9 +43,9 @@ func getRootType(typ *introspectionTypeRef) *introspectionTypeRef {
 // that have the objcet registered as a root object expose the field. This ensures that we can make
 // the hop from the root server to any of the federated servers safely before any queries are executed.
 func validateFederationKeys(serviceNames []string, serviceSchemasByName map[string]*IntrospectionQueryResult, obj *graphql.Object, keyField string) error {
-	validFederatedKey := false
 	for _, service := range serviceNames {
 		for _, typ := range serviceSchemasByName[service].Schema.Types {
+			validFederatedKey := false
 			if typ.Name == obj.Name {
 				// Check that it is a root object by checking if it has a field func called
 				// "__federation" on the object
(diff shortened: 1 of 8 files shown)


federation/schema.go
```diff
@@ -177,6 +219,9 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
 	fieldInfos := make(map[*graphql.Field]*FieldInfo)
 	for _, service := range serviceNames {
 		for _, typ := range serviceSchemasByName[service].Schema.Types {
+			if err := validateFederatedObjects(serviceNames, serviceSchemasByName, typ.Name); err != nil {
```
KevinXing, line 222: I think this call includes some redundant computation since we can just iterate each type once?

CodeBrew28: Im not sure I follow what you mean. I think for every type on every service, we need to check if that type is federated with both the "__federation" field and the "object-service" field

graphql/schemabuilder/schema.go
```diff
@@ -117,23 +117,13 @@ type objectOptionFunc func(*Schema, *Object)
...
-}
-
-var ShadowObject objectOptionFunc = func(s *Schema, m *Object) {
-	m.IsShadow = true
-}
-
 type federation struct{}
 
-func CustomShadowObject(f interface{}, options ...ObjectOption) ObjectOption {
+func FetchObjectFromKeys(f interface{}, options ...ObjectOption) ObjectOption {
```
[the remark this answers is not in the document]
CodeBrew28: I think CreateObject implies that it is a few objects being created, vs in this case we are fetching an existing object. I think `FromKeys` makes it more explicit of how the keys are used to determine the object,s but happy to make it `FetchObject` if that sounds good to you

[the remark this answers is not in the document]
KevinXing: Okay, I think I am good with what we have.

federation/schema.go
```diff
@@ -56,6 +56,7 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
 						break
 					}
 				}
+
```
amyliu345, line 59: nit: extra newline here and 94, 118 and 228

federation/schema.go
```diff
@@ -66,13 +67,55 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
...
+			for _, typ := range serviceSchemasByName[service].Schema.Types {
+				if typ.Name == objName {
+					isRootObject := false
+					for _, introspectedField := range typ.Fields {
+						if introspectedField.Name == federationField {
+							isRootObject = true
+							break
+						}
+					}
+					if !isRootObject && federatedOnOneService {
```
amyliu345, line 111: isn't federatedOnOneService here guaranteed to be true? since line 98 checks for it already?

federation/schema.go
```diff
@@ -66,13 +67,55 @@ func validateFederationKeys(serviceNames []string, serviceSchemasByName map[stri
...
-	if !validFederatedKey {
-		return oops.Errorf("Invalid federation key %s", keyField)
+
+	if federatedOnOneService && objName != "Query" && objName != "Mutation" {
+		// If the object is federated on one service, make sure every service with
+		// that object can parse the keys and fetch the object
+		for _, service := range serviceNames {
+			for _, typ := range serviceSchemasByName[service].Schema.Types {
+				if typ.Name == objName {
+					isRootObject := false
```
amyliu345, line 104: is my understanding correct that the api change makes it so that root object doesn't really mean anything anymore; aka all federated objects are the same now with no root/shadow distinction? in that case should we call this `isFederatedObject` so it's less confusing?

federation/schema.go
CodeBrew28: ah nice catch! fixed!

federation/schema.go
CodeBrew28: yup, renamed!

federation/schema.go
KevinXing: I can still see it?

CodeBrew28: Im not sure what you mean, it seems to be fixed 
image

KevinXing: I am still seeing `if !isRootObject && federatedOnOneService {`, this is the commit I am looking at. https://github.com/samsarahq/thunder/pull/410/commits/015f7a5dd0009eb68f4b6d9680d073990219b76f

CodeBrew28: It got merged in the commit after, updated!

federation/schema.go
```diff
@@ -173,6 +214,11 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
 	if err != nil {
 		return nil, err
 	}
+	for name, _ := range types {
```
KevinXing, line 217: nit: we can get rid of `_`.

PR #353, opened 2020-01-31
Directive support for skip and include

CodeBrew28:
Directive support for skip and include as following https://graphql.org/learn/queries/#directives

PR #412, opened 2020-08-07
Amyliu/federation directives

amyliu345:
This PR adds directive support (`@skip` and `@include`) for federated queries 

[Jira ticket](https://samsaradev.atlassian.net/jira/software/projects/CLOUD/boards/437?selectedIssue=CLOUD-283)

Because the ShouldIncludeNode fxn is called in the planning/flattening step, I found that I didn't have to marshal Directives into thunderpb since at that point the directives had already been applied. I actually made the proto changes and added it to the marshaling/unmarshaling of the DirectExecutorClient queries, but realized that everything works without needing to do that. Lmk if this seems wrong though.

References:
- original directive support [PR](https://github.com/samsarahq/thunder/pull/353)
- directive support bugfix [PR](https://github.com/samsarahq/thunder/pull/361)

### DIFF
--- a/federation/normalize.go
+++ b/federation/normalize.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 	"sort"
 
+	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/graphql"
 )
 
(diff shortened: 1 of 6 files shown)


federation/normalize.go
```diff
@@ -115,7 +115,14 @@ func (f *flattener) flattenFragments(selectionSet *graphql.SelectionSet, typ *gr
 
 	// Descend into fragments matching the current type.
 	for _, fragment := range selectionSet.Fragments {
-		ok, err := f.applies(typ, fragment)
+		ok, err := graphql.ShouldIncludeNode(fragment.Directives)
+		if err != nil {
+			return err
```
CodeBrew28, line 120: Can we oop.wrap the error here and in planner.go

CodeBrew28:
Ah I love this approach, it's a lot simpler than marshaling the directives over grpc!

federation/normalize.go
amyliu345: fixed!

PR #419, opened 2020-10-02
Federation Name Convention

CodeBrew28:
For apollo ts convention, field funcs need to start with a lowercase letter. We can't force the object name to be lowercase, but we can force the service name, so I switched the ordering.


amyliu345 (approved):
lgtm!

PR #427, opened 2020-11-24
Removing Repeated Selections and Fragments

CodeBrew28:
We had an issue where when we turned on federation when it was planning the query we saw high spikes on memory on the function where it merges queries with the same alias. The query was always successfully removing all the repeats, but the bug was making it run through those lines more often then it needs to 

In this example, this is the input
```
 {
    users {
        ... userFragment1
       ... userFragment2
     }
}

fragment userFragment1 {
    name
}
fragment userFragment2 {
    name
    phoneNumber
    ... userFragment3
}
fragment userFragment3 {
    name
    phoneNumber
}
```
and this is the expected output. 
```
 {
    users {
        name
        phone number
     }
}
```

When we flatten all the fragments, we check if they are all on the User type, and if they are we added it to the list of selections. At this point, its safe to remove it as a fragment since we have already flattened it, which is a step we weren't doing before. 

The other bug was when we were merging aliases, if we saw a repeated selection or fragment, we were adding it to the list again. Now we use a dictionary to check we don't add the same selection or fragment more than once, which helps reduce the number of times we need to flatten and merge aliases.

### DIFF
--- a/federation/normalize.go
+++ b/federation/normalize.go
@@ -110,7 +110,7 @@ func (f *flattener) applies(obj *graphql.Object, fragment *graphql.Fragment) (bo
 // flattenFragments flattens all fragments at the current level. It inlines the
 // selections of each fragment, but does not descend down recursively into those
 // selections.
-func (f *flattener) flattenFragments(selectionSet *graphql.SelectionSet, typ *graphql.Object, target *[]*graphql.Selection) error {
+func (f *flattener) flattenFragments(selectionSet *graphql.SelectionSet, typ *graphql.Object, target *[]*graphql.Selection, targetFragments *[]*graphql.Fragment) error {
 	// Start with the non-fragment selections.
 	*target = append(*target, selectionSet.Selections...)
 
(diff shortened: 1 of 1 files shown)


federation/normalize.go
```diff
@@ -110,9 +110,10 @@ func (f *flattener) applies(obj *graphql.Object, fragment *graphql.Fragment) (bo
 // flattenFragments flattens all fragments at the current level. It inlines the
 // selections of each fragment, but does not descend down recursively into those
 // selections.
-func (f *flattener) flattenFragments(selectionSet *graphql.SelectionSet, typ *graphql.Object, target *[]*graphql.Selection) error {
+func (f *flattener) flattenFragments(selectionSet *graphql.SelectionSet, typ *graphql.Object, target *[]*graphql.Selection, targetFragments *[]*graphql.Fragment) error {
 	// Start with the non-fragment selections.
 	*target = append(*target, selectionSet.Selections...)
+	// mergeSameAlias(*target)
```
KevinXing, line 116: remove this comment?

CodeBrew28: updated!

KevinXing (approved):

amyliu345 (approved):
nicee lgtm!

PR #429, opened 2020-11-24
Lowercase sort of pagination

CodeBrew28:
We want to allow pagination to compare strings after lowercasing them since it makes the sort more intuitive in some cases. There is now a custom sort order called `asc_lower` and `desc_lower` to support this without changing the default sorting.

### DIFF
--- a/graphql/schemabuilder/pagination.go
+++ b/graphql/schemabuilder/pagination.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"sort"
 	"sync"
+	"strings"
 
 	"github.com/samsarahq/thunder/batch"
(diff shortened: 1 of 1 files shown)


ehzhang (commented):
why not make the backwards incompatible change and always sort using lowercase?

there is no customer use case to sort without using lowercase, and to backfill all of the sort orders in application code is going to be time-consuming and non-obvious

(why include an option that _allows_ developers to shoot themselves in the foot?)

CodeBrew28:
Ah I see, in that case I think it makes sense just to make the default desc and asc sort orders lowercase before sorting

KevinXing (approved):

ehzhang (approved):

PR #432, opened 2020-12-09
Federation Bug Missing Selections

CodeBrew28:
Fixed a federation bug where for nested fragments we weren't setting the selection set to equal the flattened selections. I also added a test that would catch this scenario


amyliu345 (approved):
LGTM!

KevinXing (approved):
nice catch.

PR #436, opened 2020-12-15
graphql/schemabuilder: rename to ShouldUseBatchFunc

CodeBrew28:
We use to call it ShouldUseFallbackFunc, but if its true we use the batch func and if its false we use the
fallback func. This is a bit misleading, but since its an internlly used structure, we can rename it without
causing any regressions


KevinXing (approved):
Yay!

PR #437, opened 2020-12-15
We would like to use the external executor in mutation as well

KevinXing:


KevinXing:
We have verified this in staging env.

shaneschulte (approved):

PR #440, opened 2020-12-21
Function to get parent name for selection

CodeBrew28:
Function to get the parent name of the selection, which is useful for debugging and metrics


amyliu345 (approved):
lgtm!

PR #441, opened 2020-12-21
Federation Unmarshalling Errors

CodeBrew28:
There's a bug where unmarshaling graphql response strings into
interfaces was turning MaxInt64's into 9223372036854776000 which
overflows int64. We discovered this because we were getting errors
when trying to unmarshal json into a go struct int64 field.

This happens because json.Unmarshal treats numbers as float64s, and
to preserve the int64 value we have to use a json decoder and UseNumber
instead.

See go playground snippet for an example of this behavior:
https://play.golang.org/p/gL8f4BsJ0M-

See stackoverflow for fix:
https://stackoverflow.com/questions/16946306/preserve-int64-values-when-parsing-json-in-go

### DIFF
--- a/federation/customexecutor_test.go
+++ b/federation/customexecutor_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"testing"
+	"bytes"
 
 	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/graphql"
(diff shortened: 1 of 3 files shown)


amyliu345 (approved):
Fix lgtm! With the previous overflow fix the decoder codepath is only turned on for a couple of queries; do we have concerns about this causing issues if we start executing all queries w this decoder logic once we migrate over to federation? We can't FF in thunder right?

CodeBrew28:
Yeah so we can't feature flag it bc of thunder. Since its only serving internal/replayed traffic ar the moment I think we can turn it off if we run into issues. I also suspect it'll be easier to think about if its all using the decoder logic

PR #443, opened 2021-01-20
graphql inspection: add type_as_optional directive

shaneschulte:
this is a no-op directive that will be used on the client-side by our type generation system

See this PR for a deeper explanation:
https://github.com/samsara-dev/backend/pull/93693

Screenshot if it working in GraphiQL:
image

### DIFF
--- a/graphql/introspection/introspection.go
+++ b/graphql/introspection/introspection.go
@@ -123,6 +123,15 @@ var skipDirective = Directive{
 	},
 }
 
+var typeAsOptionalDirective = Directive{
+	Description: "Client-side-only directive that instructs the type generator to mark this field as optional. This is useful for making the generated types compliant with Troy persistence schema.",
+	Locations: []DirectiveLocation{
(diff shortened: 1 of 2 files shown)


graphql/introspection/introspection.go
```diff
@@ -123,6 +123,15 @@ var skipDirective = Directive{
 	},
 }
 
+var typeAsOptionalDirective = Directive{
+	Description: "Client-side-only directive that instructs the type generator to mark this field as optional. This is useful for making the generated types compliant with Troy persistence schema.",
+	Locations: []DirectiveLocation{
+		FIELD,
```
CodeBrew28, line 129: Just double-checking, this means we can't apply this directive to an entire fragment

shaneschulte: yep! it doesn't really make sense in that scenario

CodeBrew28 (approved):
lgtm!

PR #446, opened 2021-01-28
Go modules

eljoth:
Go modules is the future of dependency management in go. It makes developing outside the GOPATH easier.


amonks:
Thanks for the patch! I landed a similar change in #450.

PR #447, opened 2021-02-10
sqlgen: SimpleWhere IS NULL filter support

wic006:
This change updates the SimpleWhere ToSQL functionality to allow handling of IS NULL filtering. To match NULL columns, explicitly pass in the column name with nil value into the sqlgen.Filter.

### DIFF
--- a/sqlgen/mysql.go
+++ b/sqlgen/mysql.go
@@ -21,7 +21,11 @@ func (w *SimpleWhere) ToSQL() (string, []interface{}) {
 				buffer.WriteString(" AND ")
 			}
 			buffer.WriteString(column)
-			buffer.WriteString(" = ?")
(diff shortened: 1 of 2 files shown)


sqlgen/mysql.go
```diff
@@ -21,7 +21,12 @@ func (w *SimpleWhere) ToSQL() (string, []interface{}) {
 				buffer.WriteString(" AND ")
 			}
 			buffer.WriteString(column)
-			buffer.WriteString(" = ?")
+			val := w.Values[i]
+			if val == nil {
```
KevinXing, line 25: maybe just `if w.Values[i] == nil` to save a temporary variable.

wic006: updated

KevinXing (approved):

PR #450, opened 2021-02-16
test with go1.15, use go modules

amonks:
Some issue creators noted that we only test thunder in go 1.10. Since we use go 1.15
in production, this seems unwise. This commit sets the go version to 1.15. It
also makes a change to use go modules rather than the vendor folder. I don't
think we can make one change without the other, but I know very little about go
modules and I'd love to be mistaken.

XXX: I know our setup is a bit weird, and I'm a bit concerned that we'll have
to make changes to internal repos next time we revendor thunder. Maybe we can
add it as a module? That would be neat. I'm inclined to merge this change, then
figure that out.

**********************************************************************************

commands:
```bash
go mod init     // This signaled an error wrt. an "appengine" thing in vendor.json
                // I made the changes below to vendor.json, then tried again.
go mod vendor
rm -rf vendor
```

vendor.json:
```
"ignore": "appengine appenginevm test",         // I removed appengine and appenginevm.
"package": [
	{                                       // I removed this whole package entry.
		"path": "appengine/cloudsql",
		"revision": ""
	},
```

**********************************************************************************

There are also two commits that fix small build/test errors present when
running with go1.15.


amonks:
@denbeigh2000 

Pinging you as resident go-modules knower, particularly w.r.t the note under XXX in the PR description.

amonks:
@CodeBrew28 

Pinging you as thunder maintainer apparent. I think testing in 1.15 should be uncontroversial, and the test changes should be uncontroversial, but I don't know what to make of the go modules change. See the note under XXX in the PR description.

denbeigh2000:
We shouldn't have to do anything special with Thunder for internal use, we should be able to make use of the package regardless of whether it is module-ised (modularised?) or not. The core code hasn't changed, and currently performs fine with all our vendored dependencies, so I don't think upgrading would be a headache.

---

Removing the strange appengine vendoring LGTM. It seems this dependency was only present for the mysql driver:

```
$ rg 'appengine|cloudsql'
vendor/vendor.json
3:	"ignore": "appengine appenginevm test",
6:			"path": "appengine/cloudsql",

vendor/github.com/davecgh/go-spew/spew/bypass.go
19:// +build !js,!appengine,!safe,!disableunsafe

vendor/github.com/davecgh/go-spew/spew/bypasssafe.go
19:// +build js appengine safe disableunsafe

vendor/github.com/go-sql-driver/mysql/appengine.go
9:// +build appengine
14:	"appengine/cloudsql"
18:	RegisterDial("cloudsql", cloudsql.Dial)

vendor/github.com/go-sql-driver/mysql/README.md
387:user@cloudsql(project-id:instance-name)/dbname
392:user@cloudsql(project-id:regionname:instance-name)/dbname

vendor/github.com/gogo/protobuf/proto/pointer_reflect_gogo.go
29:// +build purego appengine js

vendor/github.com/gogo/protobuf/proto/pointer_reflect.go
32:// +build purego appengine js

vendor/github.com/gogo/protobuf/proto/pointer_unsafe_gogo.go
29:// +build !purego,!appengine,!js

vendor/github.com/gogo/protobuf/proto/pointer_unsafe.go
32:// +build !purego,!appengine,!js
```

This was dependency removed in https://github.com/go-sql-driver/mysql/pull/1007, and was apparently only needed for Go <=1.9, so this should be totally fine to remove.

denbeigh2000 (approved):
I approve this message

PR #453, opened 2021-03-15
batch index: implement MarshalText and UnmarshalText

yolandaz:
currently, json.Marshal on batch.Index throws an error `json: unsupported type: map[batch.Index]bool` when attempting to run. This change fixes that issue by implementing MarshalText and UnmarshalText on batch.Index

### DIFF
--- a/batch/index.go
+++ b/batch/index.go
@@ -1,5 +1,7 @@
 package batch
 
+import "encoding/json"
+
 // Index is a "key" mapper for batch requests.  Its key is not special and
 // should be treated the same as an index in a slice.
 type Index struct {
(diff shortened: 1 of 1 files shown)


batch/index.go
```diff
@@ -13,3 +15,15 @@ func NewIndex(key int) Index {
 	// TODO generate a hash to really mess with people trying to recreate batches?
 	return Index{key: key}
 }
+
+// Create another type for Index so that MarshalText and UnmarshalText
+// don't run into recursion issues.
+type index Index
+
+func (i Index) MarshalText() (text []byte, err error) {
+	return json.Marshal(index(s))
```
steveshi7, line 24: what is `s` referring to here

steveshi7 (approved):

steveshi7 (changes requested):
sorry, accidentally pressed approve

batch/index.go
yolandaz: oops should be i

steveshi7 (approved):
lgtm

PR #455, opened 2021-03-30
graphql/batch_executor: rm caching for expensive work units

jl3329:
Caching was only previously enabled for fields labeled Expensive.
The code generating the cache key uses reflect.Comparable to see
if the source can be hashed. However, this cannot determine if the
underlying type of an interface is comparable.

This means that a struct with an interface field will panic
if the underlying type of the interface is non-comparable.

e.g.
```
type Foo interface {
    Bar()
}

type Uncomparable []int

func (u Uncomparable) Bar() {}

type Baz struct {
    Field Foo
}

// The following returns true. A source with type Baz will panic.
reflect.TypeOf(Baz{Field: Uncomparable([]int{1,2,3})})
```

Since caching was only enabled for fields labeled expensive, I
think it's reasonable to disable the caching for consistency.
Note that for non-websocket use cases, the cache would likely
not be used anyways.

If caching turns out to be important enough to reimplement,
consider a hashing the sources in a way such that non-comparable
fields are handled.


steveshi7:
What are the performance/load implications of making this change? Is there a way that we can roll this out safely?

PR #460, opened 2021-04-20
go federation: remove unnecessary fragments from flatten

steveshi7:
Currently, we are maintaining a list of all fragments that are not
applicable to the current type in flattenFragments. For each
selectionSet, we append the list of unused fragments to the selection
set's fragments. However, we never again use selectionSet.Fragments
again, even in child recursive calls, and do not need it to flatten
the rest of the selectionSet. Thus, it is safe to remove this
logic.

As a result of this logic, there was a bug: if multiple selections use the same
fragment, then we mutate the fragment's selectionSet in child recursive
calls. This results in a fragment.SelectionSet.Fragments array that
grows exponentially in size, and runtime grows exponentially as well as
we process that fragment array. Removing this logic fixes this bug.

### DIFF
--- a/federation/planner.go
+++ b/federation/planner.go
@@ -4,7 +4,8 @@ import (
 	"errors"
 	"fmt"
 	"sort"
-
+	"strings"
+	
 	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/graphql"
 )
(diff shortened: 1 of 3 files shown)


CodeBrew28 (approved):

federation/planner.go
```diff
@@ -122,19 +123,19 @@ func printPlan(rootPlan *Plan) {
 	}
 }
 
-func printSelections(selectionSet *graphql.SelectionSet) {
```
CodeBrew28, line 125: Thanks for cleaning up the `printSelections` logic! Should make debugging easier in the future!

PR #459, opened 2021-04-02
go sqlgen: Add SELECT ... FOR UPDATE support

aaron-zeisler:
Added a ForUpdate option to the SelectOptions, in order to lock rows in a transaction using the SELECT ... FOR UPDATE syntax.  The mysql documentation for locking reads can be found [here](https://dev.mysql.com/doc/refman/8.0/en/innodb-locking-reads.html).

PR #464, opened 2021-04-28
go sqlgen: Add SELECT ... FOR UPDATE support

aaron-zeisler:
Added a ForUpdate option to the SelectOptions, in order to lock rows in a transaction using the SELECT ... FOR UPDATE syntax.  The mysql documentation for locking reads can be found here: https://dev.mysql.com/doc/refman/5.6/en/innodb-locking-reads.html


aaron-zeisler:
Code review and conversation for this is in #459

PR #480, opened 2021-05-27
Testing nullable enum type

felixweinberger:


graphql/introspection/testdata/TestComputeSchemaJSON.snapshots.json
```diff
@@ -128,6 +130,45 @@
...
+                  "args": [],
+                  "deprecationReason": "",
+                  "description": "",
+                  "isDeprecated": false,
+                  "name": "maybeEnum",
+                  "type": {
+                    "kind": "SCALAR",
+                    "name": "int32",
+                    "ofType": null
+                  }
```
felixweinberger, line 163: I think this should be some kind of nullable ENUM, instead it becomes a scalar of type int32

PR #481, opened 2021-05-28
Add support for nullable enum types in schemaBuilder

felixweinberger:


felixweinberger:
> This LGTM. Would you mind clarifying in the commit message that we are adding support for nullable enum types in the return value, and that nullable enums can already be used as arguments?

Added comments to point this out in the commit description.

jl3329:
As you noted in our conversation, this is a breaking change because this can potentially change the return value type if there are existing endpoints that return a pointer to a go enum. Let's put that in the commit message PR description somewhere so that it's clear

jl3329:
Going to remove approval for now until we can safely apply this change to the Samsara repo. This is a breaking change because this changes the schema's return type for existing field funcs that return a pointer to a Go enum.

In order to make this change safely, we must change all of these existing field funcs to instead return a pointer to the underlying scalar type so that the return types do not change in the schema.

PR #483, opened 2021-07-17
go graphql: add description to fieldfuncs

zdylag:
This fieldfunc option will allow users to set a description
value for their field. This will be most useful for people
who browse generated docs or Graphiql since they can
see an actual message instead of "".

### DIFF
--- a/graphql/introspection/testdata/TestComputeSchemaJSON.snapshots.json
+++ b/graphql/introspection/testdata/TestComputeSchemaJSON.snapshots.json
@@ -56,10 +56,12 @@
               "name": "skip"
             },
             {
-                "args": [],
-                "description": "Client-side-only directive that instructs the type generator to mark this field as optional. This is useful for making the generated types compliant with Troy persistence schema.",
-                "locations": ["FIELD"],
-                "name": "type_as_optional"
+              "args": [],
+              "description": "Client-side-only directive that instructs the type generator to mark this field as optional. This is useful for making the generated types compliant with Troy persistence schema.",
+              "locations": [
+                "FIELD"
+              ],
+              "name": "type_as_optional"
             }
           ],
           "mutationType": {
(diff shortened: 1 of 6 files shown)


graphql/introspection/testdata/TestComputeSchemaJSON.snapshots.json
```diff
@@ -56,10 +56,12 @@
...
             {
-                "args": [],
-                "description": "Client-side-only directive that instructs the type generator to mark this field as optional. This is useful for making the generated types compliant with Troy persistence schema.",
-                "locations": ["FIELD"],
-                "name": "type_as_optional"
+              "args": [],
+              "description": "Client-side-only directive that instructs the type generator to mark this field as optional. This is useful for making the generated types compliant with Troy persistence schema.",
+              "locations": [
+                "FIELD"
+              ],
```
zdylag, line 63: @CodeBrew28 this diff gets generated for me, unclear why. I can revert it if it fails things.

CodeBrew28: Hm, it seems like just the formatting changed, which seems okay. Not sure why it was wrong before.

graphql/schemabuilder/output.go
```diff
@@ -125,6 +125,7 @@ func (sb *schemaBuilder) buildStruct(typ reflect.Type) error {
 				return err
```
CodeBrew28, line 125: I think it might be cleaner to add setting the description in the functions that build the field (eg in  buildFunctionAndFuncCtx, buildBatchFunctionAndFuncCtx, buildPaginatedFieldWithFallback, and buildFederatedFunction). Also test plans for those cases to make sure the descriptions bubbled up correctly into the introspection schema.

CodeBrew28 (commented):
Looks awesome, thanks for adding this! Left a few comments to make sure the descriptions work in all cases (batch, pagination, federation, etc), but other than that looks good!

PR #485, opened 2021-07-28
go federation: add fix for infinite recursion

steveshi7:


KevinXing:
I changed the deep copy to only copy the first layer of selections and fragments, the benchmark is as below: 
```
 5678	    209687 ns/op	   78809 B/op	    1776 allocs/op
```

KevinXing:
The root cause is `last.SelectionSet.Selections = append(last.SelectionSet.Selections, s)` sometimes create a new slice, sometimes not, when it does not create the new slice, we will run into the bug.

PR #487, opened 2021-07-30
normalize: fix a bug in mergeSameAlias

KevinXing:
In mergeSameAlias, we make a shadow copy of each selection and then append
to copy.selectionSet.selections later.
Since append in GO does not always create a new slice, there is chance we
are appending to the original selectionSet.selections slice. This could cause
the recursively flatten function to run into an infinite loop in some cases.

This commit make a new selectionSet for each copy so we will not append to the
original selections later in the function.

This change does not increase the runtime too much. 
```
Current
BenchmarkNormalize2-8   	    5182	    193595 ns/op	  101487 B/op	    2131 allocs/op

New
BenchmarkNormalize2-16    	    5040	    248245 ns/op	  105241 B/op	    2232 allocs/op
```

### DIFF
--- a/federation/normalize.go
+++ b/federation/normalize.go
@@ -144,8 +144,12 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
 		return selections[i].Alias < selections[j].Alias
 	})
 
+	// It is safe to initialize newSelections with selections[:0] because:
+	// 1. in mergeSameAlias, newSelections will override the value in selections only after the value is used.
+	// 2. the only caller of mergeSameAlias does not reference to the input selections after it calls mergeSameAlias.
 	newSelections := selections[:0]
 	var last *graphql.Selection
+	var isLastSelectionSetCopied bool
 	for _, selection := range selections {
 		if last == nil || selection.Alias != last.Alias {
 			// Make a copy of the selection so we can modify it below
@@ -160,6 +164,7 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
 			selection = &cp
 			newSelections = append(newSelections, selection)
 			last = selection
+			isLastSelectionSetCopied = false
 			continue
 		}
 
@@ -177,13 +182,18 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
 				return nil, fmt.Errorf("one selection with alias %s has subselections and one does not",
 					selection.Alias)
 			}
+
+			if !isLastSelectionSetCopied {
+				last.SelectionSet = last.SelectionSet.ShallowCopy()
+				isLastSelectionSetCopied = true
+			}
+
 			seenSelections := make(map[string]struct{}, len(selection.SelectionSet.Selections))
 			for _, s := range selection.SelectionSet.Selections {
 				if _, ok := seenSelections[s.Alias]; !ok {
 					seenSelections[s.Alias] = struct{}{}
 					last.SelectionSet.Selections = append(last.SelectionSet.Selections, s)
 				}
-
 			}
 			seenFragments := make(map[*graphql.Fragment]struct{}, len(selection.SelectionSet.Fragments))
 			for _, f := range selection.SelectionSet.Fragments {
--- a/federation/normalize.go
+++ b/federation/normalize.go
@@ -177,18 +177,18 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
 				return nil, fmt.Errorf("one selection with alias %s has subselections and one does not",
 					selection.Alias)
 			}
-			seenSelections := make(map[string]bool, len(last.SelectionSet.Selections))
+			seenSelections := make(map[string]struct{}, len(selection.SelectionSet.Selections))
 			for _, s := range selection.SelectionSet.Selections {
-				if !seenSelections[s.Alias] {
-					seenSelections[s.Alias] = true
+				if _, ok := seenSelections[s.Alias]; !ok {
+					seenSelections[s.Alias] = struct{}{}
 					last.SelectionSet.Selections = append(last.SelectionSet.Selections, s)
 				}
 
 			}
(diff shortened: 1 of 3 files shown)


KevinXing:
This is another try to fix the bug in https://github.com/samsarahq/thunder/pull/485.

joyontime (approved):
Discussed the difficulty of testing this spooky bug offline. lgtm.

hitotito (approved):

federation/normalize.go
```diff
@@ -150,8 +150,14 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
...
 			// Make a copy of the selection so we can modify it below
 			// or when we flatten recursively later.
-			copy := *selection
-			selection = &copy
+			cp := *selection
+			if selection.SelectionSet != nil {
+				// Make a new SelectionSet for the copy so we will not append to the original slice later.
+				cp.SelectionSet = &graphql.SelectionSet{}
+				cp.SelectionSet.Selections = append(cp.SelectionSet.Selections, selection.SelectionSet.Selections...)
+				cp.SelectionSet.Fragments = append(cp.SelectionSet.Fragments, selection.SelectionSet.Fragments...)
```
ccfrost, line 158: Maintainability: if we add a new field to SelectionSet, we'll need to remember to update this call site. What do you think of adding a method to SelectionSet that returns a shallow copy, instead, to try to make it easier to do that if we do add a field later?

federation/normalize.go
```diff
@@ -150,8 +150,14 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
...
+			}
+			selection = &cp
 			newSelections = append(newSelections, selection)
 			last = selection
 			continue
```
ccfrost, line 163: Optional, followup for below: it looks like we could change `map[string]bool` into `map[string]struct{}`. That might reduce memory usage slightly. Probably a small thing, but maybe useful? Sam for the Fragment map.

Also, the initial capacity of the maps looks suspicious. I wonder if it could be changed and might reduce allocations?
1. They use len(last) but add items proportional to len(selection). Would len(selection) be more appropriate?
2. The Fragment map uses len(...Selections) rather than len(..Fragments). WOuld len(...Fragments) be more appropriate?

federation/normalize.go
ccfrost, line 158: THanks for including the benchmark!
From it, I see that this increases runtime noticeably. In thinking about how we might lower it, I wonder how often the code creates a copy but doesn't append to one or both of the selectionsets? If that happens frequently, would it be worthwhile to only copy when the code first wants to append? We would need to track whether a copy needs to be made to permit modification, but that might avoids the allocations.

KevinXing: Yeah, I tried to do this and below is the benchmark: 

```
BenchmarkNormalize2-16    	    4471	    235298 ns/op	  102405 B/op	    2132 allocs/op
```

I have to add a map[*graphql.SelectionSet]struct{} to as the input of `flatten` and `mergeSameAlias` to track whether we need to shallow copy a certain selectionSet across the recursions. 

The allocation is similar as the original and the allocation per is slightly higher.

KevinXing: I updated the commit!

federation/normalize.go
```diff
@@ -171,18 +171,18 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
...
-				if !seenSelections[s.Alias] {
-					seenSelections[s.Alias] = true
+				if _, ok := seenSelections[s.Alias]; !ok {
+					seenSelections[s.Alias] = struct{}{}
 					last.SelectionSet.Selections = append(last.SelectionSet.Selections, s)
 				}
 
 			}
-			seenFragments := make(map[*graphql.Fragment]bool, len(last.SelectionSet.Selections))
+			seenFragments := make(map[*graphql.Fragment]struct{}, len(selection.SelectionSet.Fragments))
```
ccfrost, line 189: In the commit message also call out that this corrects the initial size to be based on Fragments rather than Selections.
I expect that there are generally many more Selections than Fragments, so I think this means that this allocated space that would go unused, but it didn't cause additional allocations.

graphql/types.go
```diff
@@ -173,6 +173,20 @@ type SelectionSet struct {
...
+	if s.Selections != nil {
+		cp.Selections = make([]*Selection, len(s.Selections))
+		copy(cp.Selections, s.Selections)
+	}
+	if s.Fragments != nil {
+		cp.Fragments = make([]*Fragment, len(s.Fragments))
+		copy(cp.Fragments, s.Fragments)
+	}
+	return cp
+}
```
ccfrost, line 188: Just in case: do you want to add a unit test? (What's the expectation for this package?)

federation/normalize.go
```diff
@@ -146,6 +146,7 @@ func mergeSameAlias(selections []*graphql.Selection) ([]*graphql.Selection, erro
 
 	newSelections := selections[:0]
```
ccfrost, line 147: For this or a future PR: This looks fishy, because of how Go slices are implemented. I think that this doesn't cause an issue with the current code, but that it has a rarely-expected effect that might surprise a future dev.

I believe that the effect of this variable initialization is that that any appends to `newSelections` will overwrite the existing values in `selections`. That might actually be fine in this code, if both:
1) The caller is okay with their `selections` variable being modified
2) Once item number i is appended newSelections, then no item at index j in selection is read, where j <= i. 
I haven't checked property 1, but property 2 looks like it holds.

Two guesses on the intention of this code:
A) It is accidental. If so, how about we change it to be `newSelections := make([]*graphql.Selection, 0, len(selections)`
B) It is intentional. This approach avoids allocating a new slice. Maybe that's an important performance property? If we think this is the case, would you add a comment for future readers, so that they don't miss the subtlety?

Example of the slice overwrite behavior: https://play.golang.org/p/cgG8JSRsaRX

ccfrost (approved):
I left a few feedbacks. Otherwise LGTM!

graphql/types.go
KevinXing: I added a unit test!

PR #494, opened 2021-08-06
go batch_executor: export WorkUnit fields useBatch and field

EmmaOnThursday:
At query execution time, we would like to log whether the query being executed is batched and/or paginated. To do this reliably, we need access to these two fields on WorkUnit.

### DIFF
--- a/graphql/batch_executor.go
+++ b/graphql/batch_executor.go
@@ -16,11 +16,11 @@ import (
 // graphql query.
 type WorkUnit struct {
 	Ctx          context.Context
-	field        *Field
+	Field        *Field
 	selection    *Selection
 	sources      []interface{}
 	destinations []*outputNode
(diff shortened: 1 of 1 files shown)


graphql/batch_executor.go
```diff
@@ -16,11 +16,11 @@ import (
 // graphql query.
 type WorkUnit struct {
 	Ctx          context.Context
-	field        *Field
+	Field        *Field
```
KevinXing, line 19: I am not comfortable to expose these field since the caller will be able to edit them. 
How about we add corresponding methods on WorkUnit struct?

EmmaOnThursday: That can work!

EmmaOnThursday:
i'm going to close this PR in favor of one with an access method approach.

PR #495, opened 2021-08-09
graphql/batch_executor: expose IsExpensive and IsBatch workunit fields

EmmaOnThursday:
In order to tag query executions as batched or not, and expensive or not, we expose methods on the work unit that provide this info.


KevinXing (approved):
lgtm!

Issue #504, opened 2021-11-30
Example in pagination.md not korrect


JonasGutermuth:
Line 243: schemabuilder.PaginatedArgs should be schemabuilder.PaginationArgs

PR #505, opened 2021-11-30
Fixes https://github.com/samsarahq/thunder/issues/504

JonasGutermuth:


KevinXing (approved):
Thanks for the fixing!

PR #506, opened 2022-01-13
Update broken link in pagination.md

daniellowtw:
The current link returns a 404. Using waybackmachine I could confirm the new link now points to the same content that was referenced before.


KevinXing (approved):
Thanks for the update!

PR #510, opened 2022-02-11
graphql/schemabuilder/schema: enforce unique type names

jl3329:
Introspection assumes that each type in the schema is unique.
Note that package names are not considered when comparing type names.
In other words, foo_package.SomeType and bar_package.SomeType
are considered to have the same name.

This causes issues with the type resolution as only one type
of each name can exist in the schema. For example, one might
define 2 fieldfuncs, one returning foo_package.SomeType and
another returning bar_package.SomeType. The introspection
query will incorrectly declare that both fieldfuncs return
the same type: one of foo_package.SomeType or bar_package.SomeType.

In order to avoid type collisions, we will fail to build the
schema if we detect duplicate types in the GraphQL schema.

We iterate over every object's registered fieldfuncs,
recursing through each fieldfunc's argument and return types to
collect a set of types. Then, we compare these types to the
set of enum types.

Note that there is an existing type name uniqueness check in
getType. However, this only checks return types and does not
correctly compare slice or pointer types. For example,
registering both foo_package.SomeType and []*bar_package.SomeType
will not throw an error. The test for the existing check has been
removed in favor of the new tests introduced in this commit.

### DIFF
--- a/graphql/schemabuilder/schema.go
+++ b/graphql/schemabuilder/schema.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/graphql"
 )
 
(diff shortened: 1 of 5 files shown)


graphql/schemabuilder/schema.go
```diff
@@ -227,6 +272,60 @@ func (s *Schema) Build() (*graphql.Schema, error) {
...
+
+			// It is valid for federated fieldfuncs to contain a nil Fn field.
+			// We will validate registered fieldfuncs later in getType.
+			if fieldFuncKind := reflect.ValueOf(method.Fn).Kind(); fieldFuncKind != reflect.Func {
+				continue
+			}
+
+			fieldFuncType := reflect.TypeOf(method.Fn)
+
+			for i := 0; i < fieldFuncType.NumIn(); i++ {
```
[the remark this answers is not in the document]
KevinXing: Yeah, I think force the consistency between input argument type and return type seems better!

graphql/schemabuilder/schema.go
```diff
@@ -227,6 +272,60 @@ func (s *Schema) Build() (*graphql.Schema, error) {
...
+	}
+
+	for enum := range sb.enumMappings {
+		enumType := enum.Name()
+		enumPkg := enum.PkgPath()
+		if packageName, ok := typesToPackages[enumType]; ok && packageName != enumPkg {
+			return nil, fmt.Errorf(DuplicateTypeNameErrFormat, enumType, packageName, enumPkg)
+		}
+
+		typesToPackages[enumType] = enumPkg
```
KevinXing, line 328: Readability: do you think move this entire block to a helper function will make the `Build` easier to read? Since the main functionality of the 'Build` is to build the schema rather than checking the uniqueness.

graphql/schemabuilder/schema.go
```diff
@@ -199,6 +199,50 @@ func (s *Schema) Mutation() *Object {
...
+
+	if seenPackage, ok := typesToPackages[typeName]; ok && seenPackage != packageName {
+		return fmt.Errorf(DuplicateTypeNameErrFormat, typeName, seenPackage, packageName)
+	} else if ok && seenPackage == packageName {
+		return nil
+	}
+
+	typesToPackages[typeName] = packageName
+
+	if typ.Kind() != reflect.Struct {
```
KevinXing, line 232: How can we run into this?

KevinXing (commented):
Only one question!

graphql/schemabuilder/schema.go
jl3329: Done!

graphql/schemabuilder/schema.go
jl3329: Ah this was because of the special `graphql.SelectionSet` type. This type is actually excluded from the introspection result so I decided to do a type check for it instead here: https://github.com/samsarahq/thunder/pull/510/commits/423b78717b35e964c018fccb076298255eb2a94e#diff-414e6a761a1e5643cfd90f7d667ca5d2d5037726aee49a5864c7ace2a8b7b453R290

KevinXing (approved):
Nice work! Thanks for the detailed tests!

PR #518, opened 2022-09-09
Kevinxing/introduce non nullable list typing

KevinXing:
Breaking change: stop forcing the list entry to be nonnullable.
When return type of a fieldfunc is a go list with pointer entry like []*Item,
thunder will always generate [Item!] as the GraphQL type. This implicity causes
several problems in Samsara as client assumes the list is nonnullable whereas
backend could return a nil pointer.

To solve the problem, this commit is making a breaking change to stop forcing the
list entry to be nonnullable. E.g. []*Item now will be converted to [Item] instead
of [Item!].

In order to keep backward compatibility, this commit introduces a new schemabuilder
option ListEntryNonNullable, when set true, list entry will be forced to be nonnullable
as before.

Since this is a breaking change, we will also provide a code generator to add the
new `ListEntryNonNullable` to all the existing fieldfuncs to keep the backward compatibility.


graphql/schemabuilder/schema.go
```diff
@@ -123,7 +123,8 @@ type federation struct{}
 
 func FetchObjectFromKeys(f interface{}, options ...ObjectOption) ObjectOption {
 	// Create a method on the "Federation" object to create the shadow object from the federated keys
-	m := &method{Fn: f, Expensive: true}
+	// Force the returned list entry to be nullable to keep backward compatibility.
+	m := &method{Fn: f, Expensive: true, MarkedListEntryNonNullable: true}
```
steveshi7, line 127: what are the implications of this?

steveshi7: bump on this!

steveshi7 (approved):
looks good!

PR #522, opened 2022-10-24
go executor: move append inside mutex

alberttjin:
go append is not thread safe, so we should put it inside the mutex


KevinXing (approved):
Thanks!

PR #524, opened 2022-10-31
go federation: add pagination tests

alberttjin:
Federation has no pagination tests. This commit copies over the pagination tests we have for non-federated endpoints and federate objects to see if they still work

### DIFF
--- a/federation/executor_pagination_test.go
+++ b/federation/executor_pagination_test.go
@@ -1,0 +1,2036 @@
+package federation
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/samsarahq/thunder/batch"
+	"github.com/samsarahq/thunder/graphql"
+	"github.com/samsarahq/thunder/graphql/schemabuilder"
+	"github.com/samsarahq/thunder/internal/testgraphql"
+
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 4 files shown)


alberttjin:
the tests are the same as in `connection_test.go` if that helps making reviewing easier

I pretty much just federated all the objects and dealth with some string matching errors

federation/executor_test.go
```diff
@@ -273,6 +273,7 @@ func makeExecutors(schemas map[string]*schemabuilder.Schema) (map[string]Executo
 	return executors, nil
 }
 
+
```
KevinXing, line 276: Revert this?

federation/executor_pagination_test.go
```diff
@@ -0,0 +1,2036 @@
...
+
+	inner := schema.Object("inner", Inner{}, schemabuilder.FetchObjectFromKeys(func(args struct{ Keys []*Inner }) []*Inner {
+		return args.Keys
+	}))
+	item := schema.Object("item", Item{}, schemabuilder.FetchObjectFromKeys(func(args struct{ Keys []*Item }) []*Item {
+		return args.Keys
+	}))
+
+	item.Key("id")
+	inner.FieldFunc("innerConnection", func(args Args) []Item {
```
KevinXing, line 62: Can we make `innerConnection` and `inner` on different schemas?

PR #525, opened 2022-11-02
Weilian/expand federation schema

weilianchu:
This PR allows us to get an introspection query result from a merged federation schema. This output will exactly match an introspection output, allowing users to get the full introspection query result without a single service containing the whole schema.

### DIFF
--- a/federation/merge_schemas.go
+++ b/federation/merge_schemas.go
@@ -4,6 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+
+	"github.com/samsarahq/thunder/graphql"
+	"github.com/samsarahq/thunder/graphql/introspection"
 )
 
 // MergeMode controls how to combine two different schemas. Union is used for
@@ -75,10 +78,27 @@ type introspectionField struct {
 	Name string                    `json:"name"`
 	Type *introspectionTypeRef     `json:"type"`
 	Args []introspectionInputField `json:"args"`
+
+	DeprecationReason string `json:"deprecationReason"`
+	Description       string `json:"description"`
+	IsDeprecated      bool   `json:"isDeprecated"`
 }
 
 type introspectionEnumValue struct {
 	Name string `json:"name"`
+}
+
+type introspectiveDirectiveInputValue struct {
+	Name         string                `json:"name"`
+	Description  string                `json:"description"`
+	DefaultValue *string               `json:"defaultValue"`
+	Type         *introspectionTypeRef `json:"type"`
+}
+type introspectiveDirective struct {
+	Name        string                             `json:"name"`
+	Description string                             `json:"description"`
+	Locations   []string                           `json:"locations"`
+	Args        []introspectiveDirectiveInputValue `json:"args"`
 }
 
 type introspectionType struct {
@@ -88,10 +108,15 @@ type introspectionType struct {
 	InputFields   []introspectionInputField `json:"inputFields"`
 	PossibleTypes []*introspectionTypeRef   `json:"possibleTypes"`
 	EnumValues    []introspectionEnumValue  `json:"enumValues"`
+	Description   string                    `json:"description"`
+	Interfaces    []*introspectionTypeRef   `json:"interfaces"`
 }
 
 type introspectionSchema struct {
-	Types []introspectionType `json:"types"`
+	Types        []introspectionType      `json:"types"`
+	QueryType    introspectionType        `json:"queryType"`
+	MutationType introspectionType        `json:"mutationType"`
+	Directives   []introspectiveDirective `json:"directives"`
 }
 
 type IntrospectionQueryResult struct {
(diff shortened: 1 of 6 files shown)


federation/schema.go
```diff
@@ -196,14 +192,45 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
...
 		serviceSchemasByName[service] = serviceSchema
-
 		serviceSchemas = append(serviceSchemas, serviceSchema)
 	}
 
+	return serviceNames, serviceSchemas, serviceSchemasByName, nil
+}
+
+// MergeSchemas takes the introspection query results for all
+// verions of all executor services, and merges them into a
```
alberttjin, line 206: nit: versions

federation/merge_schemas.go
```diff
@@ -312,6 +338,36 @@ func mergeEnumValues(a, b []introspectionEnumValue, mode MergeMode) ([]introspec
...
+	names := make([]string, 0, len(directives))
+	for name := range directives {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+
+	merged := make([]introspectiveDirective, 0, len(names))
+	for _, name := range names {
+		p := directives[name]
+		if len(p) == 1 {
```
alberttjin, line 358: noob question:

what if the mode is union and the length of p is more than 1, but the two values are different? Then wouldn't we just be taking the first value? Should we check that they are the same similar to how we merge types below?

federation/merge_schemas.go
```diff
@@ -312,6 +338,36 @@ func mergeEnumValues(a, b []introspectionEnumValue, mode MergeMode) ([]introspec
 	return merged, nil
 }
 
+func mergeDirectives(a, b []introspectiveDirective, mode MergeMode) ([]introspectiveDirective, error) {
```
KevinXing, line 341: Unlike the types that can be different in gql servers, directive is a global feature that should be supported by the merged schema. 
Our directive introspection queries are hardcoded pre federation. I think it is reasonable to also hardcode them in the merged introspection schema.

weilianchu: Is the implication here that the directives will be the same for all federated schemas?

KevinXing: I think so, directives is a global feature.

federation/schema.go
```diff
@@ -196,14 +192,45 @@ func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo,
...
+// single introspection query result that represents the full
+// federated schema.
+func MergeSchemas(schemas serviceSchemas) (*IntrospectionQueryResult, error) {
```
KevinXing, line 209: Let's add an unit test for this function?

KevinXing, line 209: Maybe rename this function to `MergeIntrospectionSchema`? `MergeSchema` feels merge the full schemas to me.

federation/merge_schemas.go
weilianchu: Yeah, if two directives had the same names, we would only take the first directive. I believe this is the same behaviour as how we merge type fields as well. However, as Zhekai pointed out, the directives of an introspection query are actually hardcoded, so I will be removing this method altogether.

federation/merge_schemas.go
```diff
@@ -312,6 +338,36 @@ func mergeEnumValues(a, b []introspectionEnumValue, mode MergeMode) ([]introspec
...
+	directives := make(map[string][]introspectiveDirective)
+	for _, a := range a {
```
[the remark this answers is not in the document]
weilianchu: since the directives are hardcoded I will remove this method

weilianchu:
@KevinXing ready for review again

federation/schema.go
```diff
@@ -165,11 +165,7 @@ func validateFieldsReturningFederatedObject(serviceNames []string, serviceSchema
 	return nil
 }
 
-// ConvertVersionedSchemas takes schemas for all of versions of
-// all executors services and generates a single merged schema
-// annotated with mapping from field to all services that know
-// how to resolve the field
-func ConvertVersionedSchemas(schemas serviceSchemas) (*SchemaWithFederationInfo, error) {
+func processSchemaVersions(schemas serviceSchemas) ([]string, []*IntrospectionQueryResult, map[string]*IntrospectionQueryResult, error) {
```
KevinXing, line 168: Let's also add a comment for this function to explain the outputs?

federation/merge_schemas.go
```diff
@@ -364,7 +402,43 @@ func mergeTypes(a, b introspectionType, mode MergeMode) (*introspectionType, err
...
+	return directive
+}
+
 func mergeSchemas(a, b *IntrospectionQueryResult, mode MergeMode) (*IntrospectionQueryResult, error) {
+	if a.Schema.QueryType.Name != b.Schema.QueryType.Name {
+		return nil, fmt.Errorf("can't merge QueryType %s and %s", a.Schema.QueryType.Name, b.Schema.QueryType.Name)
+	}
+	if a.Schema.MutationType.Name != b.Schema.MutationType.Name {
+		return nil, fmt.Errorf("can't merge MutationType %s and %s", a.Schema.MutationType.Name, b.Schema.MutationType.Name)
+	}
```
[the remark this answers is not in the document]
weilianchu: yep, i wrote this for the general case, but if we're specifically dealing with introspection query results, i can hardcode.

weilianchu:
@KevinXing ready for review again

KevinXing (approved):
Nice!

PR #526, opened 2022-11-10
go sqlgen: Add support for mysql FORCE INDEX

aaron-zeisler:
Added support for mysql's FORCE INDEX clause. This allows the developer to specify one or more indexes that the mysql optimizer should choose. For more information, check out the official
documentation: https://dev.mysql.com/doc/refman/8.0/en/index-hints.html.


sqlgen/mysql.go
```diff
@@ -80,6 +81,10 @@ func (q *SelectQuery) ToSQL() (string, []interface{}) {
 	buffer.WriteString(" FROM ")
 	buffer.WriteString(q.Table)
 
+	if len(q.Options.ForceIndex) > 0 {
```
THoelzel, line 84: Have you considered also adding `USE INDEX()`? As this is an open-source library providing the option for both might be nice.

aaron-zeisler: OK, I just updated the PR to add `USE INDEX` as well 👍

sqlgen/mysql.go
```diff
@@ -80,6 +81,12 @@ func (q *SelectQuery) ToSQL() (string, []interface{}) {
...
+		buffer.WriteString(fmt.Sprintf(" FORCE INDEX(%s)", strings.Join(q.Options.ForceIndex, ",")))
```
ianvernon, line 85: The convention of the library is to use 

```
buffer.WriteString(" <MySQL directive ")
buffer.WriteString(<table / input>)
```

I wonder if we want to avoid using `fmt.Sprintf`? I'm not sure why (maybe something to do with allocations). But would be nice to follow the convention of the library.

aaron-zeisler: Updated. Hopefully `strings.Join()` is acceptable. If not, let me know.

batjaa (approved):

KevinXing:
Hello, do you mind to put [thunder](https://github.com/samsarahq/thunder) PR first?

sqlgen/mysql.go
```diff
@@ -3,6 +3,7 @@ package sqlgen
 import (
 	"bytes"
 	"fmt"
+	"strings"
```
KevinXing, line 6: Please update the mysql doc to 5.7 version link.

KevinXing (approved):
Nice!

sqlgen/mysql_test.go
```diff
@@ -129,6 +129,67 @@ func TestSelectQuery(t *testing.T) {
...
+		Options: &SelectOptions{
+			Where:    "foo = ? AND bar = ?",
+			Values:   []interface{}{25, "xyz"},
+			OrderBy:  "bar",
+			UseIndex: []string{"best_index", "another_good_index"},
+		},
+	}, "SELECT foo, bar FROM foo USE INDEX(best_index,another_good_index) WHERE foo = ? AND bar = ? ORDER BY bar", []interface{}{25, "xyz"}, t)
+
+	// Both USE INDEX and FORCE INDEX are specified: only FORCE INDEX is applied
+	testQuery(&SelectQuery{
```
ianvernon, line 182: Can you add a test case which shows what happens  if there is a bad index that doesn't exist?

aaron-zeisler: OK. It will result in a runtime error (Error 1176), similar to the situation where an invalid table name or column name was provided in the SELECT query.

aaron-zeisler: Added the test in that last push 👍

PR #528, opened 2022-11-17
go federation: add pagination tests

alberttjin:
This commit adds paginated field funcs to our current federation endpoint tests to make sure that even with multiple hops (fields across different schemas), the result from the query still is correct


KevinXing (approved):
Thanks!

PR #530, opened 2022-12-02
Pass args to sort and filter methods

levine404:
Allow users to provide filter and sort fields that implement the connection arguments of the FieldFunc.

This allows for pre-filtering on sort fields in the event that a sort field has to be filtered before providing sort values.  It also provides more flexibility in the filter fields and what type of action they can take if the arguments are passed directly to them.

Passing the ConnectionArgs.Args as userArgs to the GraphQL resolvers through the applySort and applyTextFilterField methods exposing the userArgs down to the subsequent schemabuilder methods (FilterField, BatchFilterField, SortField, BatchSortField)

This allows us to take action on these userArgs at the filter/sort level such as within a sort method, filtering on some userArgs before actually populating sort values. For example, filtering on start/end time within a sort method so that sorted values are based on a higher order filter.

For example:
```go
type UserArgs struct {
  StartTime *int64
  EndTime *int64
}

type Item struct {
  Id         int64
  Number     int64
  String     string
  Float      float64
}

inner := schema.Object("inner", Inner{})
inner.FieldFunc("innerConnectionWithPassedArgs", func(args Args) []Item {
    return []Item{
      {Id: 1, Number: 1, String: "1", Float: 1.0},
      {Id: 2, Number: 3, String: "3", Float: 3.0},
      {Id: 3, Number: 5, String: "5", Float: 5.0},
    }
  },
  schemabuilder.Paginated,
  schemabuilder.BatchSortField(
    "batchSortWithArgs",
    func(ctx context.Context, items map[batch.Index]Item, args Args) (map[batch.Index]int64, error) {
      // Batch sort using args
    }),
  schemabuilder.BatchFilterField(
    "batchFilterWithArgs",
    func(items map[batch.Index]Item, args Args) (map[batch.Index]string, error) {
      // Batch filter using args
    }),
)
```

### DIFF
--- a/graphql/connection_test.go
+++ b/graphql/connection_test.go
@@ -35,12 +35,23 @@ type ManualArgs struct {
 	PaginationArgs schemabuilder.PaginationArgs
 }
 
+type TimeBasedArgs struct {
+	StartMs int64
+	EndMs int64
+}
+
 type Item struct {
 	Id         int64
 	FilterText string
 	Number     int64
 	String     string
 	Float      float64
+}
+
+type AltItem struct {
+	Id int64
+	Time int64
+	Number int64
 }
 
 func TestConnection(t *testing.T) {
@@ -633,8 +644,7 @@ func TestPaginateBuildFailure(t *testing.T) {
 		))
 
 		_, err := schema.Build()
-		require.Error(t, err)
-		require.Contains(t, err.Error(), "text filter fields can't take arguments")
+		require.NoError(t, err)
 	})
 
 	t.Run("non-string sort return", func(t *testing.T) {
@@ -684,8 +694,7 @@ func TestPaginateBuildFailure(t *testing.T) {
 		))
 
 		_, err := schema.Build()
-		require.Error(t, err)
-		require.Contains(t, err.Error(), "sort fields can't take arguments")
+		require.NoError(t, err)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


graphql/connection_test.go
```diff
@@ -1328,6 +1375,57 @@ func TestPaginatedFilters(t *testing.T) {
...
+					}
+				}
+			}
+		}`, nil)
+	if err = graphql.PrepareQuery(context.Background(), builtSchema.Query, q.SelectionSet); err != nil {
+		t.Error(err)
+	}
+	e = testgraphql.NewExecutorWrapper(t)
+	val, err = e.Execute(context.Background(), builtSchema.Query, nil, q)
+	assert.Nil(t, err)
```
jl3329, line 1398: nit: `assert.NoError(t, err)`?

graphql/connection_test.go
```diff
@@ -1328,6 +1375,57 @@ func TestPaginatedFilters(t *testing.T) {
...
+					edges {
+						node {
+							id
+						}
+						cursor
+					}
+				}
+			}
+		}`, nil)
+	if err = graphql.PrepareQuery(context.Background(), builtSchema.Query, q.SelectionSet); err != nil {
```
jl3329, line 1393: nit: let's be consistent with using either assert or explicit error checks

graphql/connection_test.go
```diff
@@ -1276,6 +1274,55 @@ func TestPaginatedFilters(t *testing.T) {
...
+			{Id: 8, FilterText: "ban", String: "a"},
+			{Id: 9, FilterText: "dan", String: "a"},
+			{Id: 10, FilterText: "ean", String: "a"},
+			{Id: 11, FilterText: "fan", String: "a"},
+			{Id: 12, FilterText: "gan", String: "a"},
+		}
+
+	}, schemabuilder.Paginated,
+		schemabuilder.BatchFilterField("filterTextBatched",
+			func(items map[batch.Index]Item, args Args) (map[batch.Index]string, error) {
```
jl3329, line 1297: Do you mind providing a test case in which the user args are used for filtering as well as one for sorting? I think it would serve as useful documentation. Personally, I am also a little unclear on the use case for providing user arguments to filter and sort functions

mwbennett:
@levine404 could we link the RFC doc in the PR comment to connect the functionality change here to the use case we have with centralized trip assignments?

jl3329:
> @levine404 could we link the RFC doc in the PR comment to connect the functionality change here to the use case we have with centralized trip assignments?

Let's not add Samsara-specific artifacts as this is a public repo

jl3329 (approved):
It looks like there are some outstanding nits but overall, LGTM!

PR #535, opened 2022-12-22
Weilian/federation introspection client

weilianchu:
This PR adds an introspection client to the federation executor. This introspection client runs at the gateway executor and will serve all introspection queries. 

After collecting all the schemas, the SchemaSyncer will combine all federated schemas, and run an introspection query on it, populating the result to the introspection client. This allows an introspection query to be served as if only one graphql schema is running.

### DIFF
--- a/federation/introspection_client.go
+++ b/federation/introspection_client.go
@@ -1,0 +1,58 @@
+package federation
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/samsarahq/thunder/graphql/introspection"
+	"github.com/samsarahq/thunder/graphql/schemabuilder"
+)
+
+const IntrospectionClientName = "introspectionclient"
+
+type IntrospectionClient struct {
+	IntrospectionQueryResult []byte
+}
+
+func NewIntrospectionClient(res []byte) *IntrospectionClient {
+	return &IntrospectionClient{
+		IntrospectionQueryResult: res,
+	}
+}
+
+func (c *IntrospectionClient) Execute(ctx context.Context, request *QueryRequest) (*QueryResponse, error) {
+	return &QueryResponse{
+		Result: c.IntrospectionQueryResult,
+	}, nil
+}
+
+// AddIntrospectionQueryToSchemaVersions will take in a map of service name to schema version
+// to introspection query result, and append to it with another service called "introspectionclient"
+// that is capable of only serving introspection queries. This will allow the federation executor to
+// directly serve introspection queries via its built-in introspection client rather than relay its
+// query to a subservice.
+func AddIntrospectionQueryToSchemaVersions(schemas map[string]map[string]*IntrospectionQueryResult) (map[string]map[string]*IntrospectionQueryResult, error) {
+	schema := schemabuilder.NewSchemaWithName("introspectionclient")
+	schema.Query()
+	schema.Mutation()
+
+	gqlSchema := schema.MustBuild()
+	introspection.AddIntrospectionToSchema(gqlSchema)
+
+	res, err := introspection.RunIntrospectionQuery(introspection.BareIntrospectionSchema(gqlSchema))
+	if err != nil {
+		return nil, err
+	}
+
+	var introspectionClientSchema IntrospectionQueryResult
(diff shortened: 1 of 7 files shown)


graphql/introspection/introspection.go
```diff
@@ -444,3 +444,17 @@ func RunIntrospectionQuery(schema *graphql.Schema) ([]byte, error) {
 
 	return json.MarshalIndent(value, "", "  ")
 }
+
+func IntrospectionQueryTypeOrSelection(typeName, selectionName string) bool {
+	if selectionName == "__schema" ||
```
alberttjin, line 449: nit: can probably just return the result of this if statement

alberttjin (approved):
great lgtm! 
still need to make a pr on backend repo with legacy fork right?

i only made sure everything made sense in the context of the PR, don't think i have enough context on how it might affect other things so zhekai should probably review too to be safe :)

federation/introspection_client.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+func (c *IntrospectionClient) Execute(ctx context.Context, request *QueryRequest) (*QueryResponse, error) {
+	return &QueryResponse{
+		Result: c.IntrospectionQueryResult,
+	}, nil
+}
+
+func AddIntrospectionQueryToSchemaVersions(schemas map[string]map[string]*IntrospectionQueryResult) (map[string]map[string]*IntrospectionQueryResult, error) {
```
KevinXing, line 29: Let's add a comment to this function?

federation/introspection_client_test.go
```diff
@@ -0,0 +1,50 @@
...
+	assert.Nil(t, err)
+
+	assert.Equal(t, schemaVersions["service1"], schemaVersionsWithIntrospection["service1"])
+	assert.Equal(t, schemaVersions["service2"], schemaVersionsWithIntrospection["service2"])
+
+	introspectionSchema := schemaVersionsWithIntrospection[IntrospectionClientName][""]
+	assert.NotNil(t, introspectionSchema)
+
+	var containsSchemaField bool
+	for _, typ := range introspectionSchema.Schema.Types {
```
KevinXing, line 43: Maybe snapshot the introspection query result?

federation/schema_syncer.go
```diff
@@ -27,43 +26,50 @@ func NewIntrospectionSchemaSyncer(ctx context.Context, executors map[string]Exec
 	return ss
 }
 
-func (s *IntrospectionSchemaSyncer) FetchPlanner(ctx context.Context) (*Planner, error) {
+func (s *IntrospectionSchemaSyncer) FetchPlannerAndIntrospectionQueryResult(ctx context.Context) (*Planner, []byte, error) {
 	schemas := make(map[string]*IntrospectionQueryResult)
```
KevinXing, line 30: I think we don't need this map now.

federation/schema_syncer.go
```diff
@@ -27,43 +26,50 @@ func NewIntrospectionSchemaSyncer(ctx context.Context, executors map[string]Exec
...
 
-	var iq IntrospectionQueryResult
-	if err := json.Unmarshal(schema, &iq); err != nil {
-		return nil, oops.Wrapf(err, "unmarshaling introspection schema")
+	mergedIqBytes, err := json.Marshal(mergedIq)
+	if err != nil {
+		return nil, nil, oops.Wrapf(err, "marshalling introspection schema")
 	}
 
-	schemas["introspection"] = &iq
```
KevinXing, line 62: It seems we can also just change this key to "introspectionclient" rather than merging the introspection of each sub schemas? Can you explain a bit more about why we need to query the introspection query separately?

graphql/introspection/introspection.go
KevinXing, line 449: This function is not used anywhere?

weilianchu: my plan is to use this helper in our backend repo - I can move it there directly.

federation/schema_syncer.go
weilianchu: oh yeah you are right - will make that change

PR #536, opened 2023-01-04
federation: add obj passing support

alberttjin:
Sometimes its easier to define an object as a federation key rather than re-fetching the object from the db. It can also be more cost efficient if its difficult to fetch it from the database.

However, to support this, the selection set needs to recursively list all fields under all objects that are selected as keys. This commit adds that support and also a comment explanation

**This PR only supports object recursion, non nulls, and lists. Things like unions will still behave as they have been before. They can definitely be added later but to keep this PR small I just added these. Will add more in later PRs.**

Tested this locally as well. Refer to PR description in https://github.com/samsara-dev/backend/pull/148236#issue-1505363090


federation/executor_test.go
```diff
@@ -166,11 +198,22 @@ func createExecutorWithFederatedUser() (*Executor, *schemabuilder.Schema, *schem
...
+	s2.Object("Abilities", Abilities{}, schemabuilder.FetchObjectFromKeys(func(args struct{ Keys []*Abilities }) []*Abilities {
+		return args.Keys
+	}))
+
 	userWithContactInfo.Key("id")
 	userWithContactInfo.FieldFunc("secret", func(ctx context.Context, user *UserWithContactInfo) (string, error) {
 		return "shhhhh", nil
 	})
 
+	userWithContactInfo.FieldFunc("getSkills", func(ctx context.Context, user *UserWithContactInfo) (Skills, error) {
```
KevinXing, line 213: nit: name the fieldfunc with noun, maybe just "skills".

federation/executor_test.go
```diff
@@ -453,6 +503,111 @@ func TestExecutorQueriesBasic(t *testing.T) {
 	}
 }
 
+// This function checks that we are able to query for and receive an object that was passed between gql servers
+// Specifically, it tests the functions in planner_helpers.go
+func TestExecutorQueriesWithObjectKey(t *testing.T) {
```
KevinXing, line 508: I think the test does not verify that we are passing object as keys since `s1` is able to resolve all the fields. 
I think we want to remove the `UserSkills` from `s1` schema. 
Also, I would suggest to add a test in planner_test.go to directly verify the planner output.

federation/planner_helpers.go
```diff
@@ -0,0 +1,162 @@
...
+-----------------------------------------
+This code is used to recursively generate a selection set for an object. The purpose is to allow for passing entire object (not scalar)
+fields from team gql service -> team gql service. This can be used when the object fetch from db is too complicated/expensive and its easier
+to simply pass the object rather than refetch it from keys.
+Example:
+Let's say we have object Devices that has been federated:
+		type Devices struct
+		{
+			Id int | key
+			Location LocationInput | key
```
KevinXing, line 17: I think this means that `Location` also needs to be defined as a GraphQL object with `FetchObjectFromKeys`.
I'd like to call this out somewhere in the documentation.

federation/planner_helpers.go
```diff
@@ -0,0 +1,165 @@
...
+			Name:         name,
+			Alias:        name,
+			UnparsedArgs: map[string]interface{}{},
+		}
+	}
+
+	return selectionResult
+}
+
+func getFederatedSelectionsForNonNull(typ *graphql.NonNull, name string, service string, selectionsByService map[string][]*graphql.Selection) *graphql.Selection {
```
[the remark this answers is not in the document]
alberttjin: yea but it passes in a nonnull type instead of a list

federation/planner_helpers.go
alberttjin: done

federation/executor_test.go
alberttjin: done

KevinXing (approved):
Nice!

PR #538, opened 2023-01-18
Pass args down to sort and filter methods

levine404:
Allow users to provide filter and sort fields that implement the connection arguments of the FieldFunc.

This allows for pre-filtering on sort fields in the event that a sort field has to be filtered before providing sort values.  It also provides more flexibility in the filter fields and what type of action they can take if the arguments are passed directly to them.

Passing the ConnectionArgs.Args as userArgs to the GraphQL resolvers through the applySort and applyTextFilterField methods exposing the userArgs down to the subsequent schemabuilder methods (FilterField, BatchFilterField, SortField, BatchSortField)

This allows us to take action on these userArgs at the filter/sort level such as within a sort method, filtering on some userArgs before actually populating sort values. For example, filtering on start/end time within a sort method so that sorted values are based on a higher order filter.

For example:
```go
type UserArgs struct {
  StartTime *int64
  EndTime *int64
}

type Item struct {
  Id         int64
  Number     int64
  String     string
  Float      float64
}

inner := schema.Object("inner", Inner{})
inner.FieldFunc("innerConnectionWithPassedArgs", func(args UserArgs) []Item {
    return []Item{
      {Id: 1, Number: 1, String: "1", Float: 1.0},
      {Id: 2, Number: 3, String: "3", Float: 3.0},
      {Id: 3, Number: 5, String: "5", Float: 5.0},
    }
  },
  schemabuilder.Paginated,
  schemabuilder.BatchSortField(
    "batchSortWithArgs",
    func(ctx context.Context, items map[batch.Index]Item, args UserArgs) (map[batch.Index]int64, error) {
      // Batch sort using args
    }),
  schemabuilder.BatchFilterField(
    "batchFilterWithArgs",
    func(items map[batch.Index]Item, args UserArgs) (map[batch.Index]string, error) {
      // Batch filter using args
    }),
)
```

### DIFF
--- a/graphql/connection_test.go
+++ b/graphql/connection_test.go
@@ -35,12 +35,23 @@ type ManualArgs struct {
 	PaginationArgs schemabuilder.PaginationArgs
 }
 
+type TimeBasedArgs struct {
+	StartMs int64
+	EndMs int64
+}
+
 type Item struct {
 	Id         int64
 	FilterText string
 	Number     int64
 	String     string
 	Float      float64
+}
+
+type AltItem struct {
+	Id int64
+	Time int64
+	Number int64
 }
 
 func TestConnection(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jl3329 (approved):
Would you mind including a link to the previous PR for reference to our comments?

PR #539, opened 2023-01-18
Weilian/introspection client schema

weilianchu:
This commit will supply the introspection client with the full introspection schema rather than justthe introspection query result.
This will allow the introspection schema to also serve custom queries with the introspective fields.

### DIFF
--- a/federation/introspection_client.go
+++ b/federation/introspection_client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 
+	"github.com/samsarahq/thunder/graphql"
 	"github.com/samsarahq/thunder/graphql/introspection"
 	"github.com/samsarahq/thunder/graphql/schemabuilder"
 )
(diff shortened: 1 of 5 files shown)


federation/introspection_client.go
```diff
@@ -4,25 +4,41 @@ import (
...
+		Schema: schema,
 	}
 }
 
 func (c *IntrospectionClient) Execute(ctx context.Context, request *QueryRequest) (*QueryResponse, error) {
+	if err := graphql.PrepareQuery(ctx, c.Schema.Query, request.Query.SelectionSet); err != nil {
+		return nil, err
+	}
+
+	executor := graphql.NewExecutor(graphql.NewImmediateGoroutineScheduler())
```
KevinXing, line 29: can we just embed the executor in the client so we don't need to create the executor each time?

weilianchu: good idea

KevinXing (approved):

PR #542, opened 2023-01-25
go federation: move introspection queries to its own client

weilianchu:
This commit adds a helper func that will help the planner determine whether or not the selection is part of an introspection query. If so, this field will be routed to be resolved by the introspection client.

### DIFF
--- a/federation/planner.go
+++ b/federation/planner.go
@@ -5,9 +5,10 @@ import (
 	"fmt"
 	"sort"
 	"strings"
-	
+
 	"github.com/samsarahq/go/oops"
 	"github.com/samsarahq/thunder/graphql"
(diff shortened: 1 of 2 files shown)


weilianchu:
I don't know why coveralls failed, the new helper is covered in `TestFederatedIntrospectionQuery` :(

weilianchu:
The comment changes are from goimports

alberttjin (approved):

alberttjin:
> I don't know why coveralls failed, the new helper is covered in `TestFederatedIntrospectionQuery` :(

is customService == "" in the unit test? maybe it doesn't enter that block
maybe adding to planner tests would be nice? idk

weilianchu:
I did a debug session and verified that the helper is covered 🤷

